PR #46, opened 2021-09-29
Update CRD definitions

pintohutch:
- Update `AlertingSpec` to make `.AlertManager` optional.
- Add `.AlertManagerConfigs` to allow users to bring their own `alertmanager_config`s instead of using our CRD.
- Rename `SafeAuthorization` to `Authorization`.

Note: only `types.go` was manually changed. The others were programmatically regenerated via:
```
make codegen && make crds
```


pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -79,7 +79,12 @@ type RuleEvaluatorSpec struct {
 // Taking inspiration from prometheus-operator: https://github.com/prometheus-operator/prometheus-operator/blob/2c81b0cf6a5673e08057499a08ddce396b19dda4/Documentation/api.md#alertingspec
 type AlertingSpec struct {
 	// Alertmanagers contains endpoint configuration for designated Alertmanagers.
-	Alertmanagers []AlertmanagerEndpoints `json:"alertmanagers"`
+	Alertmanagers []AlertmanagerEndpoints `json:"alertmanagers,omitempty"`
+	// AlertmanagerConfigs contains a list of secret or configmap selectors
```
fabxc, line 83: So `AlertmanagerConfigs` is as an alternative to `Alertmanagers` to inject hard-coded configs instead?

pintohutch: Ultimately because I saw prometheus-operator [supports](https://github.com/prometheus-operator/prometheus-operator/blob/f95159f8646012f3cb63f2b5ff0d693b7c2d6cdf/pkg/apis/monitoring/v1/types.go#L306-L317) this. I could see how users would want to bring their own config over using our CRD in the beginning.

PR #66, opened 2021-10-26
Macxamin/flush queues

macxamin:


pkg/export/export.go
```diff
@@ -467,43 +472,60 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+			e.triggerNext()
+		}
+	}
+
+	go func() {
+		// Flush shards on collector shutdown. Stop exporter from recieving new data.
+		// Try to drain the remaining shards in 15 seconds or less. This is done to
+		// prevent data loss during a shutdown.
+		<-parentCtx.Done()
+		e.opts.Disable = true
```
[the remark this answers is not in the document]
macxamin: The most straight forward solution is to use the exporter lock. Let me know what you think. I tried other solutions but I couldn't figure out how to use kingpin flags  with structs or atomic values. 

`type disabled struct {
    mtx  sync.Mutex
    disabled bool
}

func(d *disabled) (bool flag)
    mtx.Lock()
    d.disabled = flag
    mtx.Unlock()
`
Another way would require me to add an additional option in ExporterOpts and use the disabled flag to manage it.

pkg/export/export.go
```diff
@@ -467,43 +482,72 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+			close(done)
+		}()
+		select {
+		case <-done:
+			cancel()
+		case <-time.After(cancelTimeout):
+			level.Debug(e.logger).Log("msg", "Exporter timed out without exporting all of the pending buffered data.")
+			cancel()
+		}
+	}()
```
fabxc, line 542: The main loop below keeps executing until we completed the draining in this function.
That means that there are now two goroutines reading from the shards and sending data. That means that some state (e.g. `pendingShards`) is accessed concurrently without synchronization. But more importantly, we may send samples for the same stream in parallel, causing out-of-order writes w.r.t to the timestamps.
If you run/build with `go -race` the race detector is compiled it and should reliably trigger for such scenarios. When testing code like this, using the race detector is always a good idea.

So what we need to do is have the Run's main loop below stop/enter final state when at the first opportunity after the `parentCtx` is cancelled. So you can replace `case <-ctx.Done()` with `case <-parentCtx.Done()`.  
Then you can inline the contents of this function into that `case`. You don't need to watch for `ctx.Done()` anymore as there's no one else but "ourselves" as this function who could cancel it. And we do so (delayed)  after seeing `parentCtx` being canceled.

After that we're guaranteed to only execute `drainShards` once at a time.
Then you'll want to pass `drainShards` an explicit context so that for the regular sending you can pass in `ctx` and for the drain on shutdown you can pass in a new context instantiated via `context.WithTimeout(context.Background(), cancelTimeout)`.

I think after that things should be safe. The previous discussion might have steered things in a slightly different direction – but some interactions also only became clear to me when seeing the code evolve.

fabxc (commented):
So to wrap this up, a test would be nice of course.

I'd start out constructing a test case where your test sets up a fake CreateTimeSeries backend that records requests. Then you can run the exporter and have it send to this backend, call Export() with some data, and verify that your fake backend records the expected requests.
Once we have that, this can be extended to also verify the correct shutdown behavior.

[Here](https://stackoverflow.com/questions/42102496/testing-a-grpc-service) is a nice explanation for constructing a test backend. And [here](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/freinartz/ha-writes/pkg/lease/lease_test.go) is an example in larger context of some WIP I wrote that uses that approach.

Testing code like this can be challenging. It sometimes requires changing the original code to allow adjusting behavior for tests by making constants configurable, make constructors more flexible, or making functionality replaceable entirely (e.g. remember when you provided a custom queryFunc in the tests for the rule evaluator).

macxamin:
PTAL @pintohutch @fabxc

pkg/export/export.go
```diff
@@ -384,7 +403,7 @@ const (
 )
 
 // Run sends exported samples to Google Cloud Monitoring.
-func (e *Exporter) Run(ctx context.Context) error {
+func (e *Exporter) Run(parentCtx context.Context) error {
```
pintohutch, line 406: nit: I'm fine with this reverting back to `ctx` as we're renaming the contexts used here to distinguish.

pkg/export/export.go
```diff
@@ -489,21 +503,52 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+			// On Termination:
+			// 1. stop the exporter from recieving new data.
+			// 2. Cancel ongoing drainShardsAttempt.
+			// 3. Try to drain the remaining shards within the CancelTimeout.
+			// This is done to prevent data loss during a shutdown.
+			e.disable()
+			cancelLocalCtx()
+			go func() {
+				drained := false
+				timeoutCtx, timeoutCancelCtx := context.WithTimeout(ctxLocal, cancelTimeout)
```
pintohutch, line 540: Can we rename these to `drainCtx, drainCancel` for the same reasons above?

pkg/export/export.go
```diff
@@ -489,21 +503,52 @@ func (e *Exporter) Run(ctx context.Context) error {
...
+		select {
+		case <-parentCtx.Done():
+			// On Termination:
+			// 1. stop the exporter from recieving new data.
+			// 2. Cancel ongoing drainShardsAttempt.
+			// 3. Try to drain the remaining shards within the CancelTimeout.
+			// This is done to prevent data loss during a shutdown.
+			e.disable()
+			cancelLocalCtx()
+			go func() {
```
pintohutch, line 538: I don't see why this needs to be in a go routine.

pkg/export/export.go
```diff
@@ -467,19 +488,12 @@ func (e *Exporter) Run(ctx context.Context) error {
 	// Starting index when iterating over shards. This ensures we don't always start at 0 so that
 	// some shards may never be sent in a busy collector.
 	shardOffset := 0
-
-	for {
+	drainShardsAttempt := func(drainCtx context.Context) bool {
```
pintohutch, line 491: nit: I'm also cool with renaming `drainCtx` back to `ctx` as its contained within the function scope.

pintohutch (changes requested):
We should also have some unit tests as Fabian suggested. Should be fun - happy to help!

pkg/export/export.go
macxamin: There would be two ctx one for `Run` and one for `drainShardsAttempt`. Leaving as drainCtx as to not cause confustion.

pkg/export/export.go
```diff
@@ -342,9 +348,22 @@ func (e *Exporter) getExternalLabels() labels.Labels {
...
+func (e *Exporter) disable() {
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+	e.opts.Disable = true
+}
+
 // Export enqueues the samples to be written to Cloud Monitoring.
 func (e *Exporter) Export(metadata MetadataFunc, batch []record.RefSample) {
-	if e.opts.Disable {
+	if e.isDisabled() {
```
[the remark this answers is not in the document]
macxamin: @fabxc what do you think?

PR #72, opened 2021-11-02
Scrape port numbers even if not specified on Pod

fabxc:
If a port number is not specified on the Pod resource our previous
filtering mechanism didn't generate a target.
However, we can safely produce targets in this scenario in any case
as we generate a scrape config per specified endpoint and can simply
hardcode all target candidates to the given port number.

This turned out much easier than expected.

### DIFF
--- a/pkg/operator/apis/monitoring/v1alpha1/types.go
+++ b/pkg/operator/apis/monitoring/v1alpha1/types.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -333,16 +313,58 @@ func (pm *PodMonitoring) endpontScrapeConfig(index int) (*promconfig.ScrapeConfi
...
+			TargetLabel:  "instance",
+		})
+	} else if ep.Port.IntVal != 0 {
+		relabelCfgs = append(relabelCfgs, &relabel.Config{
+			Action:       relabel.Replace,
+			SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_name"},
+			Replacement:  fmt.Sprintf("$1:%d", ep.Port.IntVal),
+			TargetLabel:  "instance",
+		})
+		// If the input target was produced by a port declared in the Pod, this port number
```
pintohutch, line 356: What happens if you don't do this step?

fabxc: It uses the pre-populated __address__ which is whatever port this target was originally generated for.
Say the pod has 2 ports declared, 2 targets would be generated. We hard-override both to the same labels via the config here and Prometheus will dedupe them into a single target.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -374,7 +396,7 @@ func (pm *PodMonitoring) endpontScrapeConfig(index int) (*promconfig.ScrapeConfi
 	return &promconfig.ScrapeConfig{
 		// Generate a job name to make it easy to track what generated the scrape configuration.
 		// The actual job label attached to its metrics is overwritten via relabeling.
-		JobName:                 fmt.Sprintf("PodMonitoring/%s/%s/%s", pm.Namespace, pm.Name, portValue),
+		JobName:                 fmt.Sprintf("PodMonitoring/%s/%s/%s", pm.Namespace, pm.Name, &ep.Port),
```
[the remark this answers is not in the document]
fabxc: Yes, it actually does not without the `&` – which doesn't harm, but it's ugly.

pintohutch (approved):

PR #74, opened 2021-11-02
Avoid allocations in metadata retrieval func

fabxc:
This map slips through escape analysis and is actually allocated each time for no good reason.

This reduces allocated bytes/sec by about 25%, which is very significant. However, CPU and memory usage seem not affected really, which is very surprising as that correlates very directly usually. Makes me think this just counts towards the alloc metric but doesn't actually allocate new memory each time – not sure if that's a thing.

Anyway, at least this will remove noise from profiling for future improvements.

### DIFF
--- a/pkg/export/transform.go
+++ b/pkg/export/transform.go
@@ -68,37 +68,38 @@ func gaugeMetadata(metric string) (MetricMetadata, bool) {
 	}, true
 }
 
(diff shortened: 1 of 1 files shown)


pintohutch:
Wow - good catch! Approving. Unsure if this should go in for 0.1.0 release or not. Seems safe though.

pintohutch (approved):

PR #75, opened 2021-11-02
Rewrite POST requests as GET requests

fabxc:


cmd/frontend/main.go
```diff
@@ -154,9 +154,12 @@ func forward(logger log.Logger, target *url.URL, transport http.RoundTripper) ht
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		u := *target
 		u.Path = path.Join(u.Path, req.URL.Path)
-		u.RawQuery = req.URL.RawQuery
 
-		newReq, err := http.NewRequestWithContext(req.Context(), req.Method, u.String(), req.Body)
+		// Write all params into the URL and make a GET request.
```
pintohutch, line 158: Can we add a little more context for why in the comment?

pintohutch (commented):
Did you test this?

cmd/frontend/main.go
fabxc: Updated. Also constrained it to that one path just in case.

fabxc:
Yes

pintohutch (approved):

PR #76, opened 2021-11-03
Break Rules into separate resources per scope

fabxc:


pkg/operator/e2e/main_test.go
```diff
@@ -673,63 +673,67 @@ func validateCollectorUpMetrics(ctx context.Context, t *testContext, job string)
...
+	replace := strings.NewReplacer(
+		"{project_id}", projectID,
+		"{cluster}", cluster,
+		"{namespace}", t.namespace,
+	).Replace
 
 	// Create multiple rules in the cluster and expect their scoped equivalents
 	// to be present in the generated rule file.
-	files := []string{`
+	content := replace(`
```
pintohutch, line 685: Why use a string literal and `Unmarshal` here over the `ClusterRules` go struct directly? We use this pattern with our other CRDs in the tests `OperatorConfig`, `PodMonitoring`. Same for `Rules`.

cmd/operator/main.go
```diff
@@ -90,9 +97,7 @@ func main() {
 	op, err := operator.New(logger, cfg, metrics, operator.Options{
 		ProjectID:               *projectID,
 		Cluster:                 *cluster,
-		Location:                *location,
```
pintohutch, line 93: Why remove location initializing here?

pkg/operator/e2e/main_test.go
```diff
@@ -778,3 +788,63 @@ spec:
...
+		if err == iterator.Done {
+			t.Logf("No data, retrying...")
+			return false, nil
+		} else if err != nil {
+			return false, errors.Wrap(err, "querying metrics failed")
+		}
+		if len(series.Points) == 0 {
+			return false, errors.New("unexpected zero points in result series")
+		}
+		// We expect exactly one result.
```
pintohutch, line 840: Why is this? I can't find where we set our rules scraping interval to 10s...

cmd/operator/main.go
```diff
@@ -41,15 +42,21 @@ func unstableFlagHelp(help string) string {
...
 		logVerbosity      = flag.Int("v", 0, "Logging verbosity")
-		projectID         = flag.String("project-id", "", "Project ID of the cluster. May be left empty on GKE.")
-		location          = flag.String("location", "", "GCP location of the cluster. May be left empty on GKE.")
-		cluster           = flag.String("cluster", "", "Name of the cluster the operator acts on. May be left empty on GKE.")
+		projectID         = flag.String("project-id", defaultProjectID, "Project ID of the cluster. May be left empty on GKE.")
+		location          = flag.String("location", "", "GCP location of the cluster. Maybe be left empty on GKE.")
+		cluster           = flag.String("cluster", defaultCluster, "Name of the cluster the operator acts on. May be left empty on GKE.")
 		operatorNamespace = flag.String("operator-namespace", operator.DefaultOperatorNamespace,
 			"Namespace in which the operator manages its resources.")
-		publicNamespace = flag.String("public-namespace", operator.DefaultOperatorNamespace,
```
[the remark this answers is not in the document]
fabxc: No idea how that happened.

cmd/operator/main.go
fabxc: Something happened in this file and I'm not sure what (:

pkg/operator/e2e/main_test.go
fabxc: Yea, I stumbled over this, too. I think some historical indecisiveness :) We should change it at some point.

pkg/operator/e2e/main_test.go
fabxc: This is for the series, not the points within it.

pintohutch: Ah right.

pintohutch (approved):
% nits

pkg/operator/e2e/main_test.go
```diff
@@ -580,12 +586,6 @@ func testCollectorSelfPodMonitoring(ctx context.Context, t *testContext) {
 // validateCollectorUpMetrics checks whether the scrape-time up metrics for all collector
 // pods can be queried from GCM.
 func validateCollectorUpMetrics(ctx context.Context, t *testContext, job string) {
-	// We rely on the default service account of the collector having write access to GCM.
```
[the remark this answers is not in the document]
fabxc: Because it's no longer true. Last section was essentially a TOOD to allow providing explicit credentials, which we recently added.

PR #78, opened 2021-11-05
DO NOT MERGE: Introduce dynamic user agent propagation

pintohutch:
Manually tested PodMonitoring, Rules, ClusterRules examples. Validated presence in GCM and alerts being sent to a local alertmanager via `amtool` successfully.
```
k apply -f examples/setup.yaml
k apply -f examples/operator.yaml
k apply -f examples/example-app.yaml
k apply -f examples/pod-monitoring.yaml
k apply -f examples/rules.yaml
k apply -f examples/clusterrules.yaml
```

- Clusterrole needed clusterrules access
- Bumped image tag everywhere else
- nit: add build/ to .gitignore to ignore any generated binaries during local builds


cmd/operator/deploy/operator/clusterrole.yaml
```diff
@@ -50,6 +50,7 @@ rules:
   - daemonsets
   - validatingwebhookconfigurations
   - certificatesigningrequests
+  - clusterrules
```
fabxc, line 53: Good catch!

fabxc (commented):
We also need to bump the version in the user agent variable .

fabxc:
Let's merge https://github.com/GoogleCloudPlatform/prometheus-engine/pull/79 first

fabxc:
Reg bump of user agent. Currently it says

```
	ClientName = "gmp-collector"
	Version    = "v2.28.1-gmp.0"```

But as it is used by different clients (prometheus and rule-evaluator) we should just identify the library as a client there. Then the respective binary should add additional info to the user agent (hardcoded or via flag).

So something like this would make sense

```
	ClientName = "prometheus-engine/export"
	Version    = "v0.1.0"```

fabxc (approved):
LGTM % nits

pintohutch:
Hrm - so I was tempted to merge this, but realized if we want the collector image to reflect the user-agent changes here, we'd need to:
1. Change the collector image used everywhere here to `gke.gcr.io/prometheus-engine/prometheus:v2.28.1-gmp.2-gke.0`.
2. Merge this and tag 0.1.0
3. Bump the prometheus-engine dependency in https://github.com/GoogleCloudPlatform/prometheus to v0.1.0.
4. Create and merge a new PR with the changes there.
5. Tag `v2.28.1-gmp.2` there.
6. Trigger a build to push the new public collector image on gke.gcr.io.
7. (Ideally) test the examples manifests here again with the updated images.

This felt like a lot to consider and commit to without discussing on a Friday afternoon, so I'm holding off on merging until we can discuss next week. The fallback here can be to use the current `gke.gcr.io/prometheus-engine/prometheus:v2.28.1-gmp.1-gke.1` collector, but we'd lose the user-agent info for the release.

fabxc:
Yea, that's definitely quite the chain. I'd honestly just do the 0.1.0 with
the updated user agent and no Prometheus version bump.

It's not inhibiting the functionality in any way so this is fine to
propagate over 2 releases. So with 0.1.1 or so we can switch to a newer
Prometheus version.

pkg/export/export.go
```diff
@@ -377,12 +392,6 @@ func (e *Exporter) triggerNext() {
 	}
 }
 
-// ClientName and Version are used to identify to User Agent. TODO(maxamin): automate versioning.
-const (
-	ClientName = "gmp-collector"
-	Version    = "v2.28.1-gmp.0"
-)
```
fabxc, line 384: Let's leave these in and not allow users to prevent them from showing up in the user agent. There's no good reason to override this. Looking at [common UA formatting](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent):

`Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:47.0) Gecko/20100101 Firefox/47.0`

So our values should probably be

```
ClientName = "prometheus-engine-export"
Version = "0.1.0"
```

So that then with a binary-specific addition the UA would look like

```
Prometheus/2.28.1-gmp.1 prometheus-engine-export/0.1.0
rule-evaluator/0.1.0 prometheus-engine-export/0.1.0
```

The first part could be made overrideable via flags or envvar as you did here.

fabxc (changes requested):

pkg/export/export.go
pintohutch: Done. See my comments around why an env var was preferred.

pkg/operator/collection.go
```diff
@@ -265,6 +268,8 @@ func (r *collectionReconciler) makeCollectorDaemonSet(spec *monitoringv1alpha1.C
 						// RPC applications) and gives us a more balanced ratio of memory and CPU usage.
 						Env: []corev1.EnvVar{
 							{Name: "GOGC", Value: "25"},
+							{Name: export.GMPApp, Value: GMPAppVal},
```
fabxc, line 271: These variables are undefined.

pkg/export/export.go
```diff
@@ -40,6 +41,11 @@ import (
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 )
 
+const (
```
fabxc, line 44: Abbreviations like VER probably do more harm than good. I'd also be in favor of explicitly overriding the UA with a single var so these envvars don't get repurposed for a bunch of different purposes over time. Once these start having conflicting requirements it gets messy.

pkg/export/export.go
```diff
@@ -365,6 +394,12 @@ func (e *Exporter) Export(metadata MetadataFunc, batch []record.RefSample) {
 	e.triggerNext()
 }
 
+const (
+	ClientName = "prometheus-engine-export"
+	// TODO(pintohutch): make version centralized and populated at build time.
+	Version = "v0.1.0"
```
fabxc, line 400: Need to drop the v at least if we want to align with usual UA format.

pkg/export/export.go
```diff
@@ -148,6 +156,9 @@ type ExporterOpts struct {
 	BatchSize uint
 	// Prefix under which metrics are written to GCM.
 	MetricTypePrefix string
+
+	// UserAgent to include in export gRPC requests.
+	UserAgent string
```
fabxc, line 161: This should only be needed one either Exporter or ExporterOptions

cmd/rule-evaluator/main.go
```diff
@@ -69,6 +69,9 @@ func main() {
 	}
 
 	exporterOptions := export.NewFlagOptions(a)
+	// TODO(danielclark): make version centralized and populated at build time.
+	exporterOptions.UserAgent = "rule-evaluator/prometheus-engine/export:v0.1.0"
```
fabxc, line 73: Per common formatting no `:`, no `v`, and no `/` except for separating component from version.

fabxc (changes requested):
This has expanded too far functionality-wise for a release tag PR and there are too many open points that block us from doing the release timely.

I submitted this PR with my last comment on updating the fixed UA constants in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/81. We can work on on dynamic setting added here for a follow-up version.

pkg/export/export.go
```diff
@@ -228,12 +239,30 @@ func New(logger log.Logger, reg prometheus.Registerer, opts ExporterOpts) (*Expo
 	if opts.BatchSize > batchSizeMax {
 		return nil, errors.Errorf("Maximum supported batch size is %d, got %d", batchSizeMax, opts.BatchSize)
 	}
+	// Inject managed collection state via environment variable.
+	// An environment variable was preferred overp a flag because:
```
macxamin, line 243: overp -> over

pintohutch:
Marked this PR to not merge. Leaving it open as a starting off reference point to a dedicated PR for dynamic user agent injection.

fabxc:
Closing due to staleness. We've got a simple version to set the UA added in #128

PR #82, opened 2021-11-09
Fix rule-evaluator reloading, Cut 0.1.1

fabxc:


cmd/operator/deploy/operator/operator.yaml
```diff
@@ -81,7 +81,7 @@ spec:
       serviceAccountName: operator
       containers:
       - name: operator
-        image: gke.gcr.io/prometheus-engine/operator:v0.1.0-gke.0
+        image: gke.gcr.io/prometheus-engine/operator:v0.1.1-gke.0
```
pintohutch, line 84: Can we mimic the args from examples/operator.yaml here?

pintohutch (approved):
Checked and all the versions are bumped.

cmd/operator/deploy/operator/operator.yaml
fabxc: Done. Also mirrored the other way by adding priority class to the examples one.
We should get rid of the duped YAMLs soon.

PR #83, opened 2021-11-11
Set tolerations and DNS policy on collector

fabxc:


pintohutch (approved):

eljefedelrodeodeljefe:
@fabxc can we update the examples as noted here https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-managed unfortunately from the outside of Google I cannot tell which images will be referenced and such.

Also to confirm: when tolerations are not set such as in 0.1.1 and one has tainted nodes, I assume scraping a from said tainted node will not be possible?

fabxc:
You are correct, in that case scraping on these nodes will not work. We've to do a new release with these changes included. Once we've done that, we'll update the documentation. If you urgently need some changes, you can of course always build a new image version yourself and use that.

eljefedelrodeodeljefe:
Appreciate it. Thanks

PR #84, opened 2021-11-12
Implement HA exporting

fabxc:


pkg/lease/lease_test.go
```diff
@@ -0,0 +1,285 @@
...
+		lease.Register(func(start, end time.Time, ok bool) {
+			hook(name, start, end, ok)
+		})
+		// Make replicas die after a random amount of time unless a fixed lifetime is given.
+		if replicaLifetime == 0 {
+			replicaLifetime = time.Duration(rand.Intn(int(50 * leaseOpts.ExtensionPeriod)))
+		}
+		t.Logf("starting %s with lifetime %s", name, replicaLifetime)
+
+		ctx, cancel := context.WithTimeout(ctx, replicaLifetime)
```
pintohutch, line 245: Here you immediately overwrite the timeout context `ctx` initialized above. Is this intentional? As this is in the for loop, it's also creating a chain of contexts with various timeouts. I don't quite follow why...

What's wrong with having a single parent context for all the replicas and timing out?

pkg/lease/lease_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// Assume well-formed requests and just validate timestamps according to the logic
+	// of the real API.
+	// We expect all clients to write against the same lease for the test.
+	iv := req.TimeSeries[0].Points[0].Interval
+	// First write always succeeds.
+	if srv.lastInterval != nil {
+		// The end timestamp must not go into the past and not be too shortly after
+		// the last one.
+		if iv.EndTime.AsTime().Before(srv.lastInterval.EndTime.AsTime()) {
```
pintohutch, line 69: IIUC these error conditions are tolerated by the lease where the state machine will proceed to follow mode. Is there utility of the error message flavors? AFAICT if they are triggered, they wouldn't cause test failures.

pkg/lease/lease_test.go
```diff
@@ -0,0 +1,285 @@
...
+func TestLease(t *testing.T) {
+	opts := &Options{
+		UpdatePeriod:    10 * time.Millisecond,
+		RetryPeriod:     1 * time.Millisecond,
+		ExtensionPeriod: 30 * time.Millisecond,
+	}
+
+	t.Run("one-lease-interval", func(t *testing.T) {
+		// This test ensures that the lease is not flip-flopping between participants if they
+		// remain alive.
```
pintohutch, line 271: This test is asserting more than that right? The `if len(intervals) != 1` looks to be checking that only one replica owned the lease during the duration of the test.

pkg/lease/lease_test.go
```diff
@@ -0,0 +1,285 @@
...
+		// while we don't have randomized transient errors.
+		if gap := iv2.start.Sub(iv1.end); gap > ri.leaseOpts.ExtensionPeriod+ri.leaseOpts.UpdatePeriod {
+			return errors.Errorf("gap of %s between intervals %s and %s too large", gap, iv1, iv2)
+		}
+		iv1 = iv2
+	}
+	return nil
+}
+
+func testLease(t *testing.T, duration time.Duration, leaseOpts *Options, replicaCount int, replicaLifetime time.Duration) []interval {
```
pintohutch, line 173: `replicaCount` is unused. I think you meant to put it in the `replicaCh` init.

pkg/lease/lease_test.go
fabxc: The replicas actually don't live as long as the test. `replicaCount` limits how many there are at a time, but they individually die  after a random amount of time. Otherwise the first replica  to get the lease would just hold it for the entire test duration.

I named it `replicaCtx` so it's not as confusing.

pkg/lease/lease_test.go
fabxc: The implementation only checks the code so the message here is just for potential debugging.

pkg/lease/lease_test.go
```diff
@@ -0,0 +1,305 @@
...
+		baseTime: ri.baseTime,
+		// A very quickly acquired lease may reach up to 1ms into the past with an
+		// empty initial state, so account for that.
+		start: ri.baseTime.Add(-time.Millisecond),
+		end:   ri.baseTime.Add(-time.Millisecond),
+	}
+
+	for _, iv2 := range ri.intervals() {
+		// Intervals are closed and thus must strictly not overlap.
+		if !iv1.end.Before(iv2.start) && !iv2.end.Before(iv1.start) {
```
pintohutch, line 163: This check reads a bit awkward to me since `ri.intervals()` returns a sorted list. Is there any reason we can't check simply:
```go
if iv2.start.Before(iv1.end) || iv2.start.Equal(iv1.end) {
```

pkg/lease/lease_test.go
```diff
@@ -0,0 +1,305 @@
...
+			return errors.Errorf("intervals for replica %s and %s overlap", iv1, iv2)
+		}
+		// Gap shouldn't be too large. It should in principle be 1 extension period plus 1 update period.
+		// But schedulding might make it a bit larger.
+		if gap := iv2.start.Sub(iv1.end); gap > 2*ri.leaseOpts.ExtensionPeriod {
```
pintohutch, line 168: Can you explain why the gap could be larger than 1 update period + scheduling? I would think if the gap is larger than that, there's some race in acquiring the lease, and we'd want to error.

pkg/lease/lease_test.go
```diff
@@ -0,0 +1,305 @@
...
+			t.Fatalf("expected exactly one interval with long-running replicas but got %d", len(intervals))
+		}
+	})
+	t.Run("two-intermittent-replicas", func(t *testing.T) {
+		// This test attempts to trigger a bad condition by having a lease change often across
+		// a high number of replicas.
+		t.Parallel()
+		testLease(t, 10*time.Second, opts, 2, 0, 0)
+	})
+	t.Run("many-intermittent-replicas", func(t *testing.T) {
```
pintohutch, line 293: These last two tests are testing the same thing, with the latter being a stronger test. The `opts` between the two are the same, so the `UpdatePeriod` should be limiting the lease change frequency in both cases.

PR #87, opened 2021-11-16
User Agent population release update guide

macxamin:


pintohutch:
Discussed with @macxamin offline about collaborating on user agent stuff in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/78 first.

pintohutch (commented):
So I would use [github permalinks](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files).

Also, examples are [autogenerated](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/85) now, so only `deploy/` yamls should be bumped.

Another thing to consider, if the future dynamic user-agent population change simplifies this, it may be worth waiting or abandoning this change until then.

PR #89, opened 2021-11-24
Correctly use webhook listen address

fabxc:
The flag was not used previously. This made the webhook server default
to 9443 while the unset port field in the ValidatingWebhookConfiguration
defaults to 443. Hence the webhooks were never active.

This is a good thing to test once we've a suite that runs the operator in a cluster.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -18,7 +18,9 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"path/filepath"
+	"strconv"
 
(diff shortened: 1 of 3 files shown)


pintohutch:
I believe they were active before the controller-runtime migration, but this looks like one of the things that fell through the cracks during that.

pintohutch (approved):
Good catch!

fabxc:
Ah, that makes sense.
The fail-open policy on the WH certainly makes this easy to miss. We should look into changing that down the line.

PR #96, opened 2021-12-03
Presubmit fixes and documentation

pintohutch:
- Added documentation on using `make presubmit`.
- Removed kind-test's private gcr registry auth as this is currently unnecessary with public images.
- Reverted the e2e test for rule-evaluator deployment to check for UpdatedReplicas over AvailableReplicas, as the test wasn't passing due to the readiness probe marking the deployment as Unavailable.
```Replicas:               1 desired | 1 updated | 1 total | 0 available | 1 unavailable```

### CODE REVIEW

**Review by tkinash — APPROVED** [review]

File: pkg/operator/e2e/main_test.go [hunk 1]
```diff
@@ -340,7 +340,7 @@ func testRuleEvaluatorDeployment(ctx context.Context, t *testContext) {
 		// When not using GCM, we check the available replicas rather than ready ones
 		// as the rule-evaluator's readyness probe does check for connectivity to GCM.
 		if skipGCM {
-			if *deploy.Spec.Replicas != deploy.Status.AvailableReplicas {
+			if *deploy.Spec.Replicas != deploy.Status.UpdatedReplicas {
```
fabxc, line 343: Oh, I misunderstood ready vs available.

What we maybe could do at some point is spin up a local CreateTimeSeries endpoint (takes maybe 20 LOC) and point kind tests at that one.

hack/presubmit.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# As this command can be slow, optimize by only running if there's difference
+# from the origin/main branch.
+codegen_diff || update_codegen
+go mod tidy && go mod vendor && go fmt ${SCRIPT_ROOT}/...
+update_crdgen
+update_examples
+update_docgen
+go test `go list ${SCRIPT_ROOT}/... | grep -v operator/e2e`
+"$@"
```
fabxc, line 125: For development/debugging it would probably still be useful to run individual functions in this file quickly.

Maybe there can be n `all` function here which contains this block and runs by default. Otherwise the first CLI arg is invoked.

Dockerfile
```diff
@@ -39,6 +39,5 @@ COPY --from=hermetic /workspace/go.sum go.sum
 COPY --from=hermetic /workspace/cmd cmd
 COPY --from=hermetic /workspace/doc doc
 COPY --from=hermetic /workspace/examples examples
-COPY --from=hermetic /workspace/hack hack
 COPY --from=hermetic /workspace/pkg pkg
-COPY --from=hermetic /workspace/vendor vendor
\ No newline at end of file
+COPY --from=hermetic /workspace/vendor vendor2
```
fabxc, line 43: Intentional?

pintohutch: Yes - I noticed that the `COPY` command wouldn't completely overwrite the local directory (e.g. if some files were deleted in the docker build stage, those deletions wouldn't be mirrored in the destination context). Copying to a different directory and subsequently replacing it works, albeit a bit hacky. I changed the name to `vendor.tmp` to make it more obvious.

hack/presubmit.sh
pintohutch: Done.

pkg/operator/e2e/main_test.go
pintohutch: Added a TODO.

hack/presubmit.sh
```diff
@@ -0,0 +1,178 @@
...
+        examples)
+          update_examples
+          ;;
+        format)
+          reformat
+          ;;
+        test)
+          run_tests
+          ;;
+      esac
```
fabxc, line 173: This would just exit when an unknown command is provided, right? Worth adding a default error message.

fabxc (approved):
LGTM % the one comment

PR #97, opened 2021-12-03
Add metric relabeling to PodMonitoring

fabxc:
Allow configuring metric relabeling through the PodMonitoring CRD with
some restrictions to not override protected labels. As a side effect
this strengthens validations of targetLabels and generated scrape
configs in general.


pkg/operator/apis/monitoring/v1alpha1/types_test.go
```diff
@@ -124,7 +128,99 @@ func TestValidatePodMonitoring(t *testing.T) {
...
+					},
+				},
+				TargetLabels: TargetLabels{
+					FromPod: []LabelMapping{
+						{From: "key1", To: "foo-bar"},
+					},
+				},
+			},
+			fail:        true,
+			errContains: `"foo-bar" is invalid 'target_label' for replace action`,
```
pintohutch, line 150: IIUC this error is formatted and returned by the Prometheus library rather than prometheus-engine code - just wanted to confirm.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -402,25 +429,132 @@ func (pm *PodMonitoring) endpontScrapeConfig(index int) (*promconfig.ScrapeConfi
...
+}
+
+var protectedLabels = []string{
+	export.KeyProjectID,
+	export.KeyLocation,
+	export.KeyCluster,
+	export.KeyNamespace,
+	export.KeyJob,
+	export.KeyInstance,
+	"__address__",
```
pintohutch, line 511: Why do we need to protect the __address__ label? Does upstream not protect it during unmarshalling or reading?

pintohutch (commented):
LGTM % can your rebase and `make presubmit`? Then I'll give it another sanity check and can approve.

pkg/operator/apis/monitoring/v1alpha1/types.go
fabxc: Upstream actually deliberate makes the scraped IP address available under that label so one can modify it via relabeling. We just don't want to allow it. It's no longer available during metric relabeling, which this PR adds, but in principle we also need to protect it in `targetLabels.fromPod`.

pkg/operator/apis/monitoring/v1alpha1/types_test.go
fabxc: Yes, so this is in principle at risk of breaking with dependency updates. But that'll be easy to fix and I the error validation in this test is necessary to confirm we triggered the intended logic and don't get a false positive due to a different error.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -402,25 +429,132 @@ func (pm *PodMonitoring) endpontScrapeConfig(index int) (*promconfig.ScrapeConfi
...
+		return nil, errors.Errorf("relabeling with action %q not allowed", r.Action)
+	case relabel.Keep, relabel.Drop:
+		// These actions don't modify a series and are OK.
+	default:
+		return nil, errors.Errorf("unknown relabeling action %q", r.Action)
+	}
+	return rcfg, nil
+}
+
+var protectedLabels = []string{
```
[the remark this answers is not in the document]
fabxc: Either works in general. I think a slice here has a few benefits

1. Reproducible errors – as map iterations are randomized we could get different errors when validating the same resource repeatedly.
1. We can print this with `strings.Join` in our error messages.
1. Probably faster than a map in practice.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -463,6 +602,28 @@ type LabelMapping struct {
 	To string `json:"to,omitempty"`
 }
 
+// RelabelingRule defines a single Prometheus relabeling rule.
+type RelabelingRule struct {
```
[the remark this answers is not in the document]
fabxc: Yes, exactly. Also those structs use `yaml` struct annotations while the libs and codegens we use want `json` tags.

vendor/modules.txt
```diff
@@ -83,6 +83,7 @@ github.com/golang/protobuf/proto
 github.com/golang/protobuf/ptypes
 github.com/golang/protobuf/ptypes/any
 github.com/golang/protobuf/ptypes/duration
+github.com/golang/protobuf/ptypes/empty
```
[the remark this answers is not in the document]
fabxc: done

pkg/operator/apis/monitoring/v1alpha1/types.go
pintohutch: Can we at least at a TODO for that, probably in `labelMappingRelabelConfigs`?

vendor/modules.txt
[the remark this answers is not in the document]
pintohutch: My only thought would be changing https://github.com/GoogleCloudPlatform/prometheus-engine/blob/4c8d1232c4be66f0e576d9f08531d70d778c717c/hack/presubmit.sh#L124 to 
```bash
rm -rf vendor && go mod tidy && go mod vendor && go fmt ${SCRIPT_ROOT}/...
```
and try again (outside of an IDE or any bash script that could be auto-running something?)

pkg/operator/apis/monitoring/v1alpha1/types.go
fabxc: A TODO about what?

vendor/modules.txt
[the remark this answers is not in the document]
fabxc: go.sum shouldn't have anything to do with it. It just caches checksums from my understanding but doesn't contribute to what's actaully vendored.

I emptied vendor/ and go.sum locally though, ran `make presubmit` again and got an empty diff again afterwards (i.e. empty stuff included again). Probably worth debugging outside of this PR.

pkg/operator/apis/monitoring/v1alpha1/types.go
pintohutch: Ah yes - ignore me, I got confused

pintohutch (approved):

PR #98, opened 2021-12-06
Add `app` label for gke get app name.

andysim3d:


pkg/operator/collection.go
```diff
@@ -151,6 +151,7 @@ func (r *collectionReconciler) ensureCollectorSecrets(ctx context.Context, spec
 			Namespace: r.opts.OperatorNamespace,
 			Labels: map[string]string{
 				LabelAppName: NameCollector,
+				KubernetesAppName: CollectorAppName,
```
pintohutch, line 154: You shouldn't need this on the secret - just pods.

pkg/operator/operator.go
```diff
@@ -71,6 +71,11 @@ const (
 	ImageCollector      = "gke.gcr.io/prometheus-engine/prometheus:v2.28.1-gmp.1-gke.1"
 	ImageConfigReloader = "gke.gcr.io/prometheus-engine/config-reloader:v0.1.1-gke.0"
 	ImageRuleEvaluator  = "gke.gcr.io/prometheus-engine/rule-evaluator:v0.1.1-gke.0"
+
+	// The k8s Application, will be exposed as component name.
+	KubernetesAppName     = "app"
+	CollectorAppName      = "managed_prometheus_collector"
+	RuleEvaluatorAppName  = "managed_promehteus_rule_evaulator"
```
pintohutch, line 78: And we'll want a `app: managed_prometheus_operator` here https://github.com/GoogleCloudPlatform/prometheus-engine/blob/eec06b4fc3c4bf801ff9cc81665074a1854ef1a8/cmd/operator/deploy/operator/05-deployment.yaml#L28 and here https://github.com/GoogleCloudPlatform/prometheus-engine/blob/eec06b4fc3c4bf801ff9cc81665074a1854ef1a8/cmd/operator/deploy/operator/05-deployment.yaml#L34.

pkg/operator/collection.go
andysim3d: done.

pkg/operator/operator.go
andysim3d: done.

pintohutch: nit: typo `managed_prometheus_rule_evaluator`

pintohutch (approved):

pkg/operator/operator.go
andysim3d: done

PR #99, opened 2021-12-07
per gke's suggestion, use dash instead of under.

andysim3d:
Also amend sample yaml file, move selector label to metadata label.

### DIFF
--- a/cmd/operator/deploy/operator/05-deployment.yaml
+++ b/cmd/operator/deploy/operator/05-deployment.yaml
@@ -18,6 +18,7 @@ metadata:
   namespace: gmp-system
   name: gmp-operator
   labels:
+    app: managed-prometheus-operator
     app.kubernetes.io/name: gmp-operator
     app.kubernetes.io/component: operator
     app.kubernetes.io/part-of: gmp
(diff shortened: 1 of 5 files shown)


cmd/operator/deploy/operator/05-deployment.yaml
```diff
@@ -18,14 +18,14 @@ metadata:
   namespace: gmp-system
   name: gmp-operator
   labels:
+    app: managed_prometheus_operator
```
pintohutch, line 21: It looks like you moved this without reformatting to dashes.

pintohutch (changes requested):
Also, be sure to run `make presubmit` to regenerate affected manifests.

cmd/operator/deploy/operator/05-deployment.yaml
andysim3d: done

cmd/operator/deploy/operator/05-deployment.yaml
```diff
@@ -18,21 +18,21 @@ metadata:
...
   name: gmp-operator
   labels:
+    app: managed-prometheus-operator
     app.kubernetes.io/name: gmp-operator
     app.kubernetes.io/component: operator
     app.kubernetes.io/part-of: gmp
 spec:
   replicas: 1
   selector:
     matchLabels:
```
pintohutch, line 28: Can you add it back here and rerun `make presubmit`? It should regenerate the latest `examples/operator.yaml` with your change here.

andysim3d: done.

pintohutch (approved):

PR #101, opened 2021-12-08
Use controller-gen to create CRD yamls

pintohutch:
* The old method only patched existing manifests with a lot of custom
scripting. This uses controller-gen more conventially.
* Move crds to autogenerated yaml paths in their own folder.
* Fix bug in Makefile test target.
* Update presubmit.sh script to use combine function more broadly.


Makefile
```diff
@@ -52,9 +52,12 @@ test:        ## Run all tests. Writes real data to GCM API under PROJECT_ID envi
...
 	@echo ">> running tests"
 ifeq ($(DOCKER), 1)
-	$(call docker_build, ./hack/Dockerfile --target hermetic -t gmp/hermetic \
-	--build-arg RUNCMD='go test `go list ./... | grep -v operator/e2e`')
+	$(call docker_build, -f ./hack/Dockerfile --target hermetic -t gmp/hermetic \
+	--build-arg RUNCMD='go test `go list ./... | grep -v operator/e2e`' .)
 else
+	kubectl apply -f examples/setup.yaml
+	kubectl apply -f cmd/operator/deploy/operator/01-priority-class.yaml
+	kubectl apply -f cmd/operator/deploy/operator/03-clusterrole.yaml
```
fabxc, line 60: +1

caught me a couple of times already

examples/setup.yaml
```diff
@@ -11,575 +11,760 @@
...
+                              type: string
+                            description: A set of labels to attach to the result of
+                              the query expression.
+                            type: object
+                          record:
+                            description: Record the result of the expression to this
+                              metric name. Only one of `record` and `alert` must be
+                              set.
+                            type: string
+                        required:
```
fabxc, line 108: The required lists are still edited manually on each generation?

fabxc (approved):

examples/setup.yaml
pintohutch: Discussed offline - consensus is that this is controlled via `"omitempty"` json tag in types.go - e.g. https://github.com/GoogleCloudPlatform/prometheus-engine/blob/55ae264b745f71b5fa28cb4daabd09d432c81aa0/pkg/operator/apis/monitoring/v1alpha1/types.go#L598

PR #102, opened 2021-12-08
Add ClusterPodMonitoring CRD

pintohutch:
* Support cluster-scoped podmonitoring across all namespaces with a
dedicated CRD.
* `make presubmit` to auto-generate docs, CRD yaml, and go client code.
* Commit to `certificates.k8s.io/v1` and 1.19+ k8s clusters in e2e tests. Supporting e2e tests for `certificates.k8s.io/v1beta1` as well is probably not necessary and saves some spaghetti code.


pkg/operator/crdstate.go
```diff
@@ -22,34 +22,47 @@ import (
...
 
 type pmState struct {
 	pm          *monitoringv1alpha1.PodMonitoring
+	cm          *monitoringv1alpha1.ClusterPodMonitoring
 	needsUpdate bool
 	conds       map[monitoringv1alpha1.MonitoringConditionType]*monitoringv1alpha1.MonitoringCondition
 }
 
-func newPMState(pm *monitoringv1alpha1.PodMonitoring, now metav1.Time) *pmState {
+func newPMState(pm *monitoringv1alpha1.PodMonitoring, cm *monitoringv1alpha1.ClusterPodMonitoring, now metav1.Time) *pmState {
```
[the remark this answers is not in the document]
pintohutch: I refactored a bit.

Since these are all instance methods, I didn't see returning an extra slice in the various function signatures as a huge win. Taking advantage of the `collectionReconciler`'s state and using a simple resource object slice there felt cleaner, so that's what I did.

Happy to discuss more.

pkg/operator/crdstate.go
```diff
@@ -58,17 +71,18 @@ func newPMState(pm *monitoringv1alpha1.PodMonitoring, now metav1.Time) *pmState
...
+		cMons: make(map[string]*pmState),
+		now:   now,
 	}
 }
 
 // SetPodMonitoringCondition adds the provided PodMonitoring resource to the managed state
 // along with the provided condition iff the resource generation has changed or there
 // is a status condition state transition.
-func (c *CRDStatusState) SetPodMonitoringCondition(pm *monitoringv1alpha1.PodMonitoring, cond *monitoringv1alpha1.MonitoringCondition) error {
+func (c *CRDStatusState) SetPodMonitoringCondition(obj metav1.Object, obsGen int64, cond *monitoringv1alpha1.MonitoringCondition) error {
```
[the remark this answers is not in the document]
fabxc: At the code level sure but not at the CRD level I think.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -219,20 +297,80 @@ func (pm *PodMonitoring) ScrapeConfigs() (res []*promconfig.ScrapeConfig, err er
...
+	GetTargetLabels() TargetLabels
+}
+
+func endpointScrapeConfig(msc monitoringScrapeConfig, ep ScrapeEndpoint) (*promconfig.ScrapeConfig, error) {
+	var (
+		ns           = msc.GetNamespace()
+		name         = msc.GetName()
+		sel          = msc.GetSelector()
+		targetLabels = msc.GetTargetLabels()
+	)
```
fabxc, line 369: Why not just add these 4 as args directly rather than defining an interface and adding methods to both types?
The signature sure is longer but it's overall a good chunk less stuff to comprehend. Also makes refactoring  simpler when we add differentiating attributes and this single generation function for both CRDs will no longer suffice.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -219,20 +297,80 @@ func (pm *PodMonitoring) ScrapeConfigs() (res []*promconfig.ScrapeConfig, err er
 	return res, nil
 }
 
+// SetPodMonitoringCondition merges the provided PodMonitoring resource to the
+//
+// along with the provided condition iff the resource generation has changed or there
```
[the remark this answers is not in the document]
pintohutch: Done.

pintohutch (commented):
Removed the interface approach and just duplicated the entire method.

pkg/operator/collection.go
```diff
@@ -138,9 +140,14 @@ func (r *collectionReconciler) Reconcile(ctx context.Context, req reconcile.Requ
 	if err := r.ensureCollectorConfig(ctx, &config.Collection); err != nil {
 		return reconcile.Result{}, errors.Wrap(err, "ensure collector config")
 	}
-	if err := r.updateCRDStatus(ctx); err != nil {
-		return reconcile.Result{}, errors.Wrap(err, "update crd status")
+
+	// Reconcile any status updates.
+	for _, obj := range r.statusUpdates {
+		if err := r.client.Status().Update(ctx, obj); err != nil {
```
fabxc, line 146: Mh, maybe log an error and try the next one regardless?

Also `statusUpdates` doesn't seem to ever be reset.

pintohutch: Ah good call - done.

pkg/operator/collection.go
```diff
@@ -138,9 +141,16 @@ func (r *collectionReconciler) Reconcile(ctx context.Context, req reconcile.Requ
...
+			logger.Error(err, "update status")
```
fabxc, line 148: I think referencing the object it failed for would be good as I'm not sure that the error will contain that info by default.

fabxc (approved):
LGTM after last comment

PR #105, opened 2021-12-14
add helm chart for managing google managed prometheus

shinsekai:
First of all, manifests which are used in this PR comes from [GoogleCloudPlatform/prometheus-engine](https://github.com/GoogleCloudPlatform/prometheus-engine).

Then, as describe within README.md, CRDs files should be installed separately (e.g. [Helm chart best practices](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts))

As discussed together by emails, I added `gmp-system` namespace creation directly in the Helm Chart for atomically destroying all resources related to the operator itself if necessary:
- gmp-operator
- gmp-frontend
- kubernetes objects (e.g serviceaccount, clusterrole, clusterrolebinding, priorityclass, service)

and ressources which are created and managed by the operator such as: 
- collector's daemonset
- rule-evaluator's deployment


chart/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
```
fabxc, line 23: For .gitignore a rule of thumb is that various IDEs, .DS_Store etc. should be in a users system-wide config rather than being repeated within individual projects. Is that an option for Helm as well?
Ideally we'd only explicitly ignore things that are part of this repo (or created by scripts within it).

chart/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v2
+name: gmp
+description: A Helm chart for deploying Google Managed Prometheus operator
```
fabxc, line 3: "Google Managed Service for Prometheus operator" :)

chart/templates/frontend_deployment.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "gmp.name" . }}-{{ .Values.frontend.name }}
```
fabxc, line 4: I'd suggest we put this into the gmp-public namespace as its a deployment meant to be talked to by other services in the cluster and not wihtin the set managed by the operator.

fabxc (commented):
Thank you for this PR!

I did a first round of reviewing. My comments largely revolve around reducing the amount of templating where we don't have a use case for user-defined values yet. The less templating is going on the more feasible it will be to keep this chart in sync with the base manifests it was derived from.

chart/.helmignore
shinsekai: I updated its content as requested by just ignore `doc/` which is not required during `helm package` process

chart/templates/frontend_deployment.yaml
shinsekai: Indeed, we can move the frontend component from `gmp-system` namespace to `gmp-public` it's relevant.

chart/README.md
```diff
@@ -0,0 +1,51 @@
...
+This helm chart generate elements for Google Managed Prometheus:
+
+- Namespace
+- PriorityClass
+- ClusterRole
+- ClusterRoleBinding
+- Deployment
+- Service
+- ServiceAccount
+- ConfigMap
```
pintohutch, line 12: You can remove `ConfigMap` from the list.

chart/templates/operator_operatorconfig.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: monitoring.googleapis.com/v1alpha1
+kind: OperatorConfig
```
pintohutch, line 2: As this is a CRD, it would need to be installed outside of helm, correct?

chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,72 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Operator selector labels
+*/}}
+{{- define "gmp.operatorSelectorLabels" -}}
+app.kubernetes.io/name: {{ include "gmp.name" . }}-{{ .Values.operator.name }}
+app.kubernetes.io/component: {{ .Values.operator.name }}
+app.kubernetes.io/part-of: gmp
```
pintohutch, line 51: We also want `app: managed-prometheus-operator` here, mirroring from this [commit](https://github.com/GoogleCloudPlatform/prometheus-engine/commit/55ae264b745f71b5fa28cb4daabd09d432c81aa0).

chart/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Retrieving metrics schema
+
+![retrieving metrics](doc/read_metrics.png)
+
+### Google Managed Prometheus CRDs
+
+Please note that CRDs are not installed through this Chart, they should by installed separately as recommended by Helm Community [Separate Charts](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts)
+
+Make sure CRDs are already deployed in the target cluster where you want to deploy Google Managed Prometheus engine therefore it will not works !
```
pintohutch, line 26: nit: "to deploy managed collection, otherwise it will not work!"

pintohutch (commented):
I also echo Fabian's comments around minimizing config values where necessary.

Otherwise - this is a great contribution - thank you!

chart/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+![collecting metrics](doc/write_metrics.png)
+
+### Retrieving metrics schema
+
+![retrieving metrics](doc/read_metrics.png)
+
+### Google Managed Prometheus CRDs
+
+Please note that CRDs are not installed through this Chart, they should by installed separately as recommended by Helm Community [Separate Charts](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts)
```
robertb724, line 24: I think it would be beneficial to host the `gmp-crds` chart alongside the `gmp` chart. Otherwise, users will need to keep up with the correct CRDs using their own process. I find the upgrade process for `kube-prometheus-stack` to be cumbersome, running the `kubectl apply -f crd.yaml` in every cluster is not ideal. Linkerd has started to follow this process: https://github.com/linkerd/linkerd2/tree/main/charts

pintohutch:
Hello - and thank you again for this contribution. We've decided to hold off on helm support for the time being. This is mainly due to:
- Configuration of the k8s manifests is largely covered by the [OperatorConfig CR](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/91d173174e102bb55405c9d486afc19be45eb3dd/doc/api.md#operatorconfig). Changing field values outside of that can lead to unpredictable behavior.
- Any manifest templating in the future will likely leverage standard [go templating](https://pkg.go.dev/text/template) and will be rolled into the repository build system to ensure stability.
- Designating fields in the manifest as "helm-configurable" can be subjective; often it will vary by use-case. Users who wish to configure the deployment of managed collection are free to do so in their own helm charts!

PR #106, opened 2021-12-14
Removing cert self-sign option from operator

macxamin:
Webhooks with self-signed certs will not be accepted by the kubernetes API server.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -73,8 +73,6 @@ func main() {
 			"Priority class at which the collector pods are run.")
 		gcmEndpoint = flag.String("cloud-monitoring-endpoint", "",
 			"Override for the Cloud Monitoring endpoint to use for all collectors.")
-		caSelfSign = flag.Bool("ca-selfsign", true,
-			"Whether to self-sign or have kube-apiserver sign certificate key pair for TLS.")
 		webhookAddr = flag.String("webhook-addr", ":8443",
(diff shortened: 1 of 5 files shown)


fabxc:
Can you run `make presubmit` and submit any potential changes as well?

macxamin:
@fabxc ran `make presubmit`. No new files were generated.

cmd/operator/main.go
```diff
@@ -73,14 +73,16 @@ func main() {
...
 		caSelfSign = flag.Bool("ca-selfsign", true,
 			"Whether to self-sign or have kube-apiserver sign certificate key pair for TLS.")
 		webhookAddr = flag.String("webhook-addr", ":8443",
 			"Address to listen to for incoming kube admission webhook connections.")
 		metricsAddr = flag.String("metrics-addr", ":18080", "Address to emit metrics on.")
 	)
 	flag.Parse()
-
+	// TODO(maxamin) Solves the "declared but not used", remove after GKE is updated.
+	_ = caSelfSign
```
fabxc, line 85: Why is this necessary? We'd update the binary and the flags it is deployed with in one change,no? @pintohutch

pintohutch: Yep - the GKE component is safe as we'll remove the option there after we tag and do a release here.

pintohutch: So it's safe to delete this flag entirely here

fabxc (approved):

PR #107, opened 2021-12-15
Use go 1.17 in our build environments

pintohutch:
* This fixes the `make presubmit` failures.
* Use Docker builds elsewhere by default.
* Build managed collection images as part of presubmit.
* Add documentation fixes.


go.mod
```diff
@@ -51,6 +51,84 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require (
```
fabxc, line 54: Why a separate block? 

Guess those are all indirect. But the above one also still contains indirect ones.

fabxc (approved):

go.mod
pintohutch: Not sure what happened there. Manually cleaned that up and reran `make presubmit`.

macxamin (approved):
Great Job! `make presubmit` works now.

pintohutch:
Cool - I also spot checked this by deploying manually and doing various e2e tests in addition to the automated tests, as this involves bumping our build base images. Thanks for the reviews!

PR #109, opened 2021-12-21
Don't set empty proxy URL

fabxc:
Setting it to the empty string actually defines a valid URL that then
fails to proxy through at scrape time. Thus we must make sure that
the parameter is truly unset in the generated config.


pintohutch (approved):
just verified the e2e tests work. LGTM

PR #110, opened 2021-12-21
Remove example-app and relocated benchmark

pintohutch:
* Removed example-app in favor of the one used in `examples/` as having both seems unnecessary
* And some nit fixes in documentation
* Re-ran `make presubmit`


fabxc:
Our example app notably has the ability to simulate CPU and memory utilization, which the other one does not.
Currently we've no recurring use so I'm fine with removing it. We can restore it in case we need it again.

Please squash commits on merge :)

fabxc (approved):

PR #111, opened 2021-12-22
Add webhook TLS cert flags

macxamin:
1. Added flags to pass-in certificate payload to the operator.
2. Added function ensureCerts that will handle the cert/key uses.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -16,6 +16,7 @@ package operator
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -126,6 +127,10 @@ type Options struct {
 	HostNetwork bool
 	// Priority class for the collector pods.
 	PriorityClass string
+	// Certificate of the server in base 64.
+	Cert string
+	// Key of the server in base 64.
+	Key string
 	// Endpoint of the Cloud Monitoring API to be used by all collectors.
 	CloudMonitoringEndpoint string
 	// Webhook serving address.
@@ -240,30 +245,9 @@ func New(logger logr.Logger, clientConfig *rest.Config, registry prometheus.Regi
 // custom resources and registers handlers with the webhook server.
 // The passsed owner references are set on the created WebhookConfiguration resources.
 func (o *Operator) setupAdmissionWebhooks(ctx context.Context, ors ...metav1.OwnerReference) error {
-	// Persisting TLS keypair to a k8s secret seems like unnecessary state to manage.
-	// It's fairly trivial to re-generate the cert and private
-	// key on each startup. Also no other GMP resources aside from the operator
-	// rely on the keypair.
-	// A downside to this approach is re-writing the validation webhook config
-	// every time with the new caBundle. This should only happen when the operator
-	// restarts, which should be infrequent.
-	var (
-		crt, key []byte
-		err      error
-		fqdn     = fmt.Sprintf("system:node:%s.%s.svc", NameOperator, o.opts.OperatorNamespace)
-	)
-
-	// Generate kube-apiserver-signed certificate/key pair.
-	crt, key, err = CreateSignedKeyPair(ctx, o.kubeClient, fqdn)
+	crt, err := o.ensureCerts(ctx, o.manager.GetWebhookServer().CertDir)
 	if err != nil {
 		return err
-	}
-
-	if err := ioutil.WriteFile(filepath.Join(o.manager.GetWebhookServer().CertDir, "tls.crt"), crt, 0666); err != nil {
-		return errors.Wrap(err, "create cert file")
-	}
-	if err := ioutil.WriteFile(filepath.Join(o.manager.GetWebhookServer().CertDir, "tls.key"), key, 0666); err != nil {
-		return errors.Wrap(err, "create key file")
 	}
 
 	whCfg := validatingWebhookConfig(
@@ -319,6 +303,43 @@ func (o *Operator) Run(ctx context.Context, ors ...metav1.OwnerReference) error 
 	o.logger.Info("starting GMP operator")
 
 	return o.manager.Start(ctx)
+}
+
+// ensureCerts writes the cert/key files to the specified directory.
+// If cert/key are not avalilable, generate them.
+func (o *Operator) ensureCerts(ctx context.Context, dir string) ([]byte, error) {
+	var (
+		crt, key []byte
+		err      error
+	)
+	if (len(o.opts.Key) == 0 && len(o.opts.Cert) > 0) || (len(o.opts.Cert) == 0 && len(o.opts.Key) > 0) {
+		return nil, errors.Errorf("Flags key-base64 and cert-base64 must both be set.")
+	} else if len(o.opts.Key) > 0 && len(o.opts.Cert) > 0 {
+		crt, err = base64.StdEncoding.DecodeString(o.opts.Cert)
+		if err != nil {
+			return nil, errors.Wrap(err, "decoding TLS certificate")
+		}
+		key, err = base64.StdEncoding.DecodeString(o.opts.Key)
+		if err != nil {
+			return nil, errors.Wrap(err, "decoding TLS key")
+		}
+	} else {
+		// Generate kube-apiserver-signed certificate/key pair.
+		fqdn := fmt.Sprintf("system:node:%s.%s.svc", NameOperator, o.opts.OperatorNamespace)
+		crt, key, err = CreateSignedKeyPair(ctx, o.kubeClient, fqdn)
+		if err != nil {
+			return nil, errors.Wrap(err, "generating kube-apiserver-signed certificate/key pair")
+		}
+	}
+
+	if err := ioutil.WriteFile(filepath.Join(dir, "tls.crt"), crt, 0666); err != nil {
+		return nil, errors.Wrap(err, "create cert file")
+	}
+	if err := ioutil.WriteFile(filepath.Join(dir, "tls.key"), key, 0666); err != nil {
+		return nil, errors.Wrap(err, "create key file")
+	}
+
+	return crt, nil
 }
 
 // namespacedNamePredicate is an event filter predicate that only allows events with
--- a/pkg/operator/operator_test.go
+++ b/pkg/operator/operator_test.go
@@ -1,0 +1,196 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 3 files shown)


pkg/operator/operator.go
```diff
@@ -321,6 +305,41 @@ func (o *Operator) Run(ctx context.Context, ors ...metav1.OwnerReference) error
...
 }
 
+// ensureCerts writes the cert/key files to the specified directory.
+// If cert/key are not avalilable, generate them.
+func (o *Operator) ensureCerts(ctx context.Context, dir string) ([]byte, error) {
+	var (
+		crt, key []byte
+		err      error
+	)
+	if len(o.opts.Key) > 0 && len(o.opts.Cert) > 0 {
```
fabxc, line 315: We should throw an error if only one is set.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,131 @@
+package operator
```
fabxc, line 1: License header missing

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"time"
+
+	v1 "k8s.io/api/certificates/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+const dir = "_tmp"
```
fabxc, line 20: I'd generally suggest moving this into the test func to not pollute the identifier space of the whole package.

But actually, here you may rather want to us this stdlib function: https://pkg.go.dev/io/ioutil#TempDir

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,131 @@
...
+	var (
+		timeout     = 3 * time.Second
+		ctx, cancel = context.WithTimeout(context.Background(), timeout)
+		csr         *v1.CertificateSigningRequest
+		fqdn        = fmt.Sprintf("system:node:%s.%s.svc", NameOperator, "")
+		certBytes   = []byte{1, 2, 3, 4}
+	)
+	t.Cleanup(cancel)
+
+	os.Mkdir(dir, 0777)
```
fabxc, line 32: Unchecked error.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			op := Operator{
+				opts:       tc.opts,
+				kubeClient: client,
+			}
+
+			_, err := op.ensureCerts(ctx, dir)
+			if err == nil == tc.expectErr {
```
fabxc, line 99: I got really confused syntactically initially. A more common way to write this would be with two conditions and dedicated errors (`err != nil && !tc.expectErr` and `err == nil && tc.expectErr`) or at least by bracing them (`err == nil) == tc.expectErr`).

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,131 @@
...
+				t.Errorf("want err: %v; got %v", tc.expectErr, err)
+			}
+			// Test outputed files.
+			if err == nil {
```
fabxc, line 103: You can reduce indentation by checking `err != nil` instead and returning early.
I think in this case even simpler would be to return (or use Fatalf instead of Errorf) after emitting the error in the previous condition and not checking again here at all.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,131 @@
...
+		csr         *v1.CertificateSigningRequest
+		fqdn        = fmt.Sprintf("system:node:%s.%s.svc", NameOperator, "")
+		certBytes   = []byte{1, 2, 3, 4}
+	)
+	t.Cleanup(cancel)
+
+	os.Mkdir(dir, 0777)
+	defer func() {
+		os.RemoveAll(dir)
+	}()
```
fabxc, line 35: You can simply write `defer os.RemoveAll(dir)`. The additional closure would only help if you changed the value of `dir` again further down as it will then only be resolved when the deferred function is actually called on exit.

pkg/operator/operator.go
```diff
@@ -321,6 +305,41 @@ func (o *Operator) Run(ctx context.Context, ors ...metav1.OwnerReference) error
...
+		crt, err = base64.StdEncoding.DecodeString(o.opts.Cert)
+		if err != nil {
+			return nil, err
+		}
+		key, err = base64.StdEncoding.DecodeString(o.opts.Key)
+		if err != nil {
+			return nil, err
```
fabxc, line 322: Wrap the errors to give the user context which string was the issue.

pkg/operator/operator_test.go
macxamin: how about now?

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,165 @@
...
+			// Test outputed files.
+			outCert, err := ioutil.ReadFile(path.Join(dir, "tls.crt"))
+			if err != nil {
+				t.Fatalf("error reading from cert file: %v", err)
+			}
+			outKey, err := ioutil.ReadFile(path.Join(dir, "tls.key"))
+			if err != nil {
+				t.Fatalf("error reading from key file: %v", err)
+			}
+			if len(op.opts.Cert) == 0 && len(op.opts.Key) == 0 {
```
fabxc, line 147: So at this point there is different validation logic for some test cases as well as a mock server only used by one of them. I think at this point the table-driven test loses it's benefit and I'd just split them into two tests, one for server-signed ones and one for explicit ones.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,199 @@
...
+	var (
+		timeout     = 3 * time.Second
+		ctx, cancel = context.WithTimeout(context.Background(), timeout)
+		certBytes   = []byte{1, 2, 3, 4}
+	)
+	t.Cleanup(cancel)
+
+	dir, err := ioutil.TempDir("", "test_ensure_certs")
+	if err != nil {
+		log.Fatal(err)
```
fabxc, line 155: t.Fatal

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,199 @@
...
+			_, err := op.ensureCerts(ctx, dir)
+			if err != nil {
+				if !tc.expectErr {
+					t.Fatalf("want err: %v; got %v", tc.expectErr, err)
+				}
+				return
+			}
+			// Test outputed files.
+			outCert, outKey := readKeyAndCertFiles(dir, t)
+			if eq := bytes.Compare(outCert, certBytes); eq != 0 {
```
fabxc, line 190: You can also convert []byte to string and compare that way – can be a bit easier to read `string(outCert) == string(certByteS)`.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		t.Fatalf("error reading from key file: %v", err)
+	}
+	return outCert, outKey
+}
+
+func TestEnsureCertsExplicit(t *testing.T) {
+	dir, err := ioutil.TempDir("", "test_ensure_certs")
+	if err != nil {
+		log.Fatal(err)
```
fabxc, line 88: t.Fatal

fabxc (approved):

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+// Mock asynchronous kube-apiserver.
+func mockKubeAPIServer(ctx context.Context, client *fake.Clientset, t *testing.T) {
+	var (
+		csr       *v1.CertificateSigningRequest
+		timeout   = 3 * time.Second
+		fqdn      = fmt.Sprintf("system:node:%s.%s.svc", NameOperator, "")
```
pintohutch, line 38: nit - can you put in `"test-ns"` or something over `""` for the namespace to make clear?

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,197 @@
...
+	t.Cleanup(cancel)
+
+	dir, err := ioutil.TempDir("", "test_ensure_certs")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
+	client := fake.NewSimpleClientset()
+	mockKubeAPIServer(ctx, client, t)
```
pintohutch, line 158: Can we switch the go routine such that you issue it here instead, while changing `mockKubeAPIServer` to be a blocking function?
i.e.
`go mockKubeAPIServer(ctx, client, t)`

and remove the go routine in the actual function. That reads a little cleaner to me and I can tell what's going on better.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+	for _, tc := range []struct {
+		desc       string
+		opts       Options
+		expectCert string
```
pintohutch, line 163: this isn't used currently. i'd just make this a byte array and move the `certBytes` var value to here and assert against that.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,197 @@
...
+		expectKey  string
```
pintohutch, line 164: not used - delete.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,197 @@
...
+			expectCert: "1,2,3,4",
+			expectErr:  false,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			op := Operator{
+				opts:       tc.opts,
+				kubeClient: client,
+			}
+			_, err := op.ensureCerts(ctx, dir)
```
pintohutch, line 179: Might as well check the return cert from `op.ensureCerts` matches what's written to disk, right? i.e.
```go
cert, err := op.ensureCerts(ctx, dir)
...
outCert, outKey := readKeyAndCertFiles

if string(outCert) != string(cert) {
    // t.fail yada yada
} else if string(outCert) != string(tc.expectCert) {
    // t.fail yada yada
}
```

pintohutch (commented):
LGTM % refine the unit test a little bit. Nice job!

pkg/operator/operator_test.go
macxamin: That causes this error `panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x1c0 pc=0x172323a]`
@fabxc any tips?

macxamin: resolved.

PR #114, opened 2022-01-06
Improve Alertmanager discovery

fabxc:


pintohutch (approved):
IIUC this enables alertmanager discovery when there's no `Service` explicitly created? Or when the service-generated `Endpoints` object is customized in some way?

fabxc:
Yes, this primarily makes it easier to use custom Endpoints objects as it no longer requires a matching Service

PR #115, opened 2022-01-06
Use fixed label selectors

pintohutch:
* Revert to using the label selectors defined in v0.1.1.
* Add comment to never modify them.

### DIFF
--- a/pkg/operator/operator_config.go
+++ b/pkg/operator/operator_config.go
@@ -304,6 +304,7 @@ func (r *operatorConfigReconciler) makeRuleEvaluatorDeployment(spec *monitoringv
 		fmt.Sprintf("--config.file=%s", path.Join(configOutDir, configFilename)),
 		fmt.Sprintf("--web.listen-address=:%d", RuleEvaluatorPort),
 	}
+
 	if r.opts.ProjectID != "" {
 		evaluatorArgs = append(evaluatorArgs, fmt.Sprintf("--export.label.project-id=%s", r.opts.ProjectID))
 	}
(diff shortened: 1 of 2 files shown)


pkg/operator/operator_config.go
```diff
@@ -334,10 +335,16 @@ func (r *operatorConfigReconciler) makeRuleEvaluatorDeployment(spec *monitoringv
 	}
 	replicas := int32(1)
 
+	// DO NOT MODIFY - label selectors are immutable by the Kubernetes API.
+	// see: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates.
+	podLabelSelector := map[string]string{
+		LabelAppName: NameRuleEvaluator,
+	}
```
fabxc, line 342: The other resources still us `rulesLabels()` is there value in having them be different?

pintohutch: That's the main issue. We need the `app` label on the pods for GKE integration. However, adding `app` to the label selector of the deployment breaks the reconcile loop as label selectors are [immutable](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates).

If we're reverting to the non-`app` label selector as that was on the v0.1.1 deployment which most people are using., then they need to be different.

pintohutch: Right - we want `app: managed-prometheus-rule-evaluator` on the rule-evaluator pod(s) and `app: managed-prometheus-collector` on the Prometheus pods. This will allow GKE to discover and record `managed-prometheus` resource addon metrics for SLIs. The label selectors of the owning deployment and daemonset should remain unchanged.

fabxc (approved):

PR #117, opened 2022-01-10
Add global rules

fabxc:
Implement a GlobalRules resource which evalutes rules without any
scoping


pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -1156,6 +1156,36 @@ type ClusterRulesList struct {
 	Items           []ClusterRules `json:"items"`
 }
 
+// GlobalRules defines Prometheus alerting and recording rules that are scoped
+// to all data in the queried project. Only metric data from the current cluster is processed
```
pintohutch, line 1160: The godoc needs to be updated; this is a description of `ClusterRules` behavior.

pintohutch (approved):
LGTM % nits

pkg/operator/apis/monitoring/v1alpha1/types.go
fabxc: No :) it is adapted.

fabxc: Ah, actually forgot to delete one old sentence. Fixed.

pintohutch: So I'm a bit hung up on
```
Only metric data from the current cluster is processed
```
Shouldn't this be something like
```
Only metric data from the queried project is processed
```

fabxc: Yes, that's the old sentence I meant. Didn't push anymore yesterday after the presubmit run.

PR #123, opened 2022-01-17
Fix webhook certificate setup

fabxc:
* Remove CSR-based certificates as with v1 there's no clear path how to
  use the approach as intended while keeping complexity in check.
* Bring back (and always fall back to) self-signed certificates. They
  stopped working due to the system:node prefixing introduced for the
  CSR-base certs before.
* Fix service port for webhook endpoint.


macxamin (approved):
LGTM

andysim3d (approved):
LGTM

PR #124, opened 2022-01-17
Revert "Fix "dockerception" and use sibling containers over child containers"

fabxc:
Reverts GoogleCloudPlatform/prometheus-engine#120

This change made `make kindtest` and `make presubmit` start failing the collector test for me. The collector crashloops as it cannot find default credentials and doesn't get explicit ones either.
This is different from the case for which our tests have the `--skip-gcm` flag, where it has credentails, but they lack permissions.

I'm not entirely sure what the reason is and how it relates to this change. But undoing first so the tests work again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,6 @@ PROJECT_ID?=$(shell gcloud config get-value core/project)
 GMP_CLUSTER?=gmp-test-cluster
 GMP_LOCATION?=us-central1-c
 API_DIR=pkg/operator/apis
-# For now assume the docker daemon is mounted through a unix socket.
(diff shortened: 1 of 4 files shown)


andysim3d (approved):

macxamin (approved):

pintohutch:
I am curious what failures you experienced here. I just checked out https://github.com/GoogleCloudPlatform/prometheus-engine/commit/4df26fe777e6ab03edea70216d9714cff0e5d713 on a fresh cloud shell instance and ran both `make presubmit` and `make kindtest` with no issues...

PR #125, opened 2022-01-18
Flush queues on collector shutdown

maxamins:
1. When shutdown is initiated stop more messages from being sent
2. Drain the shards and send the batches in under 15 seconds

### DIFF
--- a/pkg/export/export.go
+++ b/pkg/export/export.go
@@ -172,6 +172,8 @@ const (
 	// Time after context is cancelled that we use to flush the remaining buffered data.
 	// This avoids data loss on shutdown.
 	cancelTimeout = 15 * time.Second
+	// Time after the final shards are drained before the exporter is closed on shutdown.
+	flushTimeout = 100 * time.Millisecond
 	// Prefix for GCM metric.
 	MetricTypePrefix = "prometheus.googleapis.com"
 )
@@ -557,23 +559,9 @@ func (e *Exporter) SetLabelsByIDFunc(f func(storage.SeriesRef) labels.Labels) {
 	e.seriesCache.getLabelsByRef = f
 }
 
-// isDisabled checks if Exporter is disabled.
-func (e *Exporter) isDisabled() bool {
-	e.mtx.Lock()
-	defer e.mtx.Unlock()
-	return e.opts.Disable
-}
-
-// disable stops requests from being sent to Monarch.
-func (e *Exporter) disable() {
-	e.mtx.Lock()
-	defer e.mtx.Unlock()
-	e.opts.Disable = true
-}
-
 // Export enqueues the samples and exemplars to be written to Cloud Monitoring.
 func (e *Exporter) Export(metadata MetadataFunc, batch []record.RefSample, exemplarMap map[storage.SeriesRef]record.RefExemplar) {
-	if e.isDisabled() {
+	if e.opts.Disable {
 		gcmExportCalledWhileDisabled.Inc()
 		return
 	}
@@ -767,6 +755,11 @@ func (e *Exporter) Run() error {
 	send := func() {
 		e.mtx.RLock()
 		opts := e.opts
+		if e.metricClient == nil {
+			// Flush timeout reached, runner is shut down.
+			e.mtx.RUnlock()
+			return
+		}
 		sendFunc := e.metricClient.CreateTimeSeries
 		e.mtx.RUnlock()
 
@@ -790,9 +783,10 @@ func (e *Exporter) Run() error {
 		curBatch = newBatch(e.logger, opts.Efficiency.ShardCount, opts.Efficiency.BatchSize)
 	}
 
-	// Try to drain the remaining data before exiting or the timelimit (15 seconds) expires.
+	// Try to drain the remaining data before exiting or the time limit (15 seconds) expires.
 	// A sleep timer is added after draining the shards to ensure it has time to be sent.
 	drainShardsBeforeExiting := func() {
+		//nolint:errcheck
 		level.Info(e.logger).Log("msg", "Exiting Exporter - will attempt to send remaining data in the next 15 seconds.")
 		exitTimer := time.NewTimer(cancelTimeout)
 		drained := make(chan struct{}, 1)
@@ -803,14 +797,18 @@ func (e *Exporter) Run() error {
 				for _, shard := range e.shards {
 					_, remaining := shard.fill(curBatch)
 					totalRemaining += remaining
+					shard.mtx.Lock()
 					pending = pending || shard.pending
+					shard.mtx.Unlock()
 					if !curBatch.empty() {
 						send()
 					}
 				}
 				if totalRemaining == 0 && !pending {
-					// Wait for the final shards to send.
-					time.Sleep(100 * time.Millisecond)
+					// NOTE(ridwanmsharif): the sending of the batches happen asyncronously
+					// and we only wait for a fixed amount of time after the final batch is sent
+					// before shutting down the exporter.
+					time.Sleep(flushTimeout)
 					drained <- struct{}{}
 				}
 			}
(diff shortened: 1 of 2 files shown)


pintohutch:
What's the status on this PR?

pkg/export/export.go
```diff
@@ -514,14 +541,54 @@ func (e *Exporter) Run(ctx context.Context) error {
 		curBatch = newBatch(e.logger, e.opts.BatchSize)
 	}
 
+	// Try to drain the remaining data before exiting or the timelimit (15 seconds) expires.
+	// A sleep timer is added after draining the shards to ensure it has time to be sent.
+	drainShardsBeforeExiting := func() {
+		level.Info(e.logger).Log("msg", "Exiting Exporter - will attempt to send remaining data in the next 15 seconds.")
```
macxamin, line 547: This gets printed twice. Any reason why?

pkg/export/export_test.go
```diff
@@ -387,3 +387,64 @@ func TestExporter_drainBacklog(t *testing.T) {
...
+	}
+
+	e.metricClient = metricClient
+	e.SetLabelsByIDFunc(func(i uint64) labels.Labels {
+		return labels.FromStrings("project_id", "test", "location", "test")
+	})
+
+	exportCtx, cancelExport := context.WithCancel(context.Background())
+
+	for i := 0; i < 50; i++ {
```
macxamin, line 427: Maxes out on 2048 samples. I'm assuming because they only go to one shard. Do you know how to create the shard in a different way?

pintohutch:
Aside from rebasing - we'll want to benchmark this change to catch any performance regressions - cc @saketjajoo

ridwanmsharif:
@pintohutch @bwplotka PTAL, there is https://github.com/GoogleCloudPlatform/prometheus/pull/203 that has some benchmarking results but spoilers: there is no discernible change in performance - which makes sense cause this is exclusively only run on shutdown.

pintohutch:
That's great to hear! I'll defer to @bwplotka as the SME to review.

pkg/export/export.go
```diff
@@ -162,7 +169,11 @@ const (
 	// Time after an accumulating batch is flushed to GCM. This avoids data being
 	// held indefinititely if not enough new data flows in to fill up the batch.
 	batchDelayMax = 50 * time.Millisecond
-
+	// Time after context is cancelled that we use to flush the remaining buffered data.
+	// This avoids data loss on shutdown.
+	cancelTimeout = 15 * time.Second
+	// Time after the final shards are drained before the exporter is closed on shutdown.
+	flushTimeout = 100 * time.Millisecond
```
bwplotka, line 176: is 100ms making any difference?

pkg/export/export.go
```diff
@@ -765,13 +783,61 @@ func (e *Exporter) Run() error {
...
+	}
+
 	for {
 		select {
 		// NOTE(freinartz): we will terminate once context is cancelled and not flush remaining
 		// buffered data. In-flight requests will be aborted as well.
 		// This is fine once we persist data submitted via Export() but for now there may be some
 		// data loss on shutdown.
 		case <-e.ctx.Done():
+			// on termination, try to drain the remaining shards within the CancelTimeout.
```
bwplotka, line 836:
```suggestion
			// On termination, try to drain the remaining shards within the CancelTimeout.
```

pkg/export/export.go
```diff
@@ -765,13 +783,61 @@ func (e *Exporter) Run() error {
...
+				return
+			case <-drained:
+				return
+			}
+		}
+	}
+
 	for {
 		select {
 		// NOTE(freinartz): we will terminate once context is cancelled and not flush remaining
```
bwplotka, line 831: Let's update this commentary, it's incorrect with the new logic 🤗

pkg/export/export.go
```diff
@@ -765,13 +783,61 @@ func (e *Exporter) Run() error {
 		curBatch = newBatch(e.logger, opts.Efficiency.ShardCount, opts.Efficiency.BatchSize)
 	}
 
+	// Try to drain the remaining data before exiting or the time limit (15 seconds) expires.
+	// A sleep timer is added after draining the shards to ensure it has time to be sent.
+	drainShardsBeforeExiting := func() {
```
bwplotka, line 788: nit: Not a must, but it might be easier to iterate on if we move `send` and `drain` to separate `Exporter` private methods 🤔

pkg/export/export.go
```diff
@@ -765,13 +783,61 @@ func (e *Exporter) Run() error {
...
 	for {
 		select {
 		// NOTE(freinartz): we will terminate once context is cancelled and not flush remaining
 		// buffered data. In-flight requests will be aborted as well.
 		// This is fine once we persist data submitted via Export() but for now there may be some
 		// data loss on shutdown.
 		case <-e.ctx.Done():
+			// on termination, try to drain the remaining shards within the CancelTimeout.
+			// This is done to prevent data loss during a shutdown.
+			drainShardsBeforeExiting()
```
bwplotka, line 838: Are we sure this works? All sends normally take `e.ctx` context, so they will be cancelled? I assume here you (intentionally?) accept that and drain the buffer with new sends with a separate context.

Would it be cleaner to have a custom, separate context from the beginning?  🤔

pkg/export/export.go
```diff
@@ -765,13 +783,61 @@ func (e *Exporter) Run() error {
...
+					totalRemaining += remaining
+					shard.mtx.Lock()
+					pending = pending || shard.pending
+					shard.mtx.Unlock()
+					if !curBatch.empty() {
+						send()
+					}
+				}
+				if totalRemaining == 0 && !pending {
+					// NOTE(ridwanmsharif): the sending of the batches happen asyncronously
```
bwplotka, line 808: Yea, can we improve this? And get that result state from send (or create separate send?)

pkg/export/export.go
```diff
@@ -765,13 +783,61 @@ func (e *Exporter) Run() error {
...
+					// and we only wait for a fixed amount of time after the final batch is sent
+					// before shutting down the exporter.
+					time.Sleep(flushTimeout)
+					drained <- struct{}{}
+				}
+			}
+		}()
+		for {
```
bwplotka, line 816: I would suggest we don't do another goroutine and we do all in this for loop? `Send` is async anyway?

pkg/export/export.go
```diff
@@ -548,15 +561,15 @@ func (e *Exporter) SetLabelsByIDFunc(f func(storage.SeriesRef) labels.Labels) {
 
 // Export enqueues the samples and exemplars to be written to Cloud Monitoring.
 func (e *Exporter) Export(metadata MetadataFunc, batch []record.RefSample, exemplarMap map[storage.SeriesRef]record.RefExemplar) {
+	if e.opts.Disable {
+		gcmExportCalledWhileDisabled.Inc()
```
bwplotka, line 565: That's unrelated to the main PR goal, I see Max did this, let's at least update description.

pkg/export/export.go
```diff
@@ -765,13 +783,61 @@ func (e *Exporter) Run() error {
...
 		// NOTE(freinartz): we will terminate once context is cancelled and not flush remaining
 		// buffered data. In-flight requests will be aborted as well.
 		// This is fine once we persist data submitted via Export() but for now there may be some
 		// data loss on shutdown.
 		case <-e.ctx.Done():
+			// on termination, try to drain the remaining shards within the CancelTimeout.
+			// This is done to prevent data loss during a shutdown.
+			drainShardsBeforeExiting()
+			// This channel is used for unit test case.
+			e.exitc <- struct{}{}
```
bwplotka, line 840: Ideally we don't have testing specific code in critical, production flow. However, IF we make sure sending propagates some results to us back, we could have a nice production logic that emits log line telling us all was flushed successfully, can we do this?

pkg/export/export.go
bwplotka: in fact we always use e.ctx even in drain if I look correctly, are we use this code works? 🤔

bwplotka (changes requested):
Thanks! 

Some suggestions  might require some work, so let me know if this is something you have time to do, or we accept some imperfection/code spaghetti.

The main issue is to me with proceeding on send work despite reusing e.ctx which was cancelled already. 🤔

PR #127, opened 2022-01-18
Validate rules resources

fabxc:
Validate the various rules resources including running the upstream
validation logic on the generated result. While this doesn't yield the
best error messages, it gives the best guarantee that the result can
be picked up by the rule-evaluator.


macxamin (approved):
LGTM

PR #129, opened 2022-01-28
Handle cluster location correctly

fabxc:
We previously incorrectly assumed we could label data in regional
cluster by the zone of the machine it originated from. That may
cause theoretical collisions however and we must always set location
to the cluster's official location.
This in practice makes our lives easier as we can simply enforce the
location label of all rule data to match the cluster location.

### DIFF
--- a/pkg/export/setup/setup.go
+++ b/pkg/export/setup/setup.go
@@ -73,9 +73,21 @@ func FromFlags(a *kingpin.Application, userAgent string) func(log.Logger, promet
 	// Default target fields if we can detect them in GCP.
 	if metadata.OnGCE() {
 		opts.ProjectID, _ = metadata.ProjectID()
-		// These attributes are set for GKE nodes.
 		opts.Location, _ = metadata.InstanceAttributeValue("cluster-location")
-		opts.Cluster, _ = metadata.InstanceAttributeValue("cluster-name")
+		// These attributes are set for GKE nodes. For the location, we first check
+		// the clustr location, which may be a zone or a region. We must always use that value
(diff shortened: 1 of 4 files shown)


andysim3d (approved):

pkg/export/setup/setup.go
```diff
@@ -73,9 +73,21 @@ func FromFlags(a *kingpin.Application, userAgent string) func(log.Logger, promet
...
+		// the clustr location, which may be a zone or a region. We must always use that value
+		// to avoid collisions with other clusters, as the same cluster name may be reused
+		// in different locations.
+		// In particular, we cannot set the location to the node's zone for a regional cluster,
+		// even though this would provide more accuracy, as there may also be a zonal cluster
+		// with the same name.
+		// We only fallback to the node zone as the location if no cluster location exists to
+		// default for deployments on GCE.
+		if loc, _ := metadata.InstanceAttributeValue("cluster-name"); loc != "" {
+			opts.Location = loc
```
pintohutch, line 87: @fabxc - why are you overwriting the location to be the `cluster-name` here? IIUC from the comments, I would've thought the logic as:
```go
// Check if we're on a GKE node. If so, grab the `cluster-location` tag.
if name, _ := metadata.InstanceAttributeValue("cluster-name"); name != "" {
  opts.Location, _ = metadata.InstanceAttributeValue("cluster-location")
// Otherwise fallback to node zone.
} else {
  opts.Location, _ = metadata.Zone()
}
```
but I may be missing something (and shaking off some cobwebs 😄 )

PR #131, opened 2022-02-04
Fix race condition in histogram building

fabxc:
The histograms used slices for bounds and values in the proto that were
originally allocated for a memory-pooled intermediate struct. The struct
is returned to the pool before the proto is sent.
Thus the next histogram overwrites the slices concurrently. Allocating
unique slices for the proto solves the issue.


andysim3d (approved):
LGTM on fix.

PR #134, opened 2022-02-04
Add webhook CA flag

macxamin:
Added ca-cert-base64 flag


pkg/operator/operator.go
```diff
@@ -346,7 +348,14 @@ func (o *Operator) ensureCerts(ctx context.Context, dir string) ([]byte, error)
...
 			return nil, errors.Wrap(err, "decoding TLS key")
 		}
-	} else if o.opts.TLSKey == "" && o.opts.TLSCert == "" {
+		caData, err = base64.StdEncoding.DecodeString(o.opts.CACert)
+		if err != nil {
+			return nil, errors.Wrap(err, "decoding certificate authority")
+		}
+		if err = ioutil.WriteFile(filepath.Join(dir, "ca.crt"), caData, 0666); err != nil {
+			return nil, errors.Wrap(err, "create certificate authority file")
+		}
```
fabxc, line 357: We don't need to write that file. There's nothing looking at it – we only need to return it to propagate it to the webhook config resource.

pkg/operator/operator_test.go
```diff
@@ -43,40 +43,55 @@ func TestEnsureCertsExplicit(t *testing.T) {
...
+			desc:         "input key/cert/ca",
+			opts:         Options{TLSKey: "a2V5", TLSCert: "Y2VydA==", CACert: "Y2FjZXJ0", OperatorNamespace: "test-ns"},
+			expectCert:   "cert",
+			expectKey:    "key",
+			expectCaCert: "cacert",
+			expectErr:    false,
+		},
+		{
+			desc:         "cert/key and no CA",
+			opts:         Options{TLSKey: "a2V5", TLSCert: "Y2VydA==", CACert: "Y2VydA==", OperatorNamespace: "test-ns"},
```
fabxc, line 63: This doesn't seem correct. The test case says no CA is set, but the options do have CACert have it set – just to the same value as TLSCert.

pkg/operator/operator_test.go
```diff
@@ -43,40 +43,55 @@ func TestEnsureCertsExplicit(t *testing.T) {
...
+			expectCert:   "cert",
+			expectKey:    "key",
+			expectCaCert: "cert", // if ca is not available, crt will be used as ca.
+			expectErr:    true,
```
fabxc, line 67: The test description doesn't seem like one where an error is expected?

pkg/operator/operator_test.go
```diff
@@ -85,13 +100,14 @@ func TestEnsureCertsExplicit(t *testing.T) {
 			}
```
fabxc, line 100: This misses a check for expectErr being true but err being nil. That's the reason why the above test which has an incorrect expectErr=true currently passes.

Same in other test.

pkg/operator/operator_test.go
```diff
@@ -139,6 +155,10 @@ func TestEnsureCertsSelfSigned(t *testing.T) {
 			if len(outCert) == 0 {
 				t.Errorf("expected generated cert but was empty")
 			}
+			// self-generate case ca is equal to crt.
+			if string(outCert) != string(caBundle) {
+				t.Errorf("want key: %v; got %v", string(outCert), string(caBundle))
```
fabxc, line 160: Error message refers to a key but that's not what's being checked here.

fabxc (approved):

PR #137, opened 2022-02-09
Provide webhook configs at deploy time, inject CA bundle from operator

fabxc:
Letting the user decide at deploy time which webhooks with which
confiugration to deploy gives a lot more flexibility.
We simply let the operator inject the CA bundle if the webhook configs
follow the expected naming pattern.

We delete the previously operator-created webhook configuration to
migrate old clusters.

The provided webhook config manifests switch the fail-closed failure
policies.


macxamin (approved):

pintohutch:
When would a user want to not use the full suite of webhooks?

fabxc:
Unlikely but possible. The more practical concern is users editing their
operator deployment (e.g. port, service) and then not having the means to
adjust the webhook config with the previous approach. And this allows to
remove one fairly critical permission from the operator as it could in
principle be used to install blocking webhook on a bunch of unrelated
resources.

PR #155, opened 2022-03-09
Support NO_DOCKER=1 for assets

fabxc:


andysim3d (approved):

Makefile
```diff
@@ -12,11 +12,6 @@ define docker_build
 	DOCKER_BUILDKIT=1 docker build $(1)
 endef
 
-define assets_diff
```
pintohutch, line 15: @fabxc any particular reason for stripping this out? It's a little hacky but I believe it works and can help speed up local build times.

PR #156, opened 2022-03-10
Fix queue backlog handling

fabxc:


pkg/export/shard.go
```diff
@@ -58,15 +58,15 @@ func (s *shard) enqueue(hash uint64, sample *monitoring_pb.TimeSeries) {
 
 // fill adds samples to the batch until its capacity is reached or the shard
 // has no more samples for series that are not in the batch yet.
-func (s *shard) fill(batch *batch) int {
+func (s *shard) fill(batch *batch) (took, remaining int) {
```
andysim3d, line 61: it seems unused?

andysim3d (approved):
LGTM to unblock p0 bug.

PR #164, opened 2022-03-21
Support authentication using an alternative gce token and quota attribution to a different project

yungdarek122:


pkg/export/setup/setup.go
```diff
@@ -129,6 +129,15 @@ func FromFlags(a *kingpin.Application, userAgent string) func(log.Logger, promet
 	a.Flag("export.debug.batch-size", "Maximum number of points to send in one batch to the GCM API.").
 		Default(strconv.Itoa(export.BatchSizeMax)).UintVar(&opts.BatchSize)
 
+	a.Flag("export.token-url", "The request URL to generate token that's needed to ingest metrics to an alternative project").
+		StringVar(&opts.TokenURL)
+
+	a.Flag("export.token-body", "The request Body to generate token that's needed to ingest metrics to an alternative project.").
```
fabxc, line 135: request body

Can you elaborate on the "alternative project" bit?

pkg/export/gce_token_source.go
```diff
@@ -0,0 +1,87 @@
...
+	"context"
+	"encoding/json"
+	"net/http"
+	"strings"
+	"time"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/googleapi"
+	"k8s.io/client-go/util/flowcontrol"
```
fabxc, line 28: Kubernetes dependencies usually explode and basically always require Go modules hackery to get things to compile. The fewer distinct places we introduce them, the better the odds that we find a constellation that actually compiles.

Is this dependency needed or can we find a good stand-alone alternative?

yungdarek122: Removed this dependency by using the limiter from golang.org/x/time/rate instead

pkg/export/setup/setup.go
yungdarek122: This bit is incorrect. Updated the text to "to generate token that's needed to ingest metrics to the project".

andysim3d (approved):
LGTM

PR #175, opened 2022-04-01
Fix presubmit.sh and add node selectors

pintohutch:
* Cleaned up some bad merge conflicts.
* Got `make ps` running. kindtest target still needs to be fixed.
* Add linux and amd64 node selectors for managed collection and example
  manifests.


pintohutch:
Extends the work proposed in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/167.

fabxc (approved):

PR #179, opened 2022-04-05
Set container security context

fabxc:
We need to setup an empty dir volume so non-root user can modify the respective dirs.


pintohutch (approved):
LGTM

PR #180, opened 2022-04-05
Specify automount of SA token everywhere.

pintohutch:
* The motivation to make this explicitly true everywhere was the [terraform provider](https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest/submodules/workload-identity#inputs) for WI-enabled GKE clusters defaults `automountServiceToken=false`, which prevents images from reaching necessary GCP endpoints (i.e. monitoring API). As the pod spec `automountServiceAccountToken` field [takes precedence](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server) over the service account's field, we set it everywhere we can to ensure access.
* Also automate deletion of extra CRD status field

### DIFF
--- a/pkg/operator/collection.go
+++ b/pkg/operator/collection.go
@@ -43,6 +43,10 @@ import (
 	monitoringv1alpha1 "github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator/apis/monitoring/v1alpha1"
 )
 
+var (
+	TrueVar = true
+)
+
 func setupCollectionControllers(op *Operator) error {
 	// The singleton OperatorConfig is the request object we reconcile against.
 	objRequest := reconcile.Request{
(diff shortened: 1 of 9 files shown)


pkg/operator/collection.go
```diff
@@ -43,6 +43,10 @@ import (
 	monitoringv1alpha1 "github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator/apis/monitoring/v1alpha1"
 )
 
+var (
+	TrueVar = true
```
fabxc, line 47: Shouldn't be exported. If you put it at the package level instead of function level, I'd also make it a function to ensure it cannot be modified.

pintohutch: Done.

fabxc (approved):

PR #182, opened 2022-04-06
add flag to add ua information

andysim3d:
customer could pass a ua information to operator, and operator will propagate ua information to collector and evaluator.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -88,7 +88,7 @@ func main() {
 		evaluatorMemoryResource = flag.Int64("evaluator-memory-resource", 200, "The Memory Resource of evaluator pod, in mega bytes.")
 		evaluatorMemoryLimit    = flag.Int64("evaluator-memory-limit", 1000, "The Memory Limit of evaluator pod, in mega bytesv.")
 		evaluatorCPUResource    = flag.Int64("evaluator-cpu-resource", 100, "The CPU Resource of evaluator pod, in milli cpu.")
-		operatorEnv = flag.String("opeator-env", "prod", "the environment operator runs on.")
+		operatorDeployment      = flag.String("opeator-deployment", "kubectl-deployed", "the environment operator runs on.")
 	)
 	flag.Parse()
 
@@ -129,7 +129,7 @@ func main() {
 		EvaluatorCPUResource:    *evaluatorCPUResource,
 		EvaluatorMemoryResource: *evaluatorMemoryResource,
 		EvaluatorMemoryLimit:    *evaluatorMemoryLimit,
-		OperatorEnv:             *operatorEnv,
+		OperatorDeployment:      *operatorDeployment,
 	})
 	if err != nil {
 		logger.Error(err, "instantiating operator failed")
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -100,15 +100,6 @@ func main() {
 		logger.Error(err, "loading kubeconfig failed")
 		os.Exit(1)
 	}
-	switch *mode {
-	// repo manifest always defaults to "kubectl".
-	case "kubectl":
-	case "gke":
-	case "gke-auto":
-	default:
-		logger.Error(err, "--mode must be one of {'kubectl', 'gcloud', 'gcloud-auto'}")
-		os.Exit(1)
-	}
 
 	// controller-runtime creates a registry against which its metrics are registered globally.
 	// Using it as our non-global registry is the easiest way to combine metrics into a single
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -88,6 +88,7 @@ func main() {
 		evaluatorMemoryResource = flag.Int64("evaluator-memory-resource", 200, "The Memory Resource of evaluator pod, in mega bytes.")
 		evaluatorMemoryLimit    = flag.Int64("evaluator-memory-limit", 1000, "The Memory Limit of evaluator pod, in mega bytesv.")
 		evaluatorCPUResource    = flag.Int64("evaluator-cpu-resource", 100, "The CPU Resource of evaluator pod, in milli cpu.")
+		operatorEnv = flag.String("opeator-env", "prod", "the environment operator runs on.")
 	)
 	flag.Parse()
 
@@ -128,6 +129,7 @@ func main() {
 		EvaluatorCPUResource:    *evaluatorCPUResource,
 		EvaluatorMemoryResource: *evaluatorMemoryResource,
 		EvaluatorMemoryLimit:    *evaluatorMemoryLimit,
+		OperatorEnv:             *operatorEnv,
 	})
 	if err != nil {
 		logger.Error(err, "instantiating operator failed")
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -103,8 +103,8 @@ func main() {
 	switch *mode {
 	// repo manifest always defaults to "kubectl".
 	case "kubectl":
-	case "gcloud":
-	case "gcloud-auto":
+	case "gke":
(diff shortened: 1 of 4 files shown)


pkg/operator/collection.go
```diff
@@ -263,6 +263,9 @@ func (r *collectionReconciler) makeCollectorDaemonSet(spec *monitoringv1alpha1.C
 	for _, matcher := range spec.Filter.MatchOneOf {
 		collectorArgs = append(collectorArgs, fmt.Sprintf("--export.match=%s", matcher))
 	}
+	if r.opts.OperatorEnv != "" {
+		collectorArgs = append(collectorArgs, fmt.Sprintf("--export.user-agent=Prometheus-Collector/%s/%s", Version, r.opts.OperatorEnv))
```
fabxc, line 267: Can we make sure that the UA is a variation of the regular one (https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/cmd/rule-evaluator/main.go#L84).

Like just reproduce what is there and then add environment in braces behind that?

cmd/operator/main.go
```diff
@@ -88,6 +88,7 @@ func main() {
 		evaluatorMemoryResource = flag.Int64("evaluator-memory-resource", 200, "The Memory Resource of evaluator pod, in mega bytes.")
 		evaluatorMemoryLimit    = flag.Int64("evaluator-memory-limit", 1000, "The Memory Limit of evaluator pod, in mega bytesv.")
 		evaluatorCPUResource    = flag.Int64("evaluator-cpu-resource", 100, "The CPU Resource of evaluator pod, in milli cpu.")
+		operatorEnv = flag.String("opeator-env", "prod", "the environment operator runs on.")
```
pintohutch, line 91: Instead of `prod`, aren't we more interested in:
- kubectl-managed
- gcloud-managed on GKE or GKE autopilot
- self-deployed
Or is this flag for a different purpose?

andysim3d: done.

cmd/operator/main.go
```diff
@@ -88,6 +88,7 @@ func main() {
 		evaluatorMemoryResource = flag.Int64("evaluator-memory-resource", 200, "The Memory Resource of evaluator pod, in mega bytes.")
 		evaluatorMemoryLimit    = flag.Int64("evaluator-memory-limit", 1000, "The Memory Limit of evaluator pod, in mega bytesv.")
 		evaluatorCPUResource    = flag.Int64("evaluator-cpu-resource", 100, "The CPU Resource of evaluator pod, in milli cpu.")
+		operatorDeployment      = flag.String("opeator-deployment", "kubectl-deployed", "the environment operator runs on.")
```
pintohutch, line 91: typo: `opeator`. Also the flag description should say something like "how the operator is deployed".

Naming thought: to avoid overloading the term deployment (e.g. k8s resource), what about, just `mode` or `strategy`?

Also, what do you think about asserting the flag to be an enum in code to prevent abuse? e.g.

```go

mode = flag.String("mode", "kubectl", "how managed collection was provisioned")

switch mode {
// repo manifest always defaults to "kubectl".
case "kubectl":
case "gcloud":
case "gcloud-auto":
default:
	logger.Error(err, "--mode must be one of {'kubectl', 'gcloud', 'gcloud-auto'}")
	os.Exit(1)
}
```

@lyanco any thoughts on naming the enums here? AFAICT from this change, user-agent would be something like (i think this is all combinations):
- `prometheus-engine-export/0.3.3 prometheus-collector/2.28.1-gmp.7 (mode:kubectl)`
- `prometheus-engine-export/0.3.3 prometheus-collector/2.28.1-gmp.7 (mode:gcloud)`
- `prometheus-engine-export/0.3.3 prometheus-collector/2.28.1-gmp.7 (mode:gcloud-auto)`
- `prometheus-engine-export/0.3.3 prometheus/2.28.1-gmp.7`
- `prometheus-engine-export/0.3.3 rule-evaluator/0.3.3 (mode:kubectl)`
- `prometheus-engine-export/0.3.3 rule-evaluator/0.3.3 (mode:gcloud)`
- `prometheus-engine-export/0.3.3 rule-evaluator/0.3.3 (mode:gcloud-auto)`
- `prometheus-engine-export/0.3.3 rule-evaluator/0.3.3`

pkg/operator/collection.go
```diff
@@ -263,6 +263,9 @@ func (r *collectionReconciler) makeCollectorDaemonSet(spec *monitoringv1alpha1.C
 	for _, matcher := range spec.Filter.MatchOneOf {
 		collectorArgs = append(collectorArgs, fmt.Sprintf("--export.match=%s", matcher))
 	}
+	if r.opts.OperatorDeployment != "" {
+		collectorArgs = append(collectorArgs, fmt.Sprintf("--export.user-agent=prometheus-collector/%s (deployment:%s)", Version, r.opts.OperatorDeployment))
```
pintohutch, line 267: We want this `Version` to be the Prometheus version (e.g. [`2.28.1-gmp.7`](https://github.com/GoogleCloudPlatform/prometheus/blob/release-2.28.1-gmp/cmd/prometheus/main.go#L325) in our prometheus fork) rather than the `prometheus-engine-export` version (which will already [be prepended](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/492b943b5edbf467798cbf616ceb408d3fc1d0c8/pkg/export/export.go#L216-L217) in the full user agent).

cmd/operator/main.go
lyanco: I'd probably replace "gcloud" with "gke" but besides that, looks great to me

andysim3d: done

cmd/operator/main.go
```diff
@@ -99,6 +100,15 @@ func main() {
 		logger.Error(err, "loading kubeconfig failed")
 		os.Exit(1)
 	}
+	switch *mode {
```
pintohutch, line 103: Only other thought I had about this, is that since it's being passed into the `operator.Options`, you could move this check to the [`defaultAndValidate`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/51e169c405cf3f88a7036bb778ebe718759166c2/pkg/operator/operator.go#L148) call.

pkg/operator/operator.go
```diff
@@ -80,6 +80,12 @@ const (
 	KubernetesAppName    = "app"
 	CollectorAppName     = "managed-prometheus-collector"
 	RuleEvaluatorAppName = "managed-prometheus-rule-evaluator"
+
+	// The Exporter version, will be exposed as part of the user agent information.
+	ExporterVersion = "0.3.3"
```
pintohutch, line 85: This should already be available to use in [`export.Version`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/51e169c405cf3f88a7036bb778ebe718759166c2/pkg/export/export.go#L451).

pkg/operator/operator.go
```diff
@@ -213,6 +218,14 @@ func (o *Options) defaultAndValidate(logger logr.Logger) error {
 	if o.EvaluatorMemoryLimit <= o.EvaluatorMemoryResource {
 		o.EvaluatorMemoryLimit = o.EvaluatorMemoryResource * 15
 	}
+	switch *mode {
```
pintohutch, line 221: I believe this won't compile. You would want `switch o.Mode` here.

cmd/operator/main.go
```diff
@@ -99,6 +100,15 @@ func main() {
...
+	// repo manifest always defaults to "kubectl".
```
pintohutch, line 104: Delete here as you moved it to the `defaultAndValidate` call.

pintohutch (approved):
LGTM!

PR #181, opened 2022-04-06
Publish v1 of our APIs

pintohutch:

PR #187, opened 2022-04-07
Introduce V1 API

pintohutch:
This is the same essential PR as GoogleCloudPlatform/prometheus-engine#181 except the commits are broken down for easier review.

1. a06da64 - preparing v1 resources for autogeneration
2. a5ba159 - autogenerate go k8s  `v1` client code and manifests
3. 9cb3bf3 - find/replace go code calls from v1alpha1 to v1
4. 9b57ea9 - remove go k8s `v1alpha1` client code

1 is probably the most interesting, followed by 3. The others are autogenerated or rote.

I tested this on a fresh cluster as well as upgrading an existing managed-collection setup.

Overall I removed the `v1alpha1` go k8s client code, but kept the types and kubebuilder tags. This ensures our CRD contains [both api versions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions) and the code is backwards-compatible with `v1alpha1` CRs.


fabxc:
What are the semantics for existing clusters and resources?
Are existing v1alpha1 resources processed by the operator as if they were v1 resources? Do they remain stored as v1alpha1 or upgraded to v1 automatically?

I recall there are conversion webhooks but am not sure whether they'd be necessary here as the versions are identical. What semantics would we have to commit to going forward, i.e. do we only ever add features/fields to v1?

pintohutch:
I found this [link](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects) particularly helpful in answering your questions.

> Are existing v1alpha1 resources processed by the operator as if they were v1 resources?

Yes. Note that this uses `v1` as the storage version ([example](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/187/files#diff-392a07b8949abcf49d6ef59c6bb1c4c391ae2e1bda1d0003add236fe46d7f21eL292)). From the docs:
```
When you read an object, you specify the version as part of the path.
If you specify a version that is different from the object's persisted version,
Kubernetes returns the object to you at the version you requested,
but the persisted object is neither changed on disk,
nor converted in any way (other than changing the apiVersion string) while serving the request.
You can request an object at any version that is currently served.
```
I validated this by `kubectl edit`-ing an existing `v1alpha1` PodMonitoring against an operator listening for `v1` CRs. I was able to trigger the validating webhook (I gave it some bad target labels and the webhook correctly rejected it).

> Do they remain stored as v1alpha1 or upgraded to v1 automatically?

They remain stored as `v1alpha1` unless updated. From the docs:
```
When an object is written, it is persisted at the version designated as the storage version at the time of the write.
If the storage version changes, existing objects are never converted automatically.
However, newly-created or updated objects are written at the new storage version.
It is possible for an object to have been written at a version that is no longer served.
```

> I recall there are conversion webhooks but am not sure whether they'd be necessary here as the versions are identical.

Correct. The default `None` conversion strategy works for us here - no webhooks necessary.

> What semantics would we have to commit to going forward, i.e. do we only ever add features/fields to v1?

I think committing to `v1` going forward makes sense. We can eventually deprecate `v1alpha1`, but no reason to rush into that any time soon.

Do you think it makes sense to keep the non-struct items for our `v1alpha1` types, e.g. methods and unit tests? I'm inclined to rip those out as the operator will only act on v1 CRs.

fabxc:
Thanks for the context. That all makes sense.
If we don't use these going forward, I'd say go ahead and delete them.

fabxc:
Can you mark the v1/ files as branched in git in the commit they were initially created? Will avoid lost updates if we do changes to v1alpha1 before this is merged.

pintohutch:
Closing in favor of GoogleCloudPlatform/prometheus-engine#199

PR #194, opened 2022-04-14
Refactor config generation, fix numeric port handling

fabxc:
The fix of numeric port handling may make certain configurations that worked before no longer work.
This is a necessary part of the bug fix unfortunately and we'll need to mention it in the release notes.

### DIFF
--- a/pkg/operator/apis/monitoring/v1alpha1/types.go
+++ b/pkg/operator/apis/monitoring/v1alpha1/types.go
@@ -335,35 +335,6 @@ func (status *PodMonitoringStatus) SetPodMonitoringCondition(gen int64, now meta
 const EnvVarNodeName = "NODE_NAME"
 
 func (pm *PodMonitoring) endpointScrapeConfig(index int) (*promconfig.ScrapeConfig, error) {
-	// The Prometheus configuration structs do not generally have validation methods and embed their
-	// validation logic in the UnmarshalYAML methods. To keep things reasonable we don't re-validate
-	// everything and simply do a final marshal-unmarshal cycle at the end to run all validation
-	// upstream provides at the end of this method.
-
-	// Configure how Prometheus talks to the Kubernetes API server to discover targets.
-	// This configuration is the same for all scrape jobs (esp. selectors).
-	// This ensures that Prometheus can reuse the underlying client and caches, which reduces
-	// load on the Kubernetes API server.
-	discoveryCfgs := discovery.Configs{
-		&discoverykube.SDConfig{
-			HTTPClientConfig: config.DefaultHTTPClientConfig,
-			Role:             discoverykube.RolePod,
-			// Drop all potential targets not the same node as the collector. The $(NODE_NAME) variable
-			// is interpolated by the config reloader sidecar before the config reaches the Prometheus collector.
-			// Doing it through selectors rather than relabeling should substantially reduce the client and
-			// server side load.
-			Selectors: []discoverykube.SelectorConfig{
-				{
-					Role:  discoverykube.RolePod,
-					Field: fmt.Sprintf("spec.nodeName=$(%s)", EnvVarNodeName),
-				},
-			},
-		},
-	}
-
-	ep := pm.Spec.Endpoints[index]
-
-	// TODO(freinartz): validate all generated regular expressions.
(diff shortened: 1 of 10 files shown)


pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -373,18 +344,82 @@ func (pm *PodMonitoring) endpointScrapeConfig(index int) (*promconfig.ScrapeConf
...
 
-	// Filter targets that belong to selected services.
+	// Filter targets that belong to selected pods.
+	selectors, err := relabelingsForSelector(pm.Spec.Selector)
+	if err != nil {
+		return nil, err
+	}
+	relabelCfgs = append(relabelCfgs, selectors...)
+
+	metadataLabels := map[string]struct{}{}
```
pintohutch, line 354: Does it make sense to capture this logic in its own function and re-use for PodMonitoring and ClusterPodMonitoring? e.g.
```go
func relabelingsForMetadata(labels ...string) []*relabel.Config  {
	// For loop over labels and use a switch statement to capture and create each rule.
}
```

And the checking the set of the metadata list can be done in each type before the function call, e.g. for PodMonitoring
```go
if metadata := *pm.Spec.TargetLabels.Metadata; metadata != nil {
	for _, l := range metadata {
		if allowed := []string{"pod", "container", "node"}; !containsString(allowed, l) {
			return nil, errors.Errorf("metadata label %q not allowed, must be one of %v", l, allowed)
		}
	}
	metadataRelabelings = relabelingsFromMetadata(metadata...)
	relabelCfgs = append(relabelCfgs, metadataRelabelings...)
}
```
Something similar would follow for ClusterPodMonitoring.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -510,28 +514,52 @@ func (pm *PodMonitoring) endpointScrapeConfig(index int) (*promconfig.ScrapeConf
...
+		// Some users do not declare ports but expect PodMonitoring's to work with a numeric port
+		// specified nonetheless.
+		//
+		// We can support this by hard-overriding the address with the respective numeric port. If a pod
+		// has multiple declared ports that will produce multiple identical targets from the candidates. If
+		// the container metadata label is added however, they may differ by the `container` target label
+		// and not be deduplicated. This will cause the same endpoint to be scraped as different targets.
+		// Thus, we must still filter by numeric port IFF the target candidate does contain port info.
+		//
+		// Summarized, PodMonitorings with numeric ports always work if a pod has no declared ports.
```
pintohutch, line 528: Should we echo this in the API docs by putting it [here](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/0df48f25673168ee01ca12c8d41114e93f58b43d/pkg/operator/apis/monitoring/v1alpha1/types.go#L1072) in the context of a `ScrapeEndpoint`?

pkg/operator/e2e/main_test.go
```diff
@@ -352,6 +352,7 @@ func testRuleEvaluatorDeployment(ctx context.Context, t *testContext) {
 				fmt.Sprintf("--export.label.location=%s", location),
 				fmt.Sprintf("--export.label.cluster=%s", cluster),
 				fmt.Sprintf("--query.project-id=%s", projectID),
+				"--export.user-agent=prometheus-engine-export/0.3.3 rule-evaluator/0.3.3 (mode:kubectl)",
```
pintohutch, line 355: Is it a big deal to change these with each tag? Or should the tests look more like:
```go
	fmt.Sprintf("--export.user-agent=%s/%s %s %s",
		export.ClientName,
		export.Version,
		fmt.Sprintf("rule-evaluator/%s", export.Version),
		"(mode: kubectl)",
	),
```

and

```go
	fmt.Sprintf("--export.user-agent=%s/%s %s %s",
		export.ClientName,
		export.Version,
		fmt.Sprintf("prometheus-collector/%s", operator.CollectorVersion),
		"(mode: kubectl)",
	),
```

fabxc: Yea it's a bit of a hassle I think but I figured the test setup may notably need to change if we do #195 – so I didn't want to optimize this in this PR yet.

PR #195, opened 2022-04-14
[DRAFT] Create DaemonSet externally and inject config

fabxc:
Change to an approach where the collector DaemonSet is provided
externally within the operator installation bundle. The operator then
only updates runtime-dependent values into the DaemonSet.
This allows users to do advanced customization (node affinity, priority
class, etc.) without the need for configuration knobs through the
OperatorConfig or operator flags.

This approach would be extended to the rule evaluator deployment if accepted.


pintohutch (commented):
LGTM % manual sanity check and `make presubmit`.

PR #198, opened 2022-04-15
Fix RBAC by removing resourceNames

pintohutch:
I noticed that a from-scratch kubectl install from `main` was throwing a lot of permissions issues. For example:
```
"level":"error","ts":1650043097.3943744,"logger":"controller.rules","msg":"Reconciler error","reconciler group":"monitoring.googleapis.com","reconciler kind":"OperatorConfig","name":"config","namespace":"gmp-public","error":"ensure rule configmaps: create generated rules: configmaps is forbidden: User \"system:serviceaccount:gmp-system:operator\" cannot create resource \"configmaps\" in API group \"\" in the namespace \"gmp-system\"","errorVerbose":"configmaps is forbidden: User \"system:serviceaccount:gmp-system:operator\" cannot create resource \"configmaps\" in API group \"\" in the namespace \"gmp-system\"\ncreate generated rules\ngithub.com/GoogleCloudPlatform/prometheus-engine/pkg/operator.(*rulesReconciler).ensureRuleConfigs\n\t/app/pkg/operator/rules.go:191\ngithub.com/GoogleCloudPlatform/prometheus-engine/pkg/operator.(*rulesReconciler).Reconcile\n\t/app/pkg/operator/rules.go:112
```

It turns out that using `resourceNames` with the `create` verb is ineffective. As a result, it seems these policy rules were dropped. From https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources:

> Note: You cannot restrict create or deletecollection requests by their resource name. For create, this limitation is because the name of the new object may not be known at authorization time. If you restrict list or watch by resourceName, clients must include a metadata.name field selector in their list or watch request that matches the specified resourceName in order to be authorized. For example, kubectl get configmaps --field-selector=metadata.name=my-configmap

I removed the `resourceNames` altogether and the operator was able to resume normal behavior.

### DIFF
--- a/cmd/operator/deploy/operator/03-role.yaml
+++ b/cmd/operator/deploy/operator/03-role.yaml
@@ -42,29 +42,18 @@ rules:
   resources:
   - secrets
   verbs: ["create", "patch", "update"]
-  resourceNames:
-  - collection
-  - rules
 - apiGroups: [""]
   resources:
   - configmaps
   verbs: ["create", "patch", "update"]
-  resourceNames:
-  - collector
-  - rule-evaluator
-  - rules-generated
 - apiGroups: ["apps"]
   resources:
   - daemonsets
   verbs: ["create", "patch", "update"]
-  resourceNames:
-  - collector
 - apiGroups: ["apps"]
   resources:
   - deployments
   verbs: ["create", "patch", "update"]
-  resourceNames:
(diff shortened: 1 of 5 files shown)


fabxc (approved):
Thanks for catching that.

Issue #197, opened 2022-04-14
Support querying cAdvisor in kubelet


willchan:
I'd like to configure scraping of the cAdvisor metrics endpoint in the kubelet: https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/#metrics-in-kubernetes. The prometheus docs have an example of how to do this: https://github.com/prometheus/prometheus/blob/ec3d02019e84d9d793d2e137891dd7ea6d19ea60/documentation/examples/prometheus-kubernetes.yml#L110. Per the doc, the scraping requires configuring the CA certificate and authorization via the service account token. AFAICT, PodMonitoring doesn't support either of these in its [ScrapeEndpoint](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.3.1/doc/api.md#scrapeendpoint).

pintohutch:
Hello,

We recently added support for cAdvisor scraping in GoogleCloudPlatform/prometheus-engine#203, which made it into our `0.4.0` release. Closing this issue.

pintohutch:
See https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.4.0/doc/api.md#kubeletscraping for specific usage.

willchan:
I tried using this feature but I don't seem to get any cadvisor metrics. I'm not sure if I'm doing something wrong. Is there a way to debug?

pintohutch:
Adding something like the following to your OperatorConfig. I just tried this myself manually and am seeing cadvisor/kubelet metrics being scraped.
```
collection:
  kubeletScraping:
    interval: 30s
```

Are you doing something like this?

willchan:
I believe so. This is what my OperatorConfig looks like in my kubespray managed single node k8s cluster running 1.21. I'm running the 0.4.0 GMP manifests for managed collection in this cluster.

```
$ kubectl -n gmp-public get operatorconfig -o yaml
apiVersion: v1
items:
- apiVersion: monitoring.googleapis.com/v1
  collection:
    credentials:
      key: key.json
      name: cluster-gcp-service-account
    kubeletScraping:
      interval: 5m
  kind: OperatorConfig
  metadata:
    creationTimestamp: "2022-03-11T01:43:01Z"
    generation: 4
    labels:
      kustomize.toolkit.fluxcd.io/name: infrastructure
      kustomize.toolkit.fluxcd.io/namespace: flux-system
    name: config
    namespace: gmp-public
[...]

PR #203, opened 2022-04-20
Scrape kubelet metrics

fabxc:


pintohutch:
Fixes GoogleCloudPlatform/prometheus-engine#197.

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -475,6 +483,27 @@ func endpointScrapeConfig(id string, ep ScrapeEndpoint, relabelCfgs []*relabel.C
 		},
 	}
 
+	// Use the pod name as the primary identifier in the instance label. Unless the pod
+	// is controlled by a DaemonSet, in which case the node name will be used.
+	// This provides a better user experience on dashboards which template on the instance label
```
pintohutch, line 488: So if there are two DaemonSets being monitored:
- `example-ds1` with labels `app=ds1`, `app.kubernetes.io/managed-by=example-controller` emitting on port name `metrics`
- `example-ds2` with labels `app=ds2`, `app.kubernetes.io/managed-by=example-controller` emitting on port name `metrics`

Something like `instance=k8s-node-1:metrics` would capture both targets, with the `job` target label being different, assuming the two scrape configs were generated by two PodMonitorings; one scraping using selector `app=ds1`, and the other `app=ds2`.

However, if there is one PodMonitoring that captures both DaemonSets, i.e. by using selector `app.kubernetes.io/managed-by=example-controller`, then would the targets be conflicting and merged together somehow? Or am I overlooking something? Maybe `honor_labels` disambiguation takes over here?

Also, if this relabling is ok, then this seems like behavior we should surface in our api documentation. What do you think about adding a note to https://github.com/GoogleCloudPlatform/prometheus-engine/blob/b9b52f970feb4788eaf4c36c7f04f02d19dc5007/pkg/operator/apis/monitoring/v1/types.go#L857-L858 and https://github.com/GoogleCloudPlatform/prometheus-engine/blob/b9b52f970feb4788eaf4c36c7f04f02d19dc5007/pkg/operator/apis/monitoring/v1/types.go#L886-L887 something like:

```
If the label selector selects pods controlled by a DaemonSet, the generated `instance` label will contain the Kubernetes node name of the embedded pod rather than the pod name itself.
```

pkg/operator/e2e/main_test.go
```diff
@@ -764,7 +768,96 @@ func validateCollectorUpMetrics(ctx context.Context, t *testContext, job string)
...
+				return true, nil
+			}, ctx.Done())
+			if err != nil {
+				t.Fatalf("Waiting for collector metrics to appear in Cloud Monitoring failed: %s", err)
+			}
+		}
+	}
+}
+
+// testCollectorSelfPodMonitoring sets up pod monitoring of the collector itself
```
pintohutch, line 803: nit: godoc incorrect.

pkg/operator/e2e/main_test.go
```diff
@@ -764,7 +768,96 @@ func validateCollectorUpMetrics(ctx context.Context, t *testContext, job string)
...
+	}
+
+	// Wait for metric data to show up in Cloud Monitoring.
+	metricClient, err := gcm.NewMetricClient(ctx)
+	if err != nil {
+		t.Fatalf("Create GCM metric client: %s", err)
+	}
+	defer metricClient.Close()
+
+	pods, err := t.kubeClient.CoreV1().Pods(t.namespace).List(ctx, metav1.ListOptions{
```
pintohutch, line 827: Here we're not explicitly interested in collector pods, but rather all nodes in the cluster. I understand the end result in the test logic below is the same, but to be explicit, can we change this to fetching via the kube node list api instead to make that clear?

pkg/operator/apis/monitoring/v1/types.go
fabxc: You are right, good analysis. This could indeed be a conflict. By default we'd still add the `pod` label however, which would distinguish them again.

In this particular case I'd say that the benefit of `instance` having a more meaningful value outweighs the collision risk, which is fairly minimal I think and contained within a single PodMonitorig, i.e. not users affecting each other.

fabxc: Docs added on `TargetLabels` as I suspect that's where users will look to understand how they are created.

pintohutch (approved):

PR #207, opened 2022-04-28
Update Prometheus dependency

fabxc:


pkg/ui/assets_embed.go
```diff
@@ -0,0 +1,26 @@
...
+
+package ui
+
+import (
+	"net/http"
+
+	"github.com/prometheus/common/assets"
+)
+
+var Assets = http.FS(assets.New(EmbedFS))
```
pintohutch: My IDE isn't able to let me see what this stuff is. Are you able to navigate around here? I'm not able to find `EmbedFS` anywhere. I imagine this is some autogenerated byte payload in a go file somewhere...

In general, I'm having a hard time remembering this build flow. Could you ensure https://github.com/GoogleCloudPlatform/prometheus-engine/blob/freinartz/updateprometheus/pkg/ui/README.md#L26 is up-to-date with a lite description of the build flow?

third_party/prometheus_ui/override/src/App.tsx
```diff
@@ -5,26 +5,26 @@ import React, { FC } from 'react';
 import Navigation from './Navbar';
```
pintohutch, line 5: Do the permalinks need to be updated at the top here?

pkg/export/bench/.gitingore
```diff
@@ -0,0 +1,3 @@
+data1
```
pintohutch, line 1: What are these? Can you add a comment? IIUC they're local TSDB dirs...

hack/update-ui.sh
```diff
@@ -44,11 +44,13 @@ shopt -s extglob
...
+  scripts/!(compress_assets.sh)
 
 # Ensure empty target dir is committed.
 touch web/ui/static/.gitignore
 
-# Download modules according to lock file.
-cd web/ui/react-app && yarn install --frozen-lockfile
\ No newline at end of file
+# Install node_modules for vendoring.
+make ui-install
```
pintohutch, line 56: Did you run `./hack/update-ui.sh v2.35.0` manually to bring in some of these changes?

go.mod
```diff
@@ -1,136 +1,134 @@
 module github.com/GoogleCloudPlatform/prometheus-engine
 
-go 1.17
+go 1.16
```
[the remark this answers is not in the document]
fabxc: I had to bump this in the past because otherwise something about Go modules kept failing.
Now I could move it back to the Go version also defined in the Prometheus version we depend on. It's just the minimum required Go version, so this will allow building from more environments.

third_party/prometheus_ui/override/src/Navbar.tsx
```diff
@@ -2,26 +2,24 @@
 // License: https://github.com/prometheus/prometheus/blob/1cffda5de775668f7438578c90a05e8bfa58eb91/LICENSE
```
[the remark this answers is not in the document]
fabxc: Mh, don't think so – initially forked from the same file and basically hand-adjusted for this update.

pkg/export/bench/.gitingore
fabxc: Yes, TSDB dirs created when running the benchmarking script.

hack/update-ui.sh
fabxc: Yes

pkg/ui/assets_embed.go
fabxc: I was confused as well. Embedding is now a new-ish feature of Go itself. So it happens when compiling not by pre-generating a file containing all embedded contents like before.
The generated `embed.go` file merely contains a list of files to include.

README has indeed been stale for a bit. Updated it and removed a lot of stuff that's not really relevant for our development flow anymore.

go.mod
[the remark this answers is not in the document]
pintohutch: Ok - do we need to downgrade the docker image use to build our binaries to [`golang:1.16-buster`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/a7e0126d79a63715f92e63e7d5103462d42f98b8/cmd/frontend/Dockerfile#L2)?

pkg/ui/README.md
```diff
@@ -12,36 +12,19 @@ the upstream UI at a given version and apply a minimal set of overrides files.
...
-```bash
-./sync.sh yarn start
-```
-
-The webserver will proxy requests to `http://localhost:9090`. During development a local Prometheus
-server or the frontend can be run under this address for manual testing.
-
-At build time the app is statically compiled into the `ui` Go package. Binaries including the UI
-must also host a Prometheus-compatible read API as expected by the React app.
+The `hack/update-ui.sh <PROMETHEUS_TAG>` script sync the upstream UI into
```
pintohutch, line 15: nit: syncs

pkg/ui/assets_embed.go
pintohutch: Thanks

go.mod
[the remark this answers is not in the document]
fabxc: No, it's just the minimum required version. We can build with anything higher.

pintohutch (approved):

PR #211, opened 2022-04-28
Remove v1alpha1 types logic

pintohutch:
Kept the v1alpha1 types for backwards compatibility, but we don't need the methods on them anymore.


pintohutch:
`controller-gen` [uses](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/567e207ac639cafa62cc280a3181acb04a410e93/hack/presubmit.sh#L89) those structs to refresh the CRD yamls. If we delete the go structs, then those CRDs are removed from the yamls and we lose backwards compatibility.

PR #213, opened 2022-05-02
Support generator URL in rule-evaluator

fabxc:


cmd/rule-evaluator/main.go
```diff
@@ -92,6 +93,9 @@ func main() {
 		Default(fmt.Sprintf("https://monitoring.googleapis.com/v1/projects/%s/location/global/prometheus", projectIDVar)).
 		String()
 
+	generatorURLStr := a.Flag("query.generator-url", "The base URL used for the generator URL in the alert notification payload. Should point to an instance of a query frontend that accesses the same data as --query.target-url.").
```
pintohutch, line 96: Is that more correct or helpful than suggesting `https://monitoring.googleapis.com/v1/projects/PROJECT_ID/location/global/prometheus/api/v1/`?

fabxc: The generator URL generally points at a Prometheus UI, rather than the API. It's meant to quickly get human users from an alert to a graph. That why we'll also only support the frontend (i.e. Prometheus OSS UI) for now as that's what the link is generated for.

pintohutch: Ok - SGTM.

pintohutch (approved):

PR #218, opened 2022-05-09
Use port 443 for webhook (operator) service

pintohutch:
This is more to follow convention. This is typical of workloads running on GKE clusters.


fabxc (approved):
This won't cause issues on update depending on apply order?

pintohutch:
> This won't cause issues on update depending on apply order?

@fabxc - can you elaborate? I manually tested an upgrade via `kubectl apply`ing the setup and operator manifests and everything seemed ok.

PR #224, opened 2022-05-13
add seccompProfile on operator created deployment.

andysim3d:
required by GKE for security checking.


pintohutch (changes requested):
We'll want this on the rule-evaluator [deployment](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/25bfac566cdceac809a46c40fb8975405b91c0b3/pkg/operator/operator_config.go#L524) too.

pintohutch (approved):
Oh I'm sorry - nevermind.

fabxc (changes requested):
Could/should this rather be in the per-container security context?

Some options can only be set there so we went with specifying with per-container ones. Keeping things in one place is probably easier?

But also, this is probably not worth merging since #227 will make this obsolete and I can add it there right away.

fabxc:
Added this to #227 as a separate commit and also added it to the operator deployment.

PR #227, opened 2022-05-16
Make collector/rule-evaluator part of install bundle

fabxc:
Change to an approach where the collector DaemonSet and rule-evaluator
Deployment are provided externally within the operator installation bundle.
The operator then only updates runtime-dependent values.
This allows users to do advanced customization (node affinity, priority
class, etc.) without the need for configuration knobs through the
OperatorConfig or operator flags.


pkg/operator/operator_config.go
```diff
@@ -296,242 +293,64 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorSecrets(ctx context.Contex
...
-	if err := r.client.Patch(ctx, deploy, client.Merge); apierrors.IsNotFound(err) {
-		if err := r.client.Create(ctx, deploy); err != nil {
-			return errors.Wrap(err, "create rule-evaluator deployment")
-		}
-	} else if err != nil {
-		return errors.Wrap(err, "update rule-evaluator deployment")
+	var deploy appsv1.Deployment
+	err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: NameRuleEvaluator}, &deploy)
+	// Some users deliberately not want to run the rule-evaluator. Only emit a warning but don't cause
+	// retries as this logic gets re-triggered anyway if the DaemonSet is created later.
```
pintohutch, line 301: nit: Deployment is created later

pkg/operator/operator_config.go
```diff
@@ -296,242 +293,64 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorSecrets(ctx context.Contex
...
-			Name:      NameRuleEvaluator,
-		},
-		Spec: deploy,
+		deploy.Spec.Template.Spec.Containers[i].Env = repl
 	}
+
+	// Upsert rule-evaluator Deployment.
+	// We've to use Patch() as Update() with update we'll get stuck in an infinite loop as each
+	// update increases the generation since we override the managed fields metadata, which is
+	// set by the kube-controller-manager, which wants to own an annotation on deployments.
```
pintohutch, line 352: Update changed which field in particular? `observedGeneration` (owned by `kube-controller-manager`). `.metadata.generation` doesn't show up as a `managedFields` in the YAML expansion.

The only thing we're updating here is the `.spec.containers[*].env` field, which the `kube-controller-manager` doesn't own...

pkg/operator/e2e/context.go
```diff
@@ -90,7 +92,7 @@ func newTestContext(t *testing.T) *testContext {
 	t.Cleanup(cancel)
 	t.Cleanup(func() { tctx.cleanupNamespaces() })
 
-	tctx.ownerReferences, err = tctx.createBaseResources()
+	tctx.ownerReferences, err = tctx.createBaseResources(context.TODO())
```
pintohutch, line 95: Never really understood when to use this vs `context.Background()`.. any ground rules you go by?

cmd/operator/main.go
```diff
@@ -62,33 +62,12 @@ func main() {
 		publicNamespace = flag.String("public-namespace", operator.DefaultPublicNamespace,
 			"Namespace in which the operator reads user-provided resources.")
 
-		imageCollector = flag.String("image-collector", operator.ImageCollector,
```
pintohutch, line 65: Are you worried we'll be breaking any customers' existing setups by removing these flags? Obviously wouldn't be an issue on GKE clusters using the gcloud integration.

I guess those who are deploying via `kubectl` themselves would use our manifests as a guide. But I'm sure they adapt those for their specific cases.

Do you think being loud enough in our release notes about the flag removals is sufficient?

pkg/operator/e2e/main_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"sort"
+	// "sort"
```
pintohutch, line 27: nit: commented out package

pintohutch (commented):
Are `EXTRA_ARGS` prioritized over their plain flag counterparts? Can we document that behavior somewhere?

cmd/operator/main.go
fabxc: I think so. The operator is in v0 and that primarily covers flags as our CRD API is versioned seperately. So users should always read release notes.

For upgrading, one interesting point is probably that the operator may override the YAML manifests again before being upgraded itself. So some extra caution is probably in order for kubectl-installed setups.

pkg/operator/e2e/context.go
fabxc: There's no technical difference besides Background being permanent while TODO indicating this should be connected to a higher level context eventually.

cmd/operator/deploy/operator/05-deployment.yaml
```diff
@@ -65,4 +65,6 @@ spec:
             drop:
             - all
           runAsUser: 1000
-          runAsGroup: 1000
\ No newline at end of file
+          runAsGroup: 1000
+          seccompProfile:
```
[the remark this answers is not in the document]
fabxc: Several things cannot be set of the Pod level.

pkg/operator/e2e/collector.yaml
```diff
@@ -0,0 +1 @@
+../../../cmd/operator/deploy/operator/10-collector.yaml
```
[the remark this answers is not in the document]
fabxc: This is just GitHubs way of displaying softlinks.

pkg/operator/collection.go
```diff
@@ -198,253 +197,53 @@ func (r *collectionReconciler) ensureCollectorSecrets(ctx context.Context, spec
...
-		LabelAppName:      NameCollector,
-		KubernetesAppName: CollectorAppName,
-	}
+	logger, _ := logr.FromContext(ctx)
 
-	podAnnotations := map[string]string{
-		AnnotationMetricName: componentName,
+	var ds appsv1.DaemonSet
+	err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: NameCollector}, &ds)
+	// Some users deliberately not want to run the collectors. Only emit a warning but don't cause
```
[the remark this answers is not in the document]
fabxc: I think we had one user ask about disabling rule evaluators altogether. One could imagine the same thing for collectors if only rule eval is relevant (certainly seems less relevant though).

cmd/operator/main.go
```diff
@@ -62,33 +62,12 @@ func main() {
 		publicNamespace = flag.String("public-namespace", operator.DefaultPublicNamespace,
```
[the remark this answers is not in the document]
fabxc: It's certainly a more involved change now. However, it seems overall consistent: if you change where you deploy the manifests into, you also update the namespace flags in the operator YAML. And as you say, this would just be templated.

Added the explicit flag values to the operator.

pkg/operator/operator_config.go
fabxc: Right the issue may no longer be relevant as we're only setting a small set of fields. I just didn't want to change too many things at once honestly.

pkg/operator/collection.go
```diff
@@ -198,253 +197,53 @@ func (r *collectionReconciler) ensureCollectorSecrets(ctx context.Context, spec
...
+		for _, ev := range c.Env {
+			if ev.Name != "EXTRA_ARGS" {
+				repl = append(repl, ev)
+			}
+		}
+		repl = append(repl, corev1.EnvVar{Name: "EXTRA_ARGS", Value: strings.Join(flags, " ")})
+
+		ds.Spec.Template.Spec.Containers[i].Env = repl
 	}
+	return r.client.Update(ctx, &ds)
```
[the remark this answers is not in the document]
fabxc: No, for the rule-evaluator the doc comment says we use patch due to an annotation being added by the API server to Deployments. That doesn't apply to DaemonSets.

cmd/operator/deploy/operator/10-collector.yaml
```diff
@@ -0,0 +1,154 @@
...
+      - name: storage
+        emptyDir: {}
+      - name: config
+        configMap:
+          name: collector
+      - name: config-out
+        emptyDir: {}
+      - name: rules-secret
+        secret:
+          secretName: collection
```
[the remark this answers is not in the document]
fabxc: That sounds sensible? I think this can be done in a follow-up without adverse effects as this change is already fairly big. (Probably also should come with further pruning RBACs on those secerts etc.)

pkg/operator/e2e/collector.yaml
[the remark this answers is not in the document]
pintohutch: Well I assume `ioutil` also works with this...? https://github.com/GoogleCloudPlatform/prometheus-engine/blob/freinartz/yamlcollector/pkg/operator/e2e/context.go#L223

pkg/operator/collection.go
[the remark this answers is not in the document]
pintohutch: Makes sense.

cmd/operator/deploy/operator/11-rule-evaluator.yaml
```diff
@@ -0,0 +1,143 @@
...
+      nodeSelector:
+        kubernetes.io/arch: amd64
+        kubernetes.io/os: linux
+      containers:
+      - name: evaluator
+        image: gke.gcr.io/prometheus-engine/rule-evaluator:v0.4.1-gke.0
+        args:
+        - --config.file=/prometheus/config_out/config.yaml
+        - --web.listen-address=:19092
+        - --export.user-agent=rule-evaluator/0.4.0 (mode:kubectl)
```
pintohutch, line 42: Bump user agent

cmd/operator/deploy/operator/05-deployment.yaml
```diff
@@ -43,6 +43,7 @@ spec:
       - name: operator
         image: gke.gcr.io/prometheus-engine/operator:v0.4.1-gke.0
         args:
+        - "--operator-namespace=gmp-system"
         - "--public-namespace=gmp-public"
         - "--priority-class=gmp-critical"
         - "--image-collector=gke.gcr.io/prometheus-engine/prometheus:v2.28.1-gmp.7-gke.0"
```
pintohutch, line 49: I guess these flags are still valid until we publish a new image version.

Bump the image-collector to latest?

fabxc: Yea, but of temporary  drift. For that version of the operator our canonical version didn't change yet basically. So I'd say this just gets updated with the v0.5.0 release?

pkg/operator/e2e/collector.yaml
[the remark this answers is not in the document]
fabxc: The e2e tests work as intended, so yes I'd say.

cmd/operator/deploy/operator/05-deployment.yaml
pintohutch: Ah makes sense.

manifests/operator.yaml
```diff
@@ -202,6 +202,7 @@ spec:
       - name: operator
         image: gke.gcr.io/prometheus-engine/operator:v0.4.1-gke.0
         args:
+        - "--operator-namespace=gmp-system"
         - "--public-namespace=gmp-public"
         - "--priority-class=gmp-critical"
         - "--image-collector=gke.gcr.io/prometheus-engine/prometheus:v2.28.1-gmp.7-gke.0"
```
courageJ, line 208: Should these be removed?

manifests/operator.yaml
```diff
@@ -202,6 +202,7 @@ spec:
       - name: operator
         image: gke.gcr.io/prometheus-engine/operator:v0.4.1-gke.0
         args:
+        - "--operator-namespace=gmp-system"
         - "--public-namespace=gmp-public"
         - "--priority-class=gmp-critical"
```
courageJ, line 207: I don't think this is still supported, right?

pintohutch: It's still [supported](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/freinartz/yamlcollector/cmd/operator/main.go#L60-L61) and relevant for future templating - i.e. deploying all resources in a particular namespace.

manifests/operator.yaml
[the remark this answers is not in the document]
pintohutch: The image flags are still necessary for the 0.4.1 operator. When we merge this and do an official release, we'll update the manifests to remove these.

PR #234, opened 2022-05-25
Enforce project_id, location, and cluster labels

fabxc:


pkg/operator/collection.go
```diff
@@ -523,7 +523,7 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 			Type:   monitoringv1.ConfigurationCreateSuccess,
 			Status: corev1.ConditionTrue,
 		}
-		cfgs, err := pmon.ScrapeConfigs()
+		cfgs, err := pmon.ScrapeConfigs(r.opts.ProjectID, r.opts.Location, r.opts.Cluster)
```
pintohutch, line 526: So IIUC we want to prefer the `OperatorConfig.Collection.ExternalLabels` over operator args, so we want something like:
```go
var projectID = r.opts.ProjectID
if p, ok := spec.ExternalLabels["project_id"]; ok {
	projectID = p
}
...
// etc for location, cluster
cfgs, err := pmon.ScrapeConfigs(projectID, location, cluster)
```

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -486,13 +491,29 @@ func endpointScrapeConfig(id string, ep ScrapeEndpoint, relabelCfgs []*relabel.C
...
 	}
 
-	// Use the pod name as the primary identifier in the instance label. Unless the pod
-	// is controlled by a DaemonSet, in which case the node name will be used.
-	// This provides a better user experience on dashboards which template on the instance label
-	// and expect it to have meaningful value, such as common node exporter dashboards.
-	//
-	// Save the value in a temporary label and use it further down.
 	relabelCfgs = append(relabelCfgs,
+		// Force target labels so they cannot be overwritten by metric labels.
```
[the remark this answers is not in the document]
fabxc: Then they need to duplicated for PodMonitoring and ClusterPodMonitoring though.

pkg/operator/collection.go
fabxc: Yes, that'd also go for `OperatorConfig.Rules.ExternalLabels` when generating the rule files.

That the `--export.label.*` flag, which we set from the operator's flags, will be overriden by the external labels is already ensured by the precedence change from our last release. So it only needs to be considered in the generated rule and scrape configs.

That's worth doing in a follow up I think.

pintohutch: follow-up in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/248

PR #244, opened 2022-06-09
Remove legacy KSM deployment

pintohutch:
Add in HPA


examples/kube-state-metrics.yaml
```diff
@@ -259,6 +259,36 @@ rules:
   - list
   - watch
 ---
+apiVersion: autoscaling/v2beta2
```
xichen2020, line 262: Why not `autoscaling/v2`?

pintohutch: Only available on 1.23+ clusters - https://kubernetes.io/blog/2021/12/07/kubernetes-1-23-release-announcement/#horizontalpodautoscaler-v2-graduates-to-ga

xichen2020 (approved):
LGTM

PR #248, opened 2022-06-10
Prioritize operator config label set

pintohutch:
This is a follow-up of the discussion from https://github.com/GoogleCloudPlatform/prometheus-engine/pull/234#discussion_r883224189.

### DIFF
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -110,17 +110,45 @@ func newRulesReconciler(c client.Client, opts Options) *rulesReconciler {
 	}
 }
 
-func (r *rulesReconciler) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	logger, _ := logr.FromContext(ctx)
 	logger.Info("reconciling rules")
 
-	if err := r.ensureRuleConfigs(ctx); err != nil {
+	var config monitoringv1.OperatorConfig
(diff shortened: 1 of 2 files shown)


pkg/operator/rules.go
```diff
@@ -257,7 +267,7 @@ type rulesValidator struct {
 }
 
 func (v *rulesValidator) ValidateCreate(ctx context.Context, o runtime.Object) error {
-	_, err := generateRules(o.(*monitoringv1.Rules), v.opts)
+	_, err := generateRules(o.(*monitoringv1.Rules), "test_project", "test_location", "test_cluster")
```
TheSpiritXIII, line 270: Is this right?

pintohutch: Yes. So it's hard to tell what's happening here, but `rulesValidator` [implements](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/pintohutch/opcfg-ext-labels/pkg/operator/operator.go#L361-L363) the [`admission.CustomValidator`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/pintohutch/opcfg-ext-labels/vendor/sigs.k8s.io/controller-runtime/pkg/webhook/admission/validator_custom.go#L30-L31) interface from the controller-runtime library. i.e. this method is executed as part of the admission webhooks for our `Rules` CRD. 

So when a new `Rule` is created, we essentially sanity-check it here with toy `project`, `location`, and `cluster` labels to ensure it can generate valid Prometheus rules. Thus, we only care if it returns and error, and we throw away the generated rules value.

TheSpiritXIII (approved):

PR #250, opened 2022-06-13
Validate location not set to global which is unsupported

TheSpiritXIII:


pintohutch:
I forgot to mention - we can add this check to the [export package](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/export.go#L351-L353) as well. This guards against setting the `location` label using external labels or self-deployed collectors.

pintohutch:
Merging this as is - will add the check in a follow-up PR.

PR #260, opened 2022-06-21
Clarify location flag and help text

lyanco:


cmd/operator/main.go
```diff
@@ -55,7 +55,7 @@ func main() {
 	var (
 		logVerbosity      = flag.Int("v", 0, "Logging verbosity")
 		projectID         = flag.String("project-id", defaultProjectID, "Project ID of the cluster. May be left empty on GKE.")
-		location          = flag.String("location", defaultLocation, "GCP location of the cluster. Maybe be left empty on GKE.")
+		location          = flag.String("location", defaultLocation, "Google Cloud region where your data will be stored. May be left empty on GKE.")
```
xichen2020, line 58: IIRC this is not necessarily a region. The specific example is that there may be a GKE regional cluster and a GKE zonal cluster both with the same name in the same project and the same region, and the location for the zonal cluster will need to be the zone to distinguish metrics for these two clusters.

PR #265, opened 2022-06-23
Move collector and rule-evaluator to yaml

pintohutch:
Same as https://github.com/GoogleCloudPlatform/prometheus-engine/pull/227 but CLA covers myself as the contributor.


cmd/operator/deploy/operator/10-collector.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - name: storage
+        emptyDir: {}
+      - name: config
+        configMap:
+          name: collector
+      - name: config-out
+        emptyDir: {}
+      - name: collection-secret
+        secret:
+          secretName: collection
```
damemi, line 161: nit: no newline at end of file

pkg/operator/collection.go
```diff
@@ -198,259 +193,55 @@ func (r *collectionReconciler) ensureCollectorSecrets(ctx context.Context, spec
...
-				PriorityClassName:            r.opts.PriorityClass,
-				SecurityContext:              podSpecSecurityContext(),
-			},
-		},
-	}
-	// DNS policy should be set explicitly when using hostNetwork.
-	if r.opts.HostNetwork {
-		ds.Template.Spec.HostNetwork = true
-		ds.Template.Spec.DNSPolicy = "ClusterFirstWithHostNet"
-	}
```
damemi, line 445: don't see these in the new collector manifest, just checking that's intentional

cmd/operator/deploy/operator/10-collector.yaml
pintohutch: Done.

pkg/operator/collection.go
pintohutch: Yeah - we don't have a reason to run on the host network anymore. We used to provide the option via the operator, but with collectors being broken out to their own manifests, users can add that themselves if they need it.

PR #268, opened 2022-06-28
Add flag to enable gzip compression on export

damemi:
ref: b/232423369


damemi:
Closing this, discussed with @pintohutch and we want to take a different direction. But it will be easier to just start a new PR than rebase these changes off of https://github.com/GoogleCloudPlatform/prometheus-engine/pull/265 (once that merges)

PR #270, opened 2022-07-06
[CI/CD/E2E] Add a fake metrics server for tests using a Kind cluster

realschwa:
This pull requests does the following:

1. Adds a new `FakeMetricsServer` that acts as a drop in replacement for the [endpoints listed here](https://cloud.google.com/monitoring/api/ref_v3/rpc#google.monitoring.v3.metricservice). This PR only implements the `CreateTimeSeries` endpoint. A further PR will implement `ListTimeSeries`.
2. Adds unit tests to ensure the `FakeMetricsServer` is behaving as expected.

This is the first of a few PRs that will create a robust testing solution to verify GMP behavior with tests using a local kind cluster and E2E tests using a real GKE cluster.


pkg/e2e/fake_metrics_server_test.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	fms.CreateTimeSeries(context.TODO(), createTimeSeriesRequest)
+
+	// these are the subtests
+	tests := []*createTimeSeriesTest{
+		{testName: "TestNil"},
+		{
+			testName: "TestExceedMaxTimeSeriesPerRequest",
+			createTimeSeriesRequest: &monitoringpb.CreateTimeSeriesRequest{
+				Name:       "PROJECT/1234",
```
andysim3d, line 90: "projects/1234"

pkg/e2e/fake_metrics_server_test.go
```diff
@@ -0,0 +1,315 @@
...
+		Name:       "PROJECT/1234",
+		TimeSeries: timeSeries,
+	}
+	fms.CreateTimeSeries(context.TODO(), createTimeSeriesRequest)
+
+	// these are the subtests
+	tests := []*createTimeSeriesTest{
+		{testName: "TestNil"},
+		{
+			testName: "TestExceedMaxTimeSeriesPerRequest",
```
andysim3d, line 88: it says test exceed max time series per request, but the ts is empty?

pkg/e2e/fake_metrics_server_test.go
```diff
@@ -0,0 +1,315 @@
...
+			if err == nil && response != nil {
+				t.Errorf("expected an error for %q", test.testName)
+			}
+		})
+	}
+}
+
+func TestCreateTimeSeries(t *testing.T) {
+	fms := NewFakeMetricServer(200)
+	projectName := "PROJECT/1234"
```
andysim3d, line 173: keep the style consistent, either in previous test, use a project name instead of hardcoded string, or in this test, use the hardcoded string per test case.

pkg/e2e/fake_metrics_server_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	// these are the subtests
+	tests := []*createTimeSeriesTest{
```
andysim3d, line 176: I'll suggest change this into one test case structure because each cases depended on previous: new project, existing project, existing project existing time series. I guess most ppl have the assumption that each test case is independent with each other. So if you want dependency to mock a stateful service, make 3 requests in one test case.

pkg/e2e/fake_metrics_server_test.go
realschwa: yeah the check for the not exceeding the max happens before we process anything so I kept them empty for brevity. I can add a comment explaining that.

pkg/e2e/fake_metrics_server_test.go
realschwa: @andysim3d Should I revert it to it's previous state (ie: not a row test but rather what was in this commit minus the last error case https://github.com/GoogleCloudPlatform/prometheus-engine/blob/f5edd69004e461dd2fa9dce874deb63521e7edea/pkg/e2e/fake_metrics_server_test.go#L163)

realschwa: I could also have multiple `CreateTimeSeries` in the row tests struct and loop over them too

realschwa: I prefer the latter actually

andysim3d (approved):

PR #272, opened 2022-07-08
Allow gzip compression settings

damemi:
Replacing https://github.com/GoogleCloudPlatform/prometheus-engine/pull/268, following @pintohutch's suggestion:

> for this specific change, my thought would be to change the `OperatorConfig` to add a `Compression CompressionType` field to the `.CollectionSpec` and `.RuleEvaluatorSpec`, where `CompressionType` is a string enum that defaults to `"none"` but can also be `"gzip"`, and maybe other compressions in the future (e.g. snappy)

### DIFF
--- a/pkg/operator/collection.go
+++ b/pkg/operator/collection.go
@@ -225,6 +225,10 @@ func (r *collectionReconciler) ensureCollectorDaemonSet(ctx context.Context, spe
 		flags = append(flags, fmt.Sprintf("--export.credentials-file=%q", p))
 	}
 
+	if len(spec.Compression) > 0 && spec.Compression != "none" {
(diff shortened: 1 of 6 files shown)


pkg/operator/collection.go
```diff
@@ -225,6 +225,10 @@ func (r *collectionReconciler) ensureCollectorDaemonSet(ctx context.Context, spe
 		flags = append(flags, fmt.Sprintf("--export.credentials-file=%q", p))
 	}
 
+	if len(spec.Compression) > 0 && spec.Compression != "none" {
```
andysim3d, line 228: I would suggest change to 
```
if len(spec.Compression) > 0 && spec.Compression == "gzip".
```

andysim3d (approved):

PR #273, opened 2022-07-08
Deploy default AlertManager with operator

damemi:
This adds a manifest for a default AlertManager instance, as well as a boolean in the OperatorConfig to enable (or disable) wiring the rule-evaluator to point to this AlertManager.

### DIFF
--- a/cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
+++ b/cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
@@ -297,6 +297,11 @@ spec:
                       - port
                       type: object
                     type: array
+                  enableDefaultAlertManager:
+                    description: EnableDefaultAlertManager configures the rule-evaluator
+                      to point to a default instance of AlertManager installed by
+                      the operator.
+                    type: boolean
                 type: object
               credentials:
                 description: A reference to GCP service account credentials with which
--- a/cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
+++ b/cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
@@ -297,8 +297,8 @@ spec:
                       - port
                       type: object
                     type: array
-                  enableDefaultAlertManager:
-                    description: EnableDefaultAlertManager configures the rule-evaluator
+                  enableManagedAlertManager:
+                    description: EnableManagedAlertManager configures the rule-evaluator
                       to point to a default instance of AlertManager installed by
                       the operator.
                     type: boolean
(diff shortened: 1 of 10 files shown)


cmd/operator/deploy/operator/12-alertmanager.yaml
```diff
@@ -0,0 +1,90 @@
...
+      labels:
+        app.kubernetes.io/name: gmp-alertmanager
+        app: managed-prometheus-alertmanager
+      annotations:
+        components.gke.io/component-name: managed_prometheus
+        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
+    spec:
+      containers:
+      - name: alertmanager
+        image: prom/alertmanager:latest
```
damemi, line 50: note: we will need to set up our own image build, just using this for testing

cmd/operator/deploy/operator/12-alertmanager.yaml
```diff
@@ -0,0 +1,98 @@
...
+  clusterIP: None
+  ports:
+  - port: 9093
+    targetPort: 9093
+    name: alertmanager
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  namespace: gmp-system
```
damemi, line 43: @lyanco should the default alertmanager go in `monitoring` similar to https://github.com/GoogleCloudPlatform/prometheus-engine/pull/276? Users will need to be able to edit the configmap it mounts to configure it

lyanco: I think gmp-public is where AM and the config should go, that's where the operator config lives, and people are supposed to be able to edit that.

FWIW I thought of using gmp-public for #276 , however the kube-state-metrics manifest was using monitoring, and I deferred to the ghost of Fabian there (didn't know if there was a specific reason that it had to live in "monitoring" or not so I didn't want to touch it). So I moved node-exporter to monitoring instead of gmp-public to keep locally consistent. But if it wouldn't break anything, then I think they should all live in gmp-public.

damemi: sounds good, I actually thought `gmp-public` at first before I saw the other PR. updated

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -297,6 +297,11 @@ spec:
                       - port
                       type: object
                     type: array
+                  enableDefaultAlertManager:
```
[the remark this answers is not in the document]
damemi: sounds good, updated

andysim3d (approved):
LGTM

damemi:
@andysim3d I was waiting to merge this until https://github.com/GoogleCloudPlatform/prometheus-engine/pull/273#discussion_r917113795 was set up with our own image build. That's okay, but just making a note that I don't think we can sync this to internal without a google build

PR #275, opened 2022-07-11
Add ListTimeSeries to the FakeMetricServer

realschwa:
This pull requests does the following:

- Adds `ListTimeSeries` to the `FakeMetricServer` that acts as a drop in replacement for the [endpoints listed here](https://cloud.google.com/monitoring/api/ref_v3/rpc#google.monitoring.v3.metricservice). 
- Adds unit tests to ensure the `FakeMetricServer`'s `ListTimeSeries` is behaving as expected.
- Stores points in the time series in the `FakeMetricServer` in decreasing timestamp order (the way we should return it in `ListTimeSeries`.
- Fixes the unit tests that changed as a result of the above.
- Fixes the `TestNil` test in `TestCreateTimeSeriesBadInput` -- it previously was not running.
- Fixes a bug in `CreateTimeSeries` where it did not properly add multiple time series and adds a test to test the scenario

This is the second of a few PRs that will create a robust testing solution to verify GMP behavior with tests using a local kind cluster and E2E tests using a real GKE cluster.

### DIFF
--- a/pkg/e2e/fake_metric_server.go
+++ b/pkg/e2e/fake_metric_server.go
@@ -30,9 +30,9 @@ type FakeMetricServer struct {
 	maxTimeSeriesPerRequest int
 }
 
-// initialize an empty map in the FakeMetricServer since Go does not let you add to a nil map
 func NewFakeMetricServer(maxTimeSeriesPerRequest int) *FakeMetricServer {
 	return &FakeMetricServer{
+		// initialize an empty map in the FakeMetricServer since Go does not let you add to a nil map
 		timeSeriesByProject:     make(map[string][]*monitoringpb.TimeSeries),
 		maxTimeSeriesPerRequest: maxTimeSeriesPerRequest,
 	}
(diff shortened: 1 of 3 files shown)


pkg/e2e/fake_metric_server.go
```diff
@@ -99,9 +101,63 @@ func (fms *FakeMetricServer) CreateTimeSeries(ctx context.Context, req *monitori
...
+		return nil, errors.New("fake metric server does not support aggregation")
+	}
+	if req.Interval == nil {
+		return nil, errors.New("time interval required")
+	}
+	if req.View == monitoringpb.ListTimeSeriesRequest_HEADERS {
+		return nil, errors.New("header view is not supported")
+	}
+
+	// The filter string MUST be in the form:
```
realschwa, line 120: I actually need to add supporting for multiple filtering here

damemi (commented):
code lgtm, just 2 nits on the PR... before you merge, you should try to squash your commits down to as few logical components as possible. you also may want to consider using `git rebase` to sync with the main branch (instead of `git pull/merge`) so you don't end up with merge commits in your PR

damemi (approved):

PR #277, opened 2022-07-11
Update deploy manifests when building operator image

damemi:
Noticed this while testing https://github.com/GoogleCloudPlatform/prometheus-engine/pull/273, it's a little more convenient when this command updates the `deploy` manifests too.

So for example, when running the steps in https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.4.3/cmd/operator/README.md#run they will work with custom built images (ie when running `make operator` to build new code)


damemi:
ping @andysim3d 
is this okay to merge? I think it will make local testing easier

realschwa (approved):
lgtm

PR #279, opened 2022-07-13
Fix instructions for running operator locally

TheSpiritXIII:
Edit: after help from @damemi , I realized the original error I was trying to fix was caused by the instructions for running the operator locally being incorrect. This PR fixes that.

### DIFF
--- a/cmd/operator/README.md
+++ b/cmd/operator/README.md
@@ -5,24 +5,41 @@ Cloud Prometheus Engine on Kubernetes.
 
 ## Run
 
-Create or update cluster resources required by the operator.
+Deploy all CRDs:
 
 ```bash
 kubectl apply -f deploy/crds/
-kubectl apply -f deploy/operator/
 ```
 
-Run the operator locally (requires active kubectl context to have all permissions
-the operator needs):
+Deploy all of the operator required configurations besides the operator
+deployment, otherwise you will have an operator deployed in addition to your
+local one.
+
+```bash
+kubectl apply -f deploy/operator/00-namespace.yaml
+kubectl apply -f deploy/operator/01-priority-class.yaml
+kubectl apply -f deploy/operator/02-service-account.yaml
+kubectl apply -f deploy/operator/03-role.yaml
+kubectl apply -f deploy/operator/04-rolebinding.yaml
+```
+
+Run the operator locally (requires active kubectl context to have all
+permissions the operator needs):
 
 ```bash
 go run main.go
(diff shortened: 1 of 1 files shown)


pkg/operator/collection.go
```diff
@@ -239,9 +239,10 @@ func (r *collectionReconciler) ensureCollectorDaemonSet(ctx context.Context, spe
 		for _, ev := range c.Env {
 			if ev.Name != "EXTRA_ARGS" {
 				repl = append(repl, ev)
+			} else {
+				repl = append(repl, corev1.EnvVar{Name: "EXTRA_ARGS", Value: strings.Join(flags, " ")})
```
damemi, line 243: what if the `EXTRA_ARGS` flag doesn't exist yet? then none of these will get set

cmd/operator/README.md
```diff
@@ -5,23 +5,41 @@ Cloud Prometheus Engine on Kubernetes.
...
+kubectl apply -f deploy/operator/01-priority-class.yaml
+kubectl apply -f deploy/operator/02-service-account.yaml
+kubectl apply -f deploy/operator/03-role.yaml
+kubectl apply -f deploy/operator/04-rolebinding.yaml
 ```
 
-Run the operator locally (requires active kubectl context to have all permissions
-the operator needs):
+Run the operator locally (requires active kubectl context to have all
+permissions the operator needs):
```
damemi, line 27: I think it would be better if we changed this whole readme from focusing on running the operator locally to focus on running the operator in a cluster. That really simplifies the steps, and provides a more realistic experience that is consistent with actual GMP

in that case, the steps are essentially:
1. (pre-req) set up GCP service account with [artifact registry permissions](https://cloud.google.com/kubernetes-engine/docs/troubleshooting#permission_denied_error)
2. build and push the operator with `DOCKER_PUSH=1 make operator`
3. Create CRDs with `kubectl apply-f deploy/crds`
4. Create the rest of the resources with `kubectl apply -f deploy --recursive`

damemi (changes requested):

pkg/operator/collection.go
TheSpiritXIII: Redacted this change as this was incorrect.

cmd/operator/README.md
TheSpiritXIII: Please take a look!

damemi (approved):
looks great, thanks!

PR #287, opened 2022-07-25
Initialize empty config for collector and evaluator pods

damemi:
This fixes a bug where the collector and rule-evaluator pods start up before the config-reloader, causing the collector/evaluator to restart due to not being able to find a config. For example:
```
$ kubectl logs pod/rule-evaluator -n gmp-system -c evaluator
level=error ts=2022-07-25T19:14:05.872221429Z caller=main.go:144 msg="Error loading config (--config.file=/prometheus/config_out/config.yaml)" err="open /prometheus/config_out/config.yaml: no such file or directory"
```
this adds an init container to both pods that just creates the empty config file. in my testing, this appeared to fix the issue


andysim3d (approved):
LGTM

PR #285, opened 2022-07-22
Add basic auth to frontend handler

damemi:
Fixes https://github.com/GoogleCloudPlatform/prometheus-engine/issues/242

PR #292, opened 2022-07-27
Use proper response codes for frontend auth

damemi:
Following https://github.com/GoogleCloudPlatform/prometheus-engine/pull/285
Return `401` for failed authentication
Return `403` for failed authorization

I think this will be better for debugging basic auth setups for users


realschwa (approved):
lgtm

PR #297, opened 2022-07-29
Switch managed AlertManager to always-on, and use secrets

damemi:
This updates the managed AlertManager to be always on by default, to run in the `gmp-system` namespace, and to mirror user config from a Secret they create in `gmp-public`.


damemi:
Still need to test this, but opening now to show the changes

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -103,6 +103,30 @@ spec:
               object represents. Servers may infer this from the endpoint the client
               submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
+          managedAlertManager:
```
xichen2020, line 106: `managed` is implied. I'd recommend `alertManager`. 

Also remember to update description etc and documentation in other places.

I think it should be at the same level as `rules` but weird that both are under `collection`. A better section would be `alerting`.. Ah well too late to change.

pkg/operator/operator_config.go
```diff
@@ -352,20 +413,18 @@ func (r *operatorConfigReconciler) makeAlertManagerConfigs(ctx context.Context,
...
+	amNamespacedName := types.NamespacedName{
+		Namespace: r.opts.OperatorNamespace,
+		Name:      "alertmanager",
+	}
+	// if the default AlertManager exists, append it to the list of spec.Alertmanagers
+	if resourceErr := r.client.Get(ctx, amNamespacedName, &appsv1.StatefulSet{}); resourceErr == nil {
+		alertManagers = append(alertManagers, monitoringv1.AlertmanagerEndpoints{
+			Name:      amNamespacedName.Name,
+			Namespace: amNamespacedName.Namespace,
+			Port:      intstr.FromInt(9093),
```
xichen2020, line 426: Fine for now but note the possibility that users may edit the alertmanager port in the yaml and break this. Can make it configurable if that use case does come up.

xichen2020 (commented):
Only skimmed over it and looked fine, will take a second pass.

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
damemi: Since there is already an `alertmanagers` field under `rules` for users to configure their own unmanaged AM, I think it is worth the distinction here that we are specifically talking about the "default"/managed alert manager

Also this is at the same level as `rules`, not under `collection`: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/450382f6b3baa484941b0500c98b013830bd0d9c/pkg/operator/apis/monitoring/v1/types.go#L49-L58

damemi:
Updated with the latest feedback in the design discussions:
* Operator expects a specific config secret name: `gmp-public/managed-alertmanager-config`
* This default is set and shown in the OperatorConfig CRD
* Users can edit the value in the CRD to use a custom secret name

Also added some fixes to the config reloader, and an initContainer to address the same issue as https://github.com/GoogleCloudPlatform/prometheus-engine/pull/287

damemi:
@xichen2020 is this good to merge?

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -304,11 +332,6 @@ spec:
                       - port
                       type: object
                     type: array
-                  enableManagedAlertManager:
```
xichen2020, line 307: I'm assuming this change is not yet released? Otherwise I think it's a breaking change.

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -103,6 +103,34 @@ spec:
...
+                properties:
+                  key:
+                    description: The key of the secret to select from.  Must be a
+                      valid secret key.
+                    type: string
+                  name:
+                    description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                      TODO: Add other useful fields. apiVersion, kind, uid?'
+                    type: string
+                  optional:
```
xichen2020, line 127: nit: IIRC yaml fiels are optional by default, so not sure if this is needed.

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -103,6 +103,34 @@ spec:
               object represents. Servers may infer this from the endpoint the client
               submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
+          managedAlertmanager:
```
xichen2020, line 106: A bit odd that this has a single key `config`. Any reason to not just move the properties of `config` up a level and remove the `config` key?

pkg/operator/operator_config.go
```diff
@@ -280,6 +299,68 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorSecrets(ctx context.Contex
...
+	// default secret exists (so that the alertmanager pod doesn't crash due to no
+	// config found). This flow also handles user deletion/disabling of managed AM.
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "alertmanager",
+			Namespace:   r.opts.OperatorNamespace,
+			Annotations: componentAnnotations(),
+			Labels:      alertmanagerLabels(),
+		},
+		Data: map[string][]byte{"config.yaml": []byte(alertManagerNoOpConfig)},
```
xichen2020, line 318: Make `config.yaml` a file-level const and reference it here to simplify refactoring if we need to support arbitrary keys in the future.

pkg/operator/operator_config.go
```diff
@@ -280,6 +299,68 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorSecrets(ctx context.Contex
...
+		// if the user is overriding the default secret name, use that
+		secretNamespacedName.Name = spec.Config.Name
+	}
+
+	err := r.client.Get(ctx, secretNamespacedName, publicSecret)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		// if the config secret is not found, it may have been manually deleted
```
xichen2020, line 336: nit: `If ...`

pkg/operator/operator_config.go
```diff
@@ -280,6 +299,68 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorSecrets(ctx context.Contex
...
+		// (ie, to disable managed AM), so we will continue with restoring the no-op config
+		// so that the managed AM pod doesn't crash loop.
+		logger.Info("alertmanager config secret not found")
+	}
+
+	if publicSecret.Data != nil {
+		if len(publicSecret.Data) != 1 {
+			return errors.New("alertmanager config secret must have exactly 1 key")
```
xichen2020, line 344: IIUC this will cause the reconciler to reenqueue the request and encounter this error again. Is it worth retrying?

I'm curious about how users can discover this error. Will this be logged in operator logs?

pkg/operator/operator_config.go
```diff
@@ -280,6 +299,68 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorSecrets(ctx context.Contex
...
+		}
+		// Alertmanager config could have been created from any file name, so
+		// even though there is only 1 key we have to loop over it to get the data
+		for _, bytes := range publicSecret.Data {
+			secret.Data["config.yaml"] = bytes
```
xichen2020, line 349: Reference the const here.

pkg/operator/operator_config.go
```diff
@@ -342,30 +423,28 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorDeployment(ctx context.Con
...
-		if resourceErr := r.client.Get(ctx, amNamespacedName, &appsv1.StatefulSet{}); resourceErr == nil {
-			alertManagers = append(alertManagers, monitoringv1.AlertmanagerEndpoints{
-				Name:      amNamespacedName.Name,
-				Namespace: amNamespacedName.Namespace,
-				Port:      intstr.FromInt(9093),
-			})
-		}
+	amNamespacedName := types.NamespacedName{
+		Namespace: r.opts.OperatorNamespace,
+		Name:      "alertmanager",
```
xichen2020, line 439: nit: `NameAlertManager`

xichen2020 (commented):
Do we have kind tests for this?

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
damemi: correct this is not yet released, so it won't be breaking. we got it in time :)

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
damemi: this is generated by kubebuilder

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
damemi: I like that this leaves room to add more settings for the alertmanager. `config` here is referring to the actual config file, but I could see other settings under `AlertmanagerSpec` as well (replicas, auth..)

pkg/operator/operator_config.go
damemi: Yeah, this probably doesn't need to requeue. But it should be brought to the user's attention in the operator logs, so updated to a `logger.Error`

pkg/operator/operator_config.go
```diff
@@ -280,6 +300,69 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorSecrets(ctx context.Contex
...
+		},
+		Data: map[string][]byte{alertmanagerConfigKey: []byte(alertManagerNoOpConfig)},
+	}
+
+	publicSecret := &corev1.Secret{}
+	secretNamespacedName := types.NamespacedName{
+		Namespace: r.opts.PublicNamespace,
+		Name:      "alertmanager-config",
+	}
+	if spec != nil && len(spec.Config.Name) > 0 {
```
xichen2020, line 327: What are your thoughts on returning an error if `len(spec.Config.Key) > 0`? I'm mainly thinking that if a user specifies that, it's better to reject it as an error than silently ignoring it, so if we change the behavior in the future to support arbitrary keys, in the worst case it'll remain not working (if e.g., the key doesn't exist), but that's better than changing from working to not working.

xichen2020 (commented):
Thoughts on adding some kind tests for this?

pkg/operator/operator_config.go
damemi: Key is required in [SecretKeySelector](https://github.com/kubernetes/api/blob/v0.24.4/core/v1/types.go#L2054), so it needs something to be set if we're using that type

damemi:
> Thoughts on adding some kind tests for this?

I can add some basic e2e in https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/operator/e2e/main_test.go

pkg/operator/operator_config.go
xichen2020: True. Perhaps we can check against `alertmanagerConfigKey` and return an error if it's different, since we are defaulting to that?

damemi: That would require users to always create their configmap with a key that matches `alertmanagerConfigKey`. one of our intentions is to eliminate the number of steps users must get exactly right, so we don't want to do that

xichen2020: Okay it seems we really don't care about the secret key and anything goes as long as the secret has one key. In that case I think it's probably better to remove the `key` property from the config and rename `name` to something like `singleKeySecret` or `keylessSecret` to make the intent clear. If we need to properly support keys in the future I think we'll need to add another field to avoid breaking backward compatibility.

Wdyt?

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -105,31 +105,15 @@ spec:
...
-                    description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-                      TODO: Add other useful fields. apiVersion, kind, uid?'
-                    type: string
-                  optional:
-                    description: Specify whether the Secret or its key must be defined
-                    type: boolean
-                required:
-                - key
-                type: object
+              configSecret:
```
xichen2020, line 112: nit: `singleKeyConfigSecret`.

manifests/setup.yaml
```diff
@@ -1076,6 +1076,18 @@ spec:
...
+          managedAlertmanager:
+            default:
+              configSecret: managed-alertmanager-config
+            description: ManagedAlertmanager holds information for configuring the
+              managed instance of Alertmanager.
+            properties:
+              configSecret:
+                description: ConfigSecret refers to the name of a single-key Secret
+                  in the public namespace that holds the managed Alertmanager config
+                  file
```
xichen2020, line 1088: nit: `file.`

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -103,6 +103,18 @@ spec:
...
+          managedAlertmanager:
+            default:
+              configSecret: managed-alertmanager-config
+            description: ManagedAlertmanager holds information for configuring the
+              managed instance of Alertmanager.
+            properties:
+              configSecret:
+                description: ConfigSecret refers to the name of a single-key Secret
+                  in the public namespace that holds the managed Alertmanager config
+                  file
```
xichen2020, line 115: nit: `file.`

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -130,9 +133,14 @@ type ExportFilters struct {
...
-	// EnableManagedAlertManager configures the rule-evaluator to point to a
-	// default instance of AlertManager installed by the operator.
-	EnableManagedAlertManager bool `json:"enableManagedAlertManager,omitempty"`
+}
+
+// ManagedAlertmanagerSpec holds configuration information for the managed
+// Alertmanager instance.
+type ManagedAlertmanagerSpec struct {
+	// ConfigSecret refers to the name of a single-key Secret in the public namespace that
+	// holds the managed Alertmanager config file
```
xichen2020, line 142: nit: `file.`

xichen2020 (approved):

damemi:
Squashed and added e2e tests

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
damemi: do you think it's okay to leave `singleKey` out of the name since that's the only option right now? If we do add a multiKey option later, it would make sense to differentiate in the name then. but for simplicity I think the name is good as-is, and just document that it needs to have only 1 key. wdyt?

xichen2020: So in the future case we'll provide `configSecret` and `multiKeyConfigSecret`? Sounds fine to me.

damemi: yup that's what I was thinking

pkg/operator/operator_config.go
pintohutch: What we should do is actually see if the default AM [_service_ ](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/8e89c873d3bc21f735941b968461c5e48c153d9f/cmd/operator/deploy/operator/12-alertmanager.yaml#L23-L35) exists, and use the port in the retrieved service object directly.

I can do that in a follow-up PR to shake off some of the cobwebs 🙂

pintohutch: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/314

PR #301, opened 2022-08-08
Target status reporting

TheSpiritXIII:
This PR adds target polling and aggregation into the pod monitoring status subresource.

Polling happens every 10 seconds with 4 concurrent jobs by default but this can be configured, and we save metrics storing the total time taken so we can make adjustments appropriately.

## Performance
Example application GKE performance tests:
03 nodes, 02 pods each, 006 pods total -> ~0025 ms
20 nodes, 32 pods each, 640 pods total -> ~0180 ms
50 nodes, 02 pods each, 100 pods total -> ~1200 ms

Scales well with pods increasing, less well with nodes.

This is because there are exactly 2 collector pods per node and we hit each collector, so more nodes means more collectors.


lyanco:
"This is because there are exactly 2 collector pods per node and we hit each collector, so more nodes means more collectors." -- do we have to hit both? IIRC they are replicas so we could just hit one, in which case we might ~double node scalability. Thoughts?

lyanco:
Does this also support polling for ClusterPodMonitoring? Or is that later?

TheSpiritXIII:
> do we have to hit both? 
I'll double check this, but you're right. We do not. If they're replicas, there may be duplicate data in this current implementation.

> Does this also support polling for ClusterPodMonitoring?
Yes, with full unit test and integration test suites included for both PodMonitoring and ClusterPodMonitoring.

lyanco:
I imagine with very large clusters, there may be a significant delay after applying the CR before the status is correct. This could lead to people thinking their CR wasn't picked up correctly if they check this too soon after applying. Is there already, or could we add, a "last completed" time to this?

pkg/operator/collection.go
```diff
@@ -124,6 +125,20 @@ func newCollectionReconciler(c client.Client, opts Options) *collectionReconcile
 	}
 }
 
+func patchCollectionStatus(ctx context.Context, kubeClient client.Client, statusContainer monitoringv1.PodMonitoringStatusContainer) error {
```
damemi, line 128: why does this have to patch the yaml manually? could it just call `Update()` with the new object?

TheSpiritXIII: To reduce conflicts. Two different goroutines are now updating pod monitoring statuses at the same time:
1. The conditions.
2. The target statuses.

With an update, if it fails due to version mismatch, you have to re-create the object and try the request to update again. I believe we can avoid that with a patch.

TheSpiritXIII:
> I imagine with very large clusters, there may be a significant delay (~1m) after applying the CR before the status is correct. 

To also clarify, in this case, it will poll every 10 seconds but if it's still in progress, it will skip that "tick". So we can expect an update each minute. Let me know if you'd like tweaks to this.

> Is there already, or could we add, a "last completed" time to this?

Correct, we have a field named `LastUpdateTime`.

pkg/operator/collection.go
damemi: how does patching react to concurrent updates like this? Also if an update fails, you should be able to return an error in the [`reconcile.Result`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result) that re-queues the reconcile attempt and just let it run again, was this something you tried?

pkg/operator/operator.go
```diff
@@ -284,6 +318,19 @@ func (o *Operator) Run(ctx context.Context) error {
...
 		return errors.Wrap(err, "setup rule-evaluator controllers")
 	}
+	if o.targetStatus != nil {
+		go func() {
+			tickerBuilder := func() Ticker {
+				return NewTicker(targetPollInterval)
+			}
+			o.logger.Info("starting target status reporting polling")
+			for {
+				if err := o.targetStatus.Start(ctx, tickerBuilder); err != nil {
```
damemi, line 328: did you try starting this with the [controller pattern](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/5b8121fe5ec7290c65a24657e24285e035a5def7/pkg/operator/operator_config.go#L104-L130) used by the other controllers? Just thinking that using that would simplify all of your informer setup, but I'm assuming you had a reason not to

pkg/operator/target_status.go
```diff
@@ -0,0 +1,973 @@
...
+	dataManagerBuilder dataManagerBuilder
+}
+
+func NewTargetStatusPoller(logger logr.Logger, restConfig *rest.Config, registry prometheus.Registerer, opts Options) (TargetStatusProcessor, error) {
+	builder, err := newInformerDataManagerBuilder(restConfig, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return newTargetStatusPoller(logger, opts, registry, builder)
```
damemi, line 179: any reason for this nested private function?

pkg/operator/target_status.go
```diff
@@ -0,0 +1,973 @@
...
+			}()
+			if err != nil {
+				return err
+			}
+		}
+	}
+}
+
+func poll(ctx context.Context, logger logr.Logger, opts Options, manager dataManager) error {
+	targets, err := fetchTargets(ctx, logger, opts, manager)
```
damemi, line 251: same question on the private func, are `fetchTargets` and `processTargets` used or tested anywhere else or could they just be inlined?

pkg/operator/collection.go
damemi: In general each reconcile attempt should be idempotent, so your logic can always back off/retry and re-build its state each time. I see that none of our other control loop functions look like they're using this, so maybe there's a reason for that

pkg/operator/target_status.go
TheSpiritXIII: The private version is used internally for unit tests.

pkg/operator/target_status.go
TheSpiritXIII: `fetchTargets` is unit tested. I will inline the latter.

pkg/operator/collection.go
TheSpiritXIII: > your logic can always back off/retry and re-build its state each time
I think this is a harder issue than it seems. You need to introduce a retry amount so you don't keep retrying indefinitely and possibly with jitter, which makes updates very random. Should you re-fetch everything or should you retry with the stale fetched data? At that point, since I refresh the state every 10 seconds, the question is whether it's worth it to get the delayed update or just wait for the next one.

pkg/operator/operator.go
TheSpiritXIII: No. Because my logic polls every 10 seconds and controllers are event driven, if I had used controllers, I would have still needed informers as well. I need to fetch the current Kubernetes state on demand and informers are the most performant option. Correct me if I'm wrong.

pintohutch: So I actually think there may be a way to greatly reduce the code complexity by effectively polling using the [`source.Channel`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/source#Channel) controller-runtime construct, e.g.

I haven't tested this exactly, but I'm picturing something like this.

```go
// Create target status controller.
func setupTargetStatusControllers(op *Operator) error {
	events := make(chan event.GenericEvent)
	err := ctrl.NewControllerManagedBy(op.manager).
		Named("target-status").
		// Adding &corev1.Pod{fieldSelector: gmp-system} to manager cache
		// so reconciler client has cached collector pods to fetch.
		Watches(&source.Channel{
			Source: events,
		}, &handler.EnqueueRequestForObject{}).
		Complete(newTargetStatusReconciler(op.manager.GetClient(), events, op.opts))
	if err != nil {
		return errors.Wrap(err, "create target status controller")
	}
	return nil
}

// targetStatusReconciler to hold cached client state and source channel.
type targetStatusReconciler struct {
	client client.Client
	ch     chan event.GenericEvent
	opts   Options
}

// newTargetStatusReconciler creates a new targetStatusReconciler and kicks off
// the reconcile loop by sending the first event to the source channel.
func newTargetStatusReconciler(c client.Client, ch chan event.GenericEvent, opts Options) *targetStatusReconciler {
	// Kick off reconcile loop.
	ch <- event.GenericEvent{}
	return &targetStatusReconciler{
		client: c,
		ch:     ch,
		opts:   opts,
	}
}

// Reconcile polls the collector pods, fetches and aggregates target status and upserts
// into each PodMonitoring's Status field.
func (r *targetStatusReconciler) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
	minDelay := time.NewTimer(10 * time.Second)

	// Do the target status polling and updating...
	// Can use r.opts to control parallelism, fetch timeouts, etc...

	// Check if we beat the timer, otherwise wait.
	<-minDelay.C
	// Re-trigger event
	r.ch <- event.GenericEvent{}

	return reconcile.Result{}, nil
}
```

pintohutch (changes requested):
Left some comments. I haven't gone through all of `target_status.go` to see if using a controller pattern to effectively perform polling can clean things up.

I also think it makes sense to take out anything related to e2e tests (including any interfaces or fields). These are valuable, but add a lot of complexity to the review. Once we iron out the changes here, we can review the e2e tests in a separate PR.

Overall I think this will make the code easier to follow, and reduce the change to something we can iterate on a little faster. Whattya think?

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1006,13 +1035,51 @@ type RelabelingRule struct {
...
+	// Total number of active, unhealthy targets.
+	UnhealthyTargets int64 `json:"unhealthyTargets,omitempty"`
+	// Last time this status was updated.
+	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
+	// A fixed sample of targets grouped by error type.
+	SampleGroups []SampleGroup `json:"sampleGroups,omitempty"`
+	// Fraction of collectors included in status, bounded [0,1].
+	// Ideally, this should always be 1. Anything less can
+	// be considered a problem and should be investigated.
+	CollectorsFraction string `json:"collectorsFraction,omitempty"`
```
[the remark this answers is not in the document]
TheSpiritXIII: I believe the code generator had a problem with floats. Specifically, it was showing a warning about serializing them.

Makefile
```diff
@@ -108,7 +108,9 @@ kindtest:    ## Run e2e test suite against fresh kind k8s cluster.
 	@echo ">> running container"
 # We lose some isolation by sharing the host network with the kind containers.
 # However, we avoid a gcloud-shell "Dockerception" and save on build times.
-	docker run --network host --rm -v $(DOCKER_VOLUME):/var/run/docker.sock gmp/kindtest ./hack/kind-test.sh
+# Make sure we forward outer environment variables to the Docker runtime.
+	echo "Hello Github! ${GITHUB_ACTIONS}"
+	docker run --env GITHUB_ACTIONS="${GITHUB_ACTIONS}" --network host --rm -v $(DOCKER_VOLUME):/var/run/docker.sock gmp/kindtest ./hack/kind-test.sh
```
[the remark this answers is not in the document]
TheSpiritXIII: I needed the e2e tests to detect whether we're running within Github Actions, so I needed to pass the environment variable into the script. The issue is that we use the host network and Github Action runners do not support IPv6. Why target status requires IPv6 is not something I've been able to figure out. To fetch each target, we query localhost and for some reason, the k8s client-go library was trying to use the IPv6 equivalent "::1". I have not been able to reproduce this locally (I tried disabling IPv6 to no avail) so I just gave up and decided to disable these e2e tests on Github Actions.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+}
+
+func fetchTargets(ctx context.Context, logger logr.Logger, opts Options, manager dataManager) ([]*prometheusTargetsResultAndTime, error) {
+	podMap, err := manager.getPrometheusPods(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set up pod job queue and jobs
+	podDiscoveryChan := make(chan prometheusPod, opts.TargetPollConcurrency)
```
pintohutch, line 262: Did you see any performance gain when using the buffer?

I feel like a more conventional use of buffers is to allow the sender to batch a bunch of requests and then do other stuff. However, in this case the sender is only concerned with putting items in the channel.

Your concurrency is controlled in the for loop, with a fixed set of blocking goroutines, each of which are continually trying to pull new items from the pod channel. Since the sender has virtually no latency in sending more items through the channel (i.e. the receivers aren't going to "beat" the sender before it can send in a new signal) an unbounded channel should be fine.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+	wg := sync.WaitGroup{}
+	wg.Add(int(opts.TargetPollConcurrency))
+
+	// Must be unbounded or else we deadlock.
+	targetChan := make(chan *prometheusTargetsResultAndTime)
+	for i := uint16(0); i < opts.TargetPollConcurrency; i++ {
+		// Wrapper function to handle errors.
+		go func(thread uint16) {
+			fetcher, err := manager.getTargetFetcher(ctx, thread)
+			if err != nil {
```
pintohutch, line 272: Also, as mentioned before, if we can get rid of the `dataManager` interface, this flow can hopefully just become a single call, i.e.
```go
for i := 0; i < opts.TargetPollConcurrency; i++ {
	go func {
		defer wg.Done()
		// Process incoming pods and send results to target channel, blocking until pod channel closes.
		processTargetQueue(ctx, logger, podCh, targetCh)
	}
}
```

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+					port: port,
+					pod:  pod,
+				}
+			}
+		}
+
+		// Must close so jobs aren't waiting on the channel indefinitely.
+		close(podDiscoveryChan)
+
+		// Close target after we're sure all targets are queued.
```
pintohutch, line 297: I don't think these two lines need to be in the sending goroutine. They should be blocking the main thread execution until were done processing the pods. Then we can proceed with building and returning our results synchronously.

pintohutch (changes requested):
Here's some more partial feedback. I think I've gotten through everything except the details of the status building and updating, which should be pretty straightforward.

I can give my thoughts on that later.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+	results := make([]*prometheusTargetsResultAndTime, 0)
+	for target := range targetChan {
+		results = append(results, target)
+	}
+
+	return results, nil
+}
+
+func populateTargets(ctx context.Context, logger logr.Logger, manager dataManager, targets []*prometheusTargetsResultAndTime) error {
+	endpointBuilder := newScrapeEndpointBuilder()
```
pintohutch, line 311: Do we need a dedicated struct to maintain state here? In general, if we don't need to maintain state, it's easier to plumb this through a pure function. This also makes unit testing easier. i.e.
```go
statusesByJob := buildEndpointStatuses(targets)
```

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+	for _, target := range targets {
+		if err := endpointBuilder.Add(target); err != nil {
+			return err
+		}
+	}
+
+	endpointMap := endpointBuilder.Build()
+
+	podMonitorings, err := manager.getPodMonitorings(ctx)
```
pintohutch, line 320: Why do we need to re-fetch the podmonitorings here? Why not just something like:
```go
for job, eps := range statusesByJob {
	obj, err := buildObjFromJob(job)
	if err != nil {
		// log error
	}
	status := &monitoringv1.PodMonitoringStatus{EndpointStatuses: eps}
	err := patchCollectionStatus(ctx,
		client,
		obj,
		status)
	if err != nil {
		// log error
	}
}
```

pintohutch (changes requested):
Also some suggestions on building and patching the status back in.

pkg/operator/target_status.go
pintohutch: That being said, I think there is some merit in having your `scrapeEndpointStatusBuilder`. But the `buildEndpointStatuses()` function can handle the bookkeeping of setting up the `map[job][]monitoringv1.ScrapeEndpointStatuses` much like your current `scrapeEndpointBuilder` with less indirection.

I also vote moving the `scrapeEndpointStatusBuilder` to its own `endpoint_status_builder.go` file for easier parsing.

pkg/operator/apis/monitoring/v1/types.go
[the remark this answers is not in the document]
TheSpiritXIII: From `controller-gen`:
```
found float, the usage of which is highly discouraged, as support for them varies across languages. Please consider serializing your float as string instead. If you are really sure you want to use them, re-run with crd:allowDangerousTypes=true
```

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+		logger:  logger,
+	}
+
+	return reconciler, nil
+}
+
+type emptyObject struct{}
+
+func (r *targetStatusReconciler) Start(ctx context.Context, tickerBuilder func() Ticker) error {
+	r.tickerBuilder = tickerBuilder
```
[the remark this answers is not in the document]
TheSpiritXIII: This would definitely work for managed collections but for the self-hosted option (or even developers), users can create, delete and re-create `DaemonSet`, causing the reconciler to retrigger and then you'd have two reconciler loops.

An alternative idea I had to remove the `Start` is to use `manager.Add` to add a callback when the manager is started and add an event into the queue. This ensures the controller is started only once, and started at the same time as the manager without needing to call a `Start` method from the outside.

Thoughts?

pkg/operator/operator.go
```diff
@@ -357,6 +401,19 @@ func (o *Operator) Run(ctx context.Context) error {
 	if err := setupOperatorConfigControllers(o); err != nil {
 		return errors.Wrap(err, "setup rule-evaluator controllers")
 	}
+	if o.targetStatus != nil {
+		go func() {
+			tickerBuilder := func() Ticker {
+				return NewTicker(targetPollInterval)
+			}
+			o.logger.Info("starting target status reporting polling")
+			for {
```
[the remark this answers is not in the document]
TheSpiritXIII: See [this comment](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/301#discussion_r1007528728).

pkg/operator/e2e/main_test.go
```diff
@@ -557,9 +557,14 @@ func TestWebhookCABundleInjection(t *testing.T) {
 	}
 }
 
-// testCollectorDeployed does a high-level verification on whether the
-// collector is deployed to the cluster.
-func testCollectorDeployed(ctx context.Context, t *testContext) {
+func ensureOperatorConfigsExist(ctx context.Context, t *testContext) error {
```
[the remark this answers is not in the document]
TheSpiritXIII: Removed. e2e tests will be in a separate PR.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+	}
+
+	// Set up pod job queue and jobs
+	podDiscoveryChan := make(chan prometheusPod, opts.TargetPollConcurrency)
+	wg := sync.WaitGroup{}
+	wg.Add(int(opts.TargetPollConcurrency))
+
+	// Must be unbounded or else we deadlock.
+	targetChan := make(chan *prometheusTargetsResultAndTime)
+	for i := uint16(0); i < opts.TargetPollConcurrency; i++ {
```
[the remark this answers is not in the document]
TheSpiritXIII: Because that causes `i` to be a signed `int`.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+	targetStatusDuration = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "prometheus_engine_target_status_duration",
+		Help: "A metric indicating how long it took to fetch the complete target status.",
+	}, []string{})
+)
+
+// Wraps a time.Ticker so we can replace for unit testing.
+type Ticker interface {
+	// Stops the ticker from ticking.
+	Stop()
```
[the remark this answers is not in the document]
TheSpiritXIII: I agree I no longer need `Stop`. However, I believe the implementation is useful for unit testing. It lets me mock out the timer in the unit tests, letting me control precisely when a tick happens.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+			if port != nil {
+				break
+			}
+		}
+	}
+	if port == nil {
+		return nil, errors.New("Unable to detect Prometheus port")
+	}
+
+	podsMap[*port] = podsFiltered
```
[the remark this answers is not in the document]
TheSpiritXIII: I wanted to return both port number and a pod list, and I thought a map felt more natural than a tuple. My latest change will just get the pod from the outside and pass in the selector.

pkg/operator/e2e/main_test.go
```diff
@@ -780,6 +809,16 @@ func testCollectorSelfClusterPodMonitoring(ctx context.Context, t *testContext)
 	if err != nil {
 		t.Errorf("unable to validate ClusterPodMonitoring status: %s", err)
 	}
+	err = checkEndpointStatus(t.T, func() (monitoringv1.PodMonitoringStatusContainer, error) {
```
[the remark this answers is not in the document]
TheSpiritXIII: This will be part of the e2e test PR, so resolving this.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+	getPrometheusPods(ctx context.Context) (map[int32][]*corev1.Pod, error)
+
+	getPodMonitorings(ctx context.Context) ([]monitoringv1.PodMonitoring, error)
+	getClusterPodMonitorings(ctx context.Context) ([]monitoringv1.ClusterPodMonitoring, error)
+
+	patchPodMonitoring(ctx context.Context, podMonitoring *monitoringv1.PodMonitoring) error
+	patchClusterPodMonitoring(ctx context.Context, podMonitoring *monitoringv1.ClusterPodMonitoring) error
+
+	// Returns the target fetcher, with an zero-index number `thread` that signifies which goroutine this is for.
+	getTargetFetcher(ctx context.Context, thread uint16) (targetFetcher, error)
```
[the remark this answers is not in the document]
TheSpiritXIII: I made it a `targetFetcherBuilder` as opposed to a pure function because there are some lookups that only need to be done once as opposed to per-thread.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+			// Not having data is not an error.
+			if apierrors.IsNotFound(err) {
+				onceLog.Do(func() {
+					r.logger.Error(err, "Kubernetes config not found")
+				})
+				return nil
+			}
+			return err
+		}
+		// In case Kubernetes config is removed and re-added.
```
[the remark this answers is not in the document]
TheSpiritXIII: Realistically rarely. It's primarily helpful to self-deployed collection users or developers who fiddle with configurations, e.g. the collector DaemonSet. The `Once` helps prevent flooding the logs with error messages every tick, but also resets in case the errors reoccur later on. This error might also happen if somehow the Kubernetes controller watcher goes out of sync.

pkg/operator/target_status.go
TheSpiritXIII: This doesn't work because writing to an unbuffered channel is blocking on reading. The pod channel is read by the worker thread who writes to the target channel so it's blocked here. As an example, consider 2 worker threads and 3 pods. The worker threads start by waiting for work. Somebody needs to give the worker threads some work. Today this is the separate goroutine. The worker threads get 2 pods and they spit out 2 targets and they're blocked here. The main thread is reading targets to unblock the workers. The workers are unblocked and they get the next pod and spit out the next target. The main thread is still reading targets and is now blocked until the channel is closed. That's why the sending thread also closes the target channel because it knows that it sent all the pods, so all targets should be sent.

There may be a more clever way to do this with only two threads, but it's tricky because there's two channels involved.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,936 @@
...
+	// defer minDelay.Stop()
+
+	onceLog := new(sync.Once)
+
+	// Wrapper function so we could exit early.
+	err := func() error {
+		now := time.Now()
+		if err := poll(ctx, r.logger, r.opts, r.manager); err != nil {
+			// Not having data is not an error.
+			if apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
TheSpiritXIII: This is the case when there's no DaemonSet or OperatorConfig. We do log an error here.

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -200,6 +201,18 @@ type SecretOrConfigMap struct {
 	ConfigMap *v1.ConfigMapKeySelector `json:"configMap,omitempty"`
 }
 
+// PodMonitoringStatusContainer represents a Kubernetes CRD that monitors pods
+// and contains a status sub-resource.
+type PodMonitoringStatusContainer interface {
+	client.Object
+
+	// Returns a unique identifier for this CRD, that may be used as a map key.
+	GetKey() string
```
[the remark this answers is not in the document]
TheSpiritXIII: I think the interface is still useful because it lets us type our status update queue in the collection reconciler but maybe it can be reworked/renamed some other way.

pkg/operator/target_status.go
TheSpiritXIII: I ended up turning it into a builder on the outside of the concurrency loop and use the built object inside because the object has some common state. I specifically test that concurrency is working in the unit test. Let me know what you think of this interface in the new iteration of the code.

pintohutch (changes requested):
Nice! We're getting closer. A few more comments.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,560 @@
...
+)
+
+var (
+	targetStatusDuration = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "prometheus_engine_target_status_duration",
+		Help: "A metric indicating how long it took to fetch the complete target status.",
+	}, []string{})
+)
+
+// Wraps a time.Timer so we can replace for unit testing.
```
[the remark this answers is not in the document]
TheSpiritXIII: To me this is a bit clunky. Although we do not use the `Timer` interface very much today, I'm not a fan of the `Timer` interface having an implicit state between `Start` and `Channel`. The closure did well to convey that. How about two interfaces, one to create/build the ticker and one to get the channel? That way we can still avoid the closure, and matches the k8s client libraries conventions.

pkg/operator/operator.go
```diff
@@ -249,6 +294,12 @@ func New(logger logr.Logger, clientConfig *rest.Config, registry prometheus.Regi
 		return nil, errors.Wrap(err, "create client")
 	}
 
+	if !opts.TargetPollDisabled {
```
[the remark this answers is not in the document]
TheSpiritXIII: I don't think it makes sense to store the registry since we won't be adding new metrics at runtime, but I think it makes sense to have the registry as a parameter in the `Run`. In fact, we weren't even using it in the constructor previously (although we took it in) until I added the target status controller. Thoughts?

pkg/operator/target_status.go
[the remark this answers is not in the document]
TheSpiritXIII: The consequence is we'd be having double the amount of threads polling targets at a time. This could potentially happen in a deployed 

I thought of a few ways we can handle this:
1. Don't.
2. Use a fake object in the `For` in conjunction with themanager.Add`.
3. Use a `DaemonSet` and creation filter in the `For` in conjunction with an `EventHandler` to only accept the first ever create event.
4. If `controller-runtime` gave us more control over event handling, but I couldn't figure this one out based on the Go docs.

pkg/operator/target_status.go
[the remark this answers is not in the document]
TheSpiritXIII: It seems like credentials are not necessary but we still save and re-use the client for all threads.

hack/Dockerfile
```diff
@@ -29,8 +29,11 @@ COPY pkg pkg
 # git diff.
 RUN git config --global user.email "test@test.com" \
 	&& git init && git add . && git commit -am 'base'
-# Hack to get multiline build arg to run properly.
-RUN echo ${RUNCMD} | sh && echo 'done'
+# Install Kubebuilder tools to run unit tests
+# Must use bash to source environment variables from the tool.
+SHELL ["/bin/bash", "-c"]
+RUN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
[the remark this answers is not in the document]
pintohutch: Honestly, maybe `Reconcile` is the wrong method. Maybe just [`poll`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/TheSpiritXIII/target-status/pkg/operator/target_status.go#L221)?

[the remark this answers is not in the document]
TheSpiritXIII: `fake.Client` worked with calling `Reconcile` manually.

hack/Dockerfile
```diff
@@ -29,7 +29,8 @@ COPY pkg pkg
 # git diff.
 RUN git config --global user.email "test@test.com" \
 	&& git init && git add . && git commit -am 'base'
-# Hack to get multiline build arg to run properly.
+# Install Kubebuilder tools to run unit tests
```
pintohutch, line 32: I think there's some leftover code comments here. I believe this file shouldn't have a diff from `main`?

pkg/operator/operator.go
```diff
@@ -361,6 +387,12 @@ func (o *Operator) Run(ctx context.Context) error {
 		return errors.Wrap(err, "setup rule-evaluator controllers")
 	}
 
+	if !o.opts.TargetPollDisabled {
+		if err := SetupTargetStatusPoller(ctx, o, registry); err != nil {
```
pintohutch, line 391: nit: `setupTargetStatusPoller` can be a pkg-private function.

pintohutch: I also don't think it needs the `ctx` passed in.

pkg/operator/operator.go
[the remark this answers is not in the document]
pintohutch: Ah nice - agreed!

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1532 @@
...
+			return cmp.Equal(status, expected), nil
+		})
+		if err != nil {
+			if err == wait.ErrWaitTimeout {
+				var podMonitorings monitoringv1.PodMonitoringList
+				kubeClient.List(ctx, &podMonitorings)
+				if len(podMonitorings.Items) == 0 {
+					t.Fatalf("Could not find %s PodMonitoring", description)
+				} else {
+					status := podMonitorings.Items[0].Status.EndpointStatuses
```
pintohutch, line 1292: I don't think this `else` statement could ever happen?

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1532 @@
...
+			default:
+				return false, errors.Errorf("invalid PodMonitorings found: %d", amount)
+			}
+
+			status := podMonitorings.Items[0].Status.EndpointStatuses
+			normalizeEndpointStatuses(status, metav1.Time{})
+			return cmp.Equal(status, expected), nil
+		})
+		if err != nil {
+			if err == wait.ErrWaitTimeout {
```
pintohutch, line 1286: With the fake kubeClient, is there really any info we can get out of here?

i.e. what do we lose by changing this to:
```go
if err != nil {
	t.Fatalf("failed waiting for %s status: %s", desc, err)
}
```

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1532 @@
...
+
+	expectStatus := func(t *testing.T, description string, expected []monitoringv1.ScrapeEndpointStatus) {
+		// Must poll because status is updated via other thread.
+		err := wait.Poll(100*time.Millisecond, 2*time.Second, func() (done bool, err error) {
+			var podMonitorings monitoringv1.PodMonitoringList
+			kubeClient.List(ctx, &podMonitorings)
+			switch amount := len(podMonitorings.Items); amount {
+			case 0:
+				return false, nil
+			case 1:
```
pintohutch, line 1275: why not just:
```go
case 1:
	status := podMonitorings.Items[0].Status.EndpointStatuses
	normalizeEndpointStatuses(status, metav1.Time{})
	return cmp.Equal(status, expected), nil
```

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1532 @@
...
+
+			if len(targets) != amount {
+				t.Fatalf("Wrong target amount: %s", cmp.Diff(len(targets), amount))
+			}
+
+		Outer:
+			for targetIndex, targetExpected := range targetsExpected {
+				for _, foundTarget := range targets {
+					if cmp.Equal(targetExpected, foundTarget) {
+						break Outer
```
pintohutch, line 1525: Why not just `return`?

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1532 @@
...
+	if err := opts.defaultAndValidate(logger); err != nil {
+		t.Fatal("Invalid options:", err)
+	}
+
+	scheme, err := getScheme()
+	if err != nil {
+		t.Fatal("Unable to get scheme")
+	}
+
+	testData := []uint16{0, 1, 2, concurrency - 1, concurrency, concurrency + 1, concurrency * 3}
```
pintohutch, line 1430: What does the 0 case test?

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1532 @@
...
+						PodIP: fmt.Sprint(i),
+						ContainerStatuses: []corev1.ContainerStatus{{
+							Name:  "prometheus",
+							Ready: true,
+						}},
+					},
+				}
+				kubeClientBuilder.WithObjects(pod)
+
+				prometheusTargetMap[getPodKey(pod, port)] = &prometheusv1.TargetsResult{
```
pintohutch, line 1479: IIUC we can just initialize the target result once and copy it to the expected array.

pintohutch (commented):
Mostly nits and some suggestions in the tests - I think this will be the last round of feedback!

pkg/operator/target_status_test.go
TheSpiritXIII: Actually, this is supposed to `continue` (this is supposed to check the existence of each item in the array in the other array because Go lacks a `contains` method or any other basic operation). I refactored so this instead turns each array into a set and compares the set, allowing us to compare the entire list side by side instead of stopping at the first issue.

pkg/operator/target_status_test.go
TheSpiritXIII: As per https://github.com/GoogleCloudPlatform/prometheus-engine/pull/301#discussion_r1026558674, fixing this test helped me uncover some issues here. The target result was supposed to be different per-pod, so I fixed that.

pkg/operator/target_status_test.go
TheSpiritXIII: Base case where we have no pods. I figured this is a valid test to ensure that having no pods does not freeze the threads or panic anywhere. To make it more clear, I added a comment on this. Do you have any concerns?

pkg/operator/target_status_test.go
TheSpiritXIII: It's when the `cmp.Diff` inside the poll fails. Originally, the status is empty so there is a difference but we expect the status to change asynchronously because the reconciler is fired from a goroutine. To make things simpler, I refactored this code to avoid checking inside and outside the poll.

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1532 @@
...
+	return &amount
+}
+
+func lastError(err string) *string {
+	return &err
+}
+
+var date1 = metav1.Date(2022, time.January, 4, 0, 0, 0, 0, time.UTC)
+
+func getData() []targetConversionData {
```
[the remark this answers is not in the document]
TheSpiritXIII: That's a valid point. I instead called the function to get the cross-product "expand" and moved the test cases inline.

pkg/operator/target_status.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+func isPrometheusContainer(container *corev1.Container) bool {
+	return container.Name == "prometheus"
+}
+
+func getPrometheusPort(container *corev1.Container) *int32 {
+	for _, containerPort := range container.Ports {
+		// In the future, we could fall back to reading the command line args.
+		if containerPort.Name == "prom-metrics" {
```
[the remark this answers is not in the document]
TheSpiritXIII: I moved the variable to a constant in `operator_config` because it's actually used in other places. Is this what you intended or something else?

pintohutch (commented):
A few more nits in the tests and some pings on comments from last review.

pkg/operator/target_status_test.go
TheSpiritXIII: I just ended up sorting one of the arrays by instance (since the other one is already sorted by that).

pkg/operator/target_status_test.go
```diff
@@ -0,0 +1,1524 @@
...
+
+	expectStatus := func(t *testing.T, description string, expected []monitoringv1.ScrapeEndpointStatus) {
+		// Must poll because status is updated via other thread.
+		var err error
+		if pollErr := wait.Poll(100*time.Millisecond, 2*time.Second, func() (done bool, err error) {
+			var podMonitorings monitoringv1.PodMonitoringList
+			kubeClient.List(ctx, &podMonitorings)
+			switch amount := len(podMonitorings.Items); amount {
+			case 0:
+				err = fmt.Errorf("Could not find %s PodMonitoring", description)
```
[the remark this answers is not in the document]
TheSpiritXIII: No, it is necessary because the reconcile happens in a separate goroutine. We have a goroutine that ingests the events and calls reconcile, emulating the native controller runtime event handler.

pkg/operator/target_status_test.go
TheSpiritXIII: I have something very similar to this. As per https://github.com/GoogleCloudPlatform/prometheus-engine/pull/301/files#r1028189577 I cannot get rid of the Poll. Do you have any other ideas in mind?

pintohutch: Actually your way provides more information if there is a failure, so nevermind. LGTM.

pintohutch (approved):
LGTM!!

pintohutch (approved):
thanks for the additional unit test!

PR #306, opened 2022-08-15
Update alertmanager to use GKE-built image

damemi:
Now that Louhi builds are set up, this change will be necessary to use the Google-built images in managed AM deployment

### DIFF
--- a/cmd/operator/deploy/operator/12-alertmanager.yaml
+++ b/cmd/operator/deploy/operator/12-alertmanager.yaml
@@ -60,7 +60,7 @@ spec:
     spec:
       containers:
       - name: alertmanager
-        image: prom/alertmanager:latest
+        image: gke.gcr.io/prometheus-engine/alertmanager:v0.24.0-gke.0
         args:
         - --config.file=/alertmanager/config_out/config.yaml
         - --storage.path=/alertmanager-data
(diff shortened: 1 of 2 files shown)


cmd/operator/deploy/operator/12-alertmanager.yaml
```diff
@@ -60,7 +60,7 @@ spec:
     spec:
       containers:
       - name: alertmanager
-        image: prom/alertmanager:latest
+        image: gcr.io/gke-release-staging/prometheus-engine/alertmanager:v0.0.0-gmp.1-gke.0
```
damemi, line 63: obviously we won't want to release v0.0.0 (or staging), just opening this to show the change needed

cmd/operator/deploy/operator/12-alertmanager.yaml
```diff
@@ -60,7 +60,7 @@ spec:
     spec:
       containers:
       - name: alertmanager
-        image: prom/alertmanager:latest
+        image: gke.gcr.io/prometheus-engine/alertmanager:v0.24.0-gke.0
```
damemi, line 63: this image won't exist until promoted

pintohutch:
This was done in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/320 (and the image is now public). Closing.

PR #311, opened 2022-08-24
Reduce ClusterRole

TheSpiritXIII:
Constrains many of our ClusterRoles to a namespaced Role where possible (e.g. where we were already filtering by gmp-public or gmp-system). Further, this change updates the `controller-runtime` manager's watch-list to only watch those namespaces improving security and performance.

### DIFF
--- a/cmd/operator/deploy/operator/03-role.yaml
+++ b/cmd/operator/deploy/operator/03-role.yaml
@@ -45,7 +45,7 @@ rules:
 - apiGroups: [""]
   resources:
   - secrets
-  resourceNames: ["collection", "rules"]
+  resourceNames: ["collection", "rules", "alertmanager"]
   verbs: ["get", "patch", "update"]
 - apiGroups: [""]
   resources:
@@ -70,6 +70,11 @@ rules:
   - deployments
   resourceNames: ["collector", "rule-evaluator"]
   verbs: ["get", "delete", "patch", "update"]
+- apiGroups: ["apps"]
+  resources:
+  - statefulsets
+  resourceNames: ["alertmanager"]
+  verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
@@ -84,10 +89,6 @@ rules:
 - apiGroups: ["monitoring.googleapis.com"]
   resources:
   - operatorconfigs
-  verbs: ["get", "list", "watch"]
-- apiGroups: ["apps"]
-  resources:
-  - statefulsets
   verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/cmd/operator/deploy/operator/03-role.yaml
+++ b/cmd/operator/deploy/operator/03-role.yaml
@@ -45,34 +45,37 @@ rules:
 - apiGroups: [""]
   resources:
   - configmaps
-  verbs: ["create", "patch", "update"]
+  verbs: ["get", "list", "watch", "create", "patch", "update"]
 - apiGroups: ["apps"]
   resources:
   - daemonsets
-  verbs: ["create", "delete", "patch", "update"]
+  - deployments
+  verbs: ["get", "list", "watch", "delete", "patch", "update"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: gmp-public
+  name: operator
+rules:
+- apiGroups: [""]
+  resources:
+  - secrets
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["monitoring.googleapis.com"]
+  resources:
+  - operatorconfigs
+  verbs: ["get", "list", "watch"]
 - apiGroups: ["apps"]
   resources:
-  - deployments
-  verbs: ["create", "delete", "patch", "update"]
+  - statefulsets
+  verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: gmp-system:operator
 rules:
-# The following read permissions are cluster-wide even though they could be namespace-constrained
-# as the controller-runtime library does not allow mixing cluster-wide and namespaced watches.
-- apiGroups: [""]
-  resources:
-  - configmaps
-  - secrets
-  verbs: ["get", "list", "watch"]
-- apiGroups: ["apps"]
-  resources:
-  - daemonsets
-  - deployments
-  - statefulsets
-  verbs: ["get", "list", "watch"]
 # Permission to inject CA bundles into webhook configs of fixed name.
 - apiGroups: ["admissionregistration.k8s.io"]
   resources:
@@ -96,7 +99,6 @@ rules:
   - clusterpodmonitorings
   - clusterrules
   - globalrules
-  - operatorconfigs
   - podmonitorings
   - rules
   verbs: ["get", "list", "watch"]
--- a/cmd/operator/deploy/operator/03-role.yaml
+++ b/cmd/operator/deploy/operator/03-role.yaml
@@ -41,16 +41,35 @@ rules:
 - apiGroups: [""]
   resources:
   - secrets
-  verbs: ["create", "patch", "update"]
+  verbs: ["list", "watch", "create"]
+- apiGroups: [""]
+  resources:
+  - secrets
+  resourceNames: ["collection", "rules"]
+  verbs: ["get", "patch", "update"]
 - apiGroups: [""]
   resources:
   - configmaps
-  verbs: ["get", "list", "watch", "create", "patch", "update"]
+  verbs: ["list", "watch", "create"]
+- apiGroups: [""]
+  resources:
+  - configmaps
+  resourceNames: ["collector", "rule-evaluator", "rules-generated"]
+  verbs: ["get", "patch", "update"]
 - apiGroups: ["apps"]
   resources:
   - daemonsets
-  - deployments
+  resourceNames: ["collector"]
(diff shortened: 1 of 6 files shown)


pkg/operator/operator.go
```diff
@@ -157,14 +160,60 @@ func New(logger logr.Logger, clientConfig *rest.Config, registry prometheus.Regi
...
 		// Don't run a metrics server with the manager. Metrics are being served
 		// explicitly in the main routine.
 		MetricsBindAddress: "0",
-		CertDir:            certDir,
+		// Manage cluster-wide and namespace resources at the same time.
+		NewCache: cache.NewCacheFunc(func(config *rest.Config, options cache.Options) (cache.Cache, error) {
+			return cache.New(clientConfig, cache.Options{
+				Scheme: options.Scheme,
+				// The presence of metadata.namespace has special handling internally causing the
+				// cache's watch-list to only watch that namespace.
```
damemi, line 175: do you have a link to the docs on this?

TheSpiritXIII: It's an optimization in the code: https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.12/pkg/cache/internal/informers_map.go#L294

That line looks for the presence of a metadata.namespace field in the selectors. If it exists (or the root manager specified a namespace, which we do not), then the watch only applies for that namespace.

manifests/operator.yaml
```diff
@@ -75,32 +75,37 @@ rules:
...
+  verbs: ["get", "list", "watch", "create", "patch", "update"]
 - apiGroups: ["apps"]
   resources:
   - daemonsets
-  verbs: ["create", "delete", "patch", "update"]
-- apiGroups: ["apps"]
-  resources:
   - deployments
-  verbs: ["create", "delete", "patch", "update"]
+  verbs: ["get", "list", "watch", "delete", "patch", "update"]
```
vinayakankugoyal, line 83: are there particular deployments that you want to delete, patch and update?

cmd/operator/deploy/operator/03-role.yaml
```diff
@@ -45,34 +45,37 @@ rules:
 - apiGroups: [""]
   resources:
   - configmaps
-  verbs: ["create", "patch", "update"]
+  verbs: ["get", "list", "watch", "create", "patch", "update"]
```
vinayakankugoyal, line 48: are there particular configmaps that you patch and update?

manifests/operator.yaml
TheSpiritXIII: Just within the operator's own namespace (`gmp-system`). We delete, patch and update [as part of a cleanup](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/operator/operator.go#L309).

vinayakankugoyal: right but do we know before hand the name of the deployment? If so we can restrict this even more by using resourceNames.

cmd/operator/deploy/operator/03-role.yaml
vinayakankugoyal: Do we know before hand the name of the deployment? If so we can restrict this even more by using resourceNames.

pkg/operator/operator.go
```diff
@@ -157,14 +160,60 @@ func New(logger logr.Logger, clientConfig *rest.Config, registry prometheus.Regi
...
+					},
+					&monitoringv1.Rules{}: {
+						Field: fields.Everything(),
+					},
+					&corev1.Secret{}: {
+						// We also update secrets from `opts.OperatorNamespace` but never through a
+						// controller. Caches restrict namespaces to either 1 or none so the client
+						// returned by this manager (and delegating to this cache) cannot fetch
+						// secrets other than this namespace `PublicNamespace`.
+						Field: fields.SelectorFromSet(fields.Set{"metadata.namespace": opts.PublicNamespace}),
```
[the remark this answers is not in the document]
TheSpiritXIII: I made an additional commit here that reconciles the gmp-system Secret. Please take a look.

cmd/operator/deploy/operator/03-role.yaml
TheSpiritXIII: I reviewed all gmp-system resource names for each resource type and updated them. Lists and watches have to be under all because you cannot watch multiple resource names in Kubernetes and create must be under all as well because you cannot give create permissions with a name a priori. All other verbs are covered, though.

cmd/operator/deploy/operator/03-role.yaml
```diff
@@ -41,38 +41,60 @@ rules:
 - apiGroups: [""]
   resources:
   - secrets
-  verbs: ["create", "patch", "update"]
+  verbs: ["get", "list", "watch", "create"]
```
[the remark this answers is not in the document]
TheSpiritXIII: I think you're absolutely right. Please take another look!

cmd/operator/deploy/operator/03-role.yaml
```diff
@@ -41,38 +41,60 @@ rules:
...
+  resources:
+  - configmaps
+  resourceNames: ["collector", "rule-evaluator", "rules-generated"]
+  verbs: ["get", "patch", "update"]
 - apiGroups: ["apps"]
   resources:
   - daemonsets
-  verbs: ["create", "delete", "patch", "update"]
+  resourceNames: ["collector"]
+  verbs: ["get", "list", "watch", "delete", "patch", "update"]
```
vinayakankugoyal, line 63: I believe you will need to break this rule similar to what you did for the others?
```
- apiGroups: ["apps"]
   resources:
   - daemonsets
   resourceNames: ["collector"]
   verbs: ["get", "delete", "patch", "update"]
- apiGroups: ["apps"]
   resources:
   - daemonsets
   verbs: ["list", "watch"]
  ```

Comment (TheSpiritXIII) [reply cmd/operator/deploy/operator/03-role.yaml:63]: You can watch on a single resource name or all but not any other discrete value. I had to break up the other rules because they had more than 1 `resourceNames`. That's why in this same commit I made a code change to watch only the `collector` `daemonset` resource name.

Comment (vinayakankugoyal) [reply cmd/operator/deploy/operator/03-role.yaml:63]: oh gotcha! thanks!

**Review by vinayakankugoyal — APPROVED** [review]

File: pkg/operator/operator_config.go [hunk 7]
```diff
@@ -109,24 +114,23 @@ func setupOperatorConfigControllers(op *Operator) error {
...
-		// As the set of secrets to watch is not static, we've to watch them all.
-		// A viable alternative could be for the rule-evaluator (or a sidecar) to generate
-		// the configuration locally and maintain a more constrained watch.
 		Watches(
-			&source.Kind{Type: &corev1.Secret{}},
+			source.NewKindWithCache(&corev1.Secret{}, op.managedNamespacesCache),
 			enqueueConst(objRequest),
 			builder.WithPredicates(predicate.NewPredicateFuncs(secretFilter(op.opts.PublicNamespace))),
 		).
+		// Detect and undo changes to the secret.
```
damemi, line 129: this also needs to reconcile the configmaps that the collector writes to gmp-system, such as the generated rule-evaluator config

damemi (changes requested):

pkg/operator/collection.go
```diff
@@ -66,6 +66,11 @@ func setupCollectionControllers(op *Operator) error {
 		namespace: op.opts.OperatorNamespace,
 		name:      NameCollector,
 	}
+	// Collector secret.
+	objFilterSecret := namespacedNamePredicate{
+		namespace: op.opts.OperatorNamespace,
+		name:      CollectionSecretName,
```
damemi, line 72: does this limit the operator to just watching the collector secret in gmp-system? what about the others, such as the new alertmanager secret?

cmd/operator/deploy/operator/03-role.yaml
```diff
@@ -41,38 +41,60 @@ rules:
 - apiGroups: [""]
   resources:
   - secrets
-  verbs: ["create", "patch", "update"]
+  verbs: ["list", "watch", "create"]
+- apiGroups: [""]
+  resources:
+  - secrets
+  resourceNames: ["collection", "rules"]
```
damemi, line 48: there's `alertmanager` now too

pkg/operator/collection.go
TheSpiritXIII: Added.

cmd/operator/deploy/operator/03-role.yaml
TheSpiritXIII: Done.

pkg/operator/operator_config.go
damemi: sorry, this controller doesn't write any configmaps. disregard

damemi (approved):
looks good, thanks @TheSpiritXIII!
I tested this locally just for sanity, but as a follow up it would be good to add some e2e tests on that target resource reconciliation (ie that the generated files are recreated if someone messes with them)

pkg/operator/collection.go
```diff
@@ -104,6 +109,11 @@ func setupCollectionControllers(op *Operator) error {
 				objFilterCollector,
 				predicate.GenerationChangedPredicate{},
 			)).
+		// Detect and undo changes to the secret.
+		Watches(
+			source.NewKindWithCache(&corev1.Secret{}, op.managedNamespacesCache),
```
pintohutch, line 114: Is using a dedicated, separate informer cache necessary when you've defined the constrained one already [here](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/8e89c873d3bc21f735941b968461c5e48c153d9f/pkg/operator/operator.go#L177)? I'm assuming that's the cache that's injected by default.

PR #312, opened 2022-08-25
Revert "[CI/CD/E2E] Add a fake metrics server for tests using a Kind" -- Moving this to a feature branch

realschwa:
We should revert out the fake metrics server code since it is not being used at the moment since the broader CI/CD project got deprioritized in favor of other projects. 

I will be moving this code to a feature branch and making PR's against that branch to keep the main branch clean.

#270 is the original PR to add this code in.


damemi (approved):

damemi:
For reference could you just link the original PR in the description of this one?

PR #314, opened 2022-08-30
use AM service port at runtime

pintohutch:
Avoids having a divergence between the provided [manifest port](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/8e89c873d3bc21f735941b968461c5e48c153d9f/cmd/operator/deploy/operator/12-alertmanager.yaml#L33) if users make changes.


damemi (approved):

pkg/operator/operator_config.go
```diff
@@ -455,12 +455,18 @@ func (r *operatorConfigReconciler) makeAlertmanagerConfigs(ctx context.Context,
 		Namespace: r.opts.OperatorNamespace,
 		Name:      NameAlertmanager,
 	}
-	// if the default Alertmanager exists, append it to the list of spec.Alertmanagers
-	if resourceErr := r.client.Get(ctx, amNamespacedName, &appsv1.StatefulSet{}); resourceErr == nil {
+	// If the default Alertmanager exists, append it to the list of spec.Alertmanagers.
+	var amSvc corev1.Service
+	if resourceErr := r.client.Get(ctx, amNamespacedName, &amSvc); resourceErr == nil {
+		// Assume first port on service is the correct endpoint.
+		port := amSvc.Spec.Ports[0].Port
```
xichen2020, line 462: nit: Might want to be a bit more defensive and check for `len(Ports)` to protect against the edge case someone defines an `ExternalName` service with the same name.

PR #323, opened 2022-09-19
Use pod-level seccompProfile for multiple resources

StevenYCChou:


pintohutch (changes requested):
Let's remove the container-level seccompProfiles then.

StevenYCChou:
> Let's remove the container-level seccompProfiles then.

Removed. PTAL.

PR #329, opened 2022-09-22
add safe eviction annotation

pintohutch:
Mimics our [collector](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/f753d7786f64c12be9b8c4408f2ab2c67705423f/cmd/operator/deploy/operator/10-collector.yaml#L31-L33) annotations for the Prometheus example.

### DIFF
--- a/examples/prometheus.yaml
+++ b/examples/prometheus.yaml
@@ -59,7 +59,9 @@ metadata:
   name: prometheus-test
   annotations:
     # The emptyDir for the storage and config directories prevents cluster
-    # autoscaling unless this annotation is set.
+    # autoscaling unless this annotation is set. See
+    # https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node
+    # for details.
     cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
   labels:
     prometheus: test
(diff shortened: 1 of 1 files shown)


examples/prometheus.yaml
```diff
@@ -57,6 +57,10 @@ kind: StatefulSet
 metadata:
   namespace: gmp-test
   name: prometheus-test
+  annotations:
+    # The emptyDir for the storage and config directories prevents cluster
+    # autoscaling unless this annotation is set.
```
StevenYCChou, line 62:
```suggestion
    # autoscaling unless this annotation is set. See
    # https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node
    # for details.
```

StevenYCChou: Suggest adding comments for context. Also suggesting adding the relative context in https://github.com/GoogleCloudPlatform/prometheus-engine/blob/f753d7786f64c12be9b8c4408f2ab2c67705423f/cmd/operator/deploy/operator/10-collector.yaml#L31-L33 .

Let me know if you have any objections.

StevenYCChou (changes requested):

StevenYCChou (approved):
Approve first as the current change already LGTM.

PR #330, opened 2022-09-23
Pass in Prometheus registry to e2e tests

TheSpiritXIII:
To prevent NPEs when it's used

### DIFF
--- a/pkg/operator/e2e/context.go
+++ b/pkg/operator/e2e/context.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap/zapcore"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/e2e/context.go
```diff
@@ -97,7 +98,7 @@ func newTestContext(t *testing.T) *testContext {
 		t.Fatalf("create test namespace: %s", err)
 	}
 
-	op, err := operator.New(globalLogger, kubeconfig, nil, operator.Options{
+	op, err := operator.New(globalLogger, kubeconfig, prometheus.NewRegistry(), operator.Options{
```
pintohutch, line 101: I'm fine with it in the e2e tests - but I don't see it being used anywhere in the operator currently. Did you add that somewhere?

pintohutch (approved):
% nit

pkg/operator/e2e/context.go
TheSpiritXIII: Used in #301

PR #332, opened 2022-09-28
Configure Makefile to use buildx for arm64

macxamin:
1. Updated Makefile to use buildx for arm64 support
2. Updated operator.yaml to support arm64 
- Changed gke.gcr.io/gke-distroless/bash to gke.gcr.io/gke-distroless/bash:20220419 for arm64 support 
- Updated nodeSelector to nodeAffinity

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,14 @@ define docker_build
 	DOCKER_BUILDKIT=1 docker build $(1)
 endef
 
-define docker_tag_push
-	docker tag $(1) $(2)
-	docker push $(2)
+# Install QEMU packages required to build ARM images
+# Create a multi-arch builder if one doesn't already exist (default builder doesn't support multi-arch).
+define start_buildx
+	if ! docker buildx inspect multi-arch-builder; then \
+		docker run --rm --privileged multiarch/qemu-user-static --reset --credential yes --persistent yes; \
+		docker buildx create --name multi-arch-builder --use; \
+	fi
+	docker buildx use multi-arch-builder
 endef
 
 help:        ## Show this help.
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -18,6 +18,18 @@ define docker_build
 	DOCKER_BUILDKIT=1 docker build $(1)
 endef
 
+define docker_buildx
+	docker buildx build $(1) --platform linux/arm64,linux/amd64
+endef
+
+define start_buildx
+	if docker buildx inspect multi-arch-builder | grep -q "no builder "; then \
```
pintohutch, line 26: this first step wasn't working for me
```
% make bin
>> building binaries
if docker buildx inspect multi-arch-builder | grep -q "no builder "; then docker run --rm --privileged multiarch/qemu-user-static --reset --credential yes --persistent yes; docker buildx create --name multi-arch-builder --use; fi
error: no builder "multi-arch-builder" found
docker buildx use multi-arch-builder
error: failed to find instance "multi-arch-builder": open /usr/local/google/home/danielclark/.docker/buildx/instances/multi-arch-builder: no such file or directory
```

Makefile
```diff
@@ -51,7 +63,10 @@ ifeq ($(NO_DOCKER), 1)
 	if [ "$@" = "frontend" ]; then pkg/ui/build.sh; fi
 	CGO_ENABLED=0 go build -tags builtinassets -mod=vendor -o ./build/bin/$@ ./cmd/$@/*.go
 else
-	$(call docker_build, --tag gmp/$@ -f ./cmd/$@/Dockerfile .)
+# Install QEMU packages required to build ARM images
+# Create a multi-arch builder if one doesn't already exist (default builder doesn't support multi-arch).
+	$(call start_buildx)
+	$(call docker_buildx, --tag gmp/$@ -f ./cmd/$@/Dockerfile .)
```
pintohutch, line 69: Let's just delete lines 80-82, ie
```
	mkdir -p build/bin
	@echo ">> exporting built image to local 'build/' dir"
	printf 'FROM scratch\nCOPY --from=gmp/$@ /bin/$@ /$@' | $(call docker_build, -o ./build/bin -)
```

pintohutch (commented):
I'm still reviewing, but here's some stuff to look into in the meantime

Makefile
pintohutch: I think you may be able to change the if statement to simply:
```
if ! docker buildx inspect multi-arch-builder; then \
  docker run --rm --privileged multiarch/qemu-user-static; \
  docker buildx create --name multi-arch-builder --use; \
fi
```

pintohutch:
This may be helpful instead of commenting inline https://github.com/GoogleCloudPlatform/prometheus-engine/pull/335.

pintohutch (approved):

PR #339, opened 2022-09-30
Extract auth function and add auth to UI

mhoran:
This makes authentication for the UI in the browser when `AUTH_USERNAME` and `AUTH_PASSWORD` are set. Otherwise, the UI is unprotected but errors because it cannot access `/api`.

### DIFF
--- a/cmd/frontend/main.go
+++ b/cmd/frontend/main.go
@@ -166,7 +166,7 @@ func authenticate(next http.Handler) http.Handler {
 		if len(username) > 0 && len(password) > 0 {
 			reqUser, reqPass, ok := req.BasicAuth()
 			if !ok {
-				w.Header().Set("WWW-Authenticate", `Basic realm="restricted", charset="UTF-8"`)
+				w.Header().Set("WWW-Authenticate", "Basic")
 				w.WriteHeader(http.StatusUnauthorized)
 				return
 			}
--- a/cmd/frontend/main.go
+++ b/cmd/frontend/main.go
@@ -130,7 +130,7 @@ func main() {
 
 		server := &http.Server{Addr: *listenAddress}
 		http.Handle("/metrics", promhttp.HandlerFor(metrics, promhttp.HandlerOpts{Registry: metrics}))
-		http.Handle("/api/", forward(logger, targetURL, transport))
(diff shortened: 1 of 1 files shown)


mhoran:
Any thoughts on this? I found this necessary so the frontend could be used when auth is enabled. It's been working well for me since September.

cmd/frontend/main.go
```diff
@@ -159,16 +159,14 @@ func main() {
...
-
+func authenticate(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		username := os.Getenv(authUsernameEnv)
 		password := os.Getenv(authPasswordEnv)
 		if len(username) > 0 && len(password) > 0 {
-			level.Info(logger).Log("msg", "AUTH_USERNAME and AUTH_PASSWORD are set, handling request with basic auth")
 			reqUser, reqPass, ok := req.BasicAuth()
 			if !ok {
+				w.Header().Set("WWW-Authenticate", `Basic realm="restricted", charset="UTF-8"`)
```
bwplotka, line 169: Curious, why this is needed in your case?

bwplotka (approved):
LGTM, thanks!

Thanks, it makes sense.

It might surprise users who used UI without auth and without access to API, and I don't have info if this is a common use case (new here). Seems to be UI was unusable with basic auth enabled anyway.

cmd/frontend/main.go
bwplotka: I see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate, but do you have any system that actually uses this other than good practice? (:

bwplotka: My point is that realm restricted is not very useful, is it? 🤔

mhoran:
I removed `realm` from `WWW-Authenticate` as it is optional. However, `WWW-Authenticate` is required with a 401 response for the browser to prompt the user to enter their credentials. Without `WWW-Authenticate` I cannot log in to the frontend even by manually providing the credentials, at least in Firefox.

bwplotka:
Then let's keep it - we were not aware of this. Sounds very useful!

mhoran:
Thanks for the merge! Leaving out the `realm` worked for me, so there was no need to add it back.

PR #346, opened 2022-10-14
Use cloudbuild for multi-arch images

pintohutch:
This was running too slow locally and is usually not needed during normal iteration.

Move the multi-arch images to cloudbuild.

Also add some doc nits in the manifests.


pintohutch:
@TheSpiritXIII friendly ping here 🙂

realschwa (approved):

PR #350, opened 2022-10-21
Switch webhook cert setup from polling to controller

TheSpiritXIII:
The previous implementation did a `sleep` before updating the webhooks. This caused a race condition where if the operator tries to touch a resource which is backed by a webhook before the webhook is updated, the call will fail because the webhook does not have a valid certificate.


pkg/operator/operator.go
```diff
@@ -200,9 +201,7 @@ func New(logger logr.Logger, clientConfig *rest.Config, registry prometheus.Regi
 						// need to access secrets from multiple namespaces, we
 						// specify one here so that the manager's client
 						// accesses secrets from this namespace through a cache.
-						Field: fields.SelectorFromSet(fields.Set{
-							"metadata.namespace": opts.PublicNamespace,
-						}),
+						Field: fields.OneTermEqualSelector("metadata.namespace", opts.PublicNamespace),
 					},
 					&monitoringv1.OperatorConfig{}: {
 						Field: fields.SelectorFromSet(fields.Set{"metadata.namespace": opts.PublicNamespace}),
```
pintohutch, line 207: Looks like you can use `OneTermEqualSelector` here too for the OperatorConfig.

pkg/operator/operator.go
```diff
@@ -269,31 +274,82 @@ func (o *Operator) setupAdmissionWebhooks(ctx context.Context) error {
...
-		}
-		// Initial sleep for the client to initialize before our first calls.
-		// Ideally we could explicitly wait for it.
-		time.Sleep(5 * time.Second)
+	err = ctrl.NewControllerManagedBy(o.manager).
+		Named("validating-webhook-cert-updater").
+		For(
+			&arv1.ValidatingWebhookConfiguration{},
+			builder.WithPredicates(predicate.Funcs{
+				CreateFunc: func(ev event.CreateEvent) bool {
```
pintohutch, line 282: What about using the [`NewPredicateFuncs`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/21478f108665efe083f1ae90cea8f542d3b5044c/vendor/sigs.k8s.io/controller-runtime/pkg/predicate/predicate.go#L100-L103) helper?

pkg/operator/operator.go
```diff
@@ -269,31 +274,82 @@ func (o *Operator) setupAdmissionWebhooks(ctx context.Context) error {
...
-		// Only inject if we've an explicit CA bundle ourselves. Otherwise the webhook configs
-		// may already have been created with one.
-		if len(caBundle) == 0 {
-			return
-		}
-		// Initial sleep for the client to initialize before our first calls.
-		// Ideally we could explicitly wait for it.
-		time.Sleep(5 * time.Second)
+	err = ctrl.NewControllerManagedBy(o.manager).
+		Named("validating-webhook-cert-updater").
```
pintohutch, line 278: I believe you'll need a
```go
		// Filter events without changes for all watches.
		WithEventFilter(predicate.ResourceVersionChangedPredicate{}).
```
to prevent infinite reconciling, no? This is used in our other event-based controllers, e.g. [`operator_config.go`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/21478f108665efe083f1ae90cea8f542d3b5044c/pkg/operator/collection.go#L79-L80).

pkg/operator/operator.go
```diff
@@ -269,31 +274,82 @@ func (o *Operator) setupAdmissionWebhooks(ctx context.Context) error {
...
 			}
+			return reconcile.Result{}, nil
+		}))
+	if err != nil {
+		return errors.Wrap(err, "create mutating webhook updater controller")
+	}
+
+	// This method runs after the webhook server and client are started but
+	// before the controllers are started. Running it here prevents race
+	// conditions.
```
pintohutch, line 337: Can you elaborate more on why this is necessary? What are the specific race conditions you're concerned with?

pintohutch (changes requested):
Nice change overall. Have you tested this out manually to ensure it doesn't break any prior behavior?

pkg/operator/operator.go
TheSpiritXIII: We want to ignore delete events.

PR #354, opened 2022-10-27
Reorganize examples directory

yqlu:
- one subdirectory for every exporter we have documentation for
- add recommended rules when applicable
- use `.yaml.snippet` extension when snippet shows the recommended delta and cannot be applied as-is as standalone yaml
- add minimal README for each subdirectory pointing readers to the docs pages


lyanco:
LGTM. Can you also add directories or READMEs for kubelet/cAdvisor and control plane, even if all they do is point people to the docs?

lyanco (approved):

yqlu:
Yep, done.

PR #356, opened 2022-10-29
Handle histogram with out-of-order buckets

wopol:
This PR is aimed to fix handling histograms with buckets out of the order. Prometheus does not guarantee buckets order. When `+Inf` bucket is not the last bucket we are not able to build distribution because `hasInfBucket` property will be overwritten. The following example cannot be parsed:
```
# HELP custom_query_time Execution time histogram for slow queries [s]
# TYPE custom_query_time histogram
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="+Inf"} 35
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="1000"} 22
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="125"} 18
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="16"} 17
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="250"} 18
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="32"} 18
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="4"} 1
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="500"} 20
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="63"} 18
custom_query_time_bucket{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084",le="8"} 1
custom_query_time_sum{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084"} 205.824447
custom_query_time_count{role="coordinator",server="crdn-vc7b48n9",shortname="Coordinator0084"} 35
```

I believe that order does not matter because we sort built distribution afterward here: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/export/transform.go#L241

### DIFF
--- a/pkg/export/transform.go
+++ b/pkg/export/transform.go
@@ -405,7 +405,10 @@ Loop:
 				prometheusSamplesDiscarded.WithLabelValues("NaN-bucket-value").Inc()
 				continue
 			}
-			dist.hasInfBucket = math.IsInf(bound, 1)
+			// Handle cases where +Inf bucket is out-of-order by not overwriting on the last-consumed bucket.
+			if !dist.hasInfBucket {
+				dist.hasInfBucket = math.IsInf(bound, 1)
+			}
 			dist.bounds = append(dist.bounds, bound)
 			dist.values = append(dist.values, int64(v))
 
--- a/pkg/export/transform_test.go
+++ b/pkg/export/transform_test.go
@@ -1192,6 +1192,79 @@ func TestSampleBuilder(t *testing.T) {
 					}},
 				},
 			},
+		}, {
+			doc: "histogram is not in-order",
+			metadata: testMetadataFunc(metricMetadataMap{
+				"metric1": {Type: textparse.MetricTypeHistogram, Help: "metric1 help text"},
+			}),
+			series: seriesMap{
+				1: labels.FromStrings("job", "job1", "instance", "instance1", "__name__", "metric1_sum"),
+				2: labels.FromStrings("job", "job1", "instance", "instance1", "__name__", "metric1_count"),
+				3: labels.FromStrings("job", "job1", "instance", "instance1", "__name__", "metric1_bucket", "le", "+Inf"),
+				4: labels.FromStrings("job", "job1", "instance", "instance1", "__name__", "metric1_bucket", "le", "1"),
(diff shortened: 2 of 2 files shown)


realschwa (commented):
Hey -- thanks for the PR!

This makes sense to me as a way to handle this case, but if I understand correctly both the [prometheus exposition format](https://github.com/prometheus/docs/blob/9f705a7423f1f6e76ddf2501cad4f9506e7993a7/content/docs/instrumenting/exposition_formats.md#histograms-and-summaries) and the [open metrics format]() require the histogram buckets to be in increasing order of `le` label.

Have you observed the buckets being out of increasing order in your applications? What instrumentation libraries are you using if you have seen this happen?

wopol:
@realschwa fair point. I am experiencing this behavior in [ArangoDB](https://github.com/arangodb/arangodb) metrics endpoint. It looks like they use hand made instrumentation tool written in c++.

pkg/export/transform.go
```diff
@@ -405,7 +405,9 @@ Loop:
 				prometheusSamplesDiscarded.WithLabelValues("NaN-bucket-value").Inc()
 				continue
 			}
-			dist.hasInfBucket = math.IsInf(bound, 1)
+			if !dist.hasInfBucket {
```
pintohutch, line 408: Can you add a comment here, e.g.
```
// Handle cases where +Inf bucket is out-of-order by not overwriting on the last-consumed bucket.
```

wopol: 👌

realschwa (approved):
Gotcha -- If vanilla Prometheus allows this then we should too 🙂

wopol:
@realschwa When could we expect merge and rolling out?

lyanco:
Rolling out to GKE can take a few months, unfortunately. It should be available to deploy via kubectl as soon as it's tagged, likely this month.

PR #361, opened 2022-11-10
Fix action defaulting

pintohutch:
Allow Prometheus to default to `replace` when `RelabelingRule.Action` is unspecified, as per the [docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config).

### DIFF
--- a/pkg/operator/apis/monitoring/v1/types_test.go
+++ b/pkg/operator/apis/monitoring/v1/types_test.go
@@ -539,6 +539,9 @@ func TestPodMonitoring_ScrapeConfig(t *testing.T) {
 							Modulus:      3,
 							TargetLabel:  "__tmp_mod",
 						}, {
+							SourceLabels: []string{"mlabel_4"},
+							TargetLabel:  "mlabel_5",
+						}, {
 							Action:  "keep",
 							Regex:   "foo_.+",
 							Modulus: 3,
@@ -641,6 +644,8 @@ metric_relabel_configs:
   modulus: 3
   target_label: __tmp_mod
   action: hashmod
+- source_labels: [mlabel_4]
+  target_label: mlabel_5
 - regex: foo_.+
   modulus: 3
   action: keep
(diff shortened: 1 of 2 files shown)


pkg/operator/apis/monitoring/v1/types_test.go
```diff
@@ -231,6 +231,22 @@ func TestValidatePodMonitoringCommon(t *testing.T) {
 					},
 				},
 			},
+		}, {
+			desc: "metric relabeling: blank 'action' defaults to 'replace'",
```
StevenYCChou, line 235:
```suggestion
			desc: "metric relabeling: blank 'action' is valid and it defaults to 'replace'",
```

nit - as this test is for validation - not for the actual logic for default, hence the suggestion.

StevenYCChou (approved):
This change itself LGTM based on [the prometheus doc website](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_action). Please address the comment on the test description before merging this PR.

A side question:
1. [this implementation](https://github.com/prometheus/prometheus/blob/b1566f761ef7d1c8ab6446afd017c0ce17e9224e/model/relabel/relabel.go#L106-L108) doesn't allow parsing on the empty string... is this code path utilized in our code? Is the prometheus doc incorrect or the implementation incorrect?

pintohutch:
> A side question:

> [this implementation](https://github.com/prometheus/prometheus/blob/b1566f761ef7d1c8ab6446afd017c0ce17e9224e/model/relabel/relabel.go#L106-L108) doesn't allow parsing on the empty string... is this code path utilized in our code? Is the prometheus doc incorrect or the implementation incorrect?

Great question @StevenYCChou! We actually do validation through YAML marshal/unmarshal calls when [building](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/f4d668267dee2da80ad99fe6ffc251784613abc2/pkg/operator/apis/monitoring/v1/types.go#L677-L688) the final scrape config.

AFAICT the reason we don't hit that case is because that method [uses](https://github.com/GoogleCloudPlatform/prometheus/blob/v2.35.0-gmp.2/model/relabel/relabel.go#L95) a [`DefaultRelabelConfig`](https://github.com/GoogleCloudPlatform/prometheus/blob/v2.35.0-gmp.2/model/relabel/relabel.go#L31-L36) that sets the `Action` to "replace" before unmarshalling the live yaml.

I've tested the empty case myself btw and it indeed works.

PR #363, opened 2022-11-15
Delete extra examples manifests

pintohutch:
Delete extra manifests and move example PodMonitoring in-line with app.


examples/example-app.yaml
```diff
@@ -52,3 +52,17 @@ spec:
...
+apiVersion: monitoring.googleapis.com/v1
+kind: PodMonitoring
+metadata:
+  name: prom-example
+  labels:
+    app.kubernetes.io/name: prom-example
+spec:
+  selector:
+    matchLabels:
+      app: prom-example
```
TheSpiritXIII, line 65: Should this selector be `app` or `app.kubernetes.io/name`? It does not match the actual label name. Note, the `Deployment` uses `app`.

yqlu:
Context: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/354#issuecomment-1293924524

Should we hold off for a couple of weeks? The English docs have been updated (https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/node_exporter links to examples/node_exporter/node_exporter.yaml) but I just took a quick look at other languages and the change is still propagating...

Specifically [Deutsch](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters?hl=de) and [Portugese](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters?hl=pt-br) still have the "old" page live that embeds these docs, when the other languages correctly redirect this old page `/docs/managed-prometheus/exporters` --> the new landing page `/docs/managed-prometheus/exporters/introduction`

pintohutch:
Yea - there's no rush here. Happy to let it sit a few weeks.

examples/example-app.yaml
pintohutch: I went with `app.kubernetes.io/name` for everything, as that's convention in our other [examples](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.6.0-rc.0/examples/aerospike/pod-monitoring.yaml#L20).

PR #366, opened 2022-11-18
use more accurate self-metrics

pintohutch:
`samplesExported` is a little confusing - especially with [`samplesSent`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/d268d39aeb5ecdf1277b533b8640a77568978479/pkg/export/export.go#L804). This makes it more clear and accurate.

### DIFF
--- a/pkg/export/export.go
+++ b/pkg/export/export.go
@@ -402,7 +402,7 @@ func (e *Exporter) Export(metadata MetadataFunc, batch []record.RefSample) {
 	e.mtx.Unlock()
 
 	if !ok {
-		prometheusSamplesDiscarded.WithLabelValues("no-ha-range").Add(float64(len(batch)))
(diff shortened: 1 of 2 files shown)


pkg/export/export.go
```diff
@@ -397,7 +402,7 @@ func (e *Exporter) Export(metadata MetadataFunc, batch []record.RefSample) {
 	e.mtx.Unlock()
 
 	if !ok {
-		prometheusSamplesDiscarded.WithLabelValues("no-ha-range").Inc()
+		prometheusSamplesDiscarded.WithLabelValues("no-ha-range").Add(float64(batchSize))
```
realschwa, line 405: should this be `prometheusSamplesDropped` rather than `prometheusSamplesDiscarded`

pintohutch: Yes - done.

realschwa (approved):

PR #368, opened 2022-11-21
Standardize all recommended scrape configs at 30s interval

yqlu:


examples/self-pod-monitoring.yaml
```diff
@@ -26,6 +26,6 @@ spec:
       app.kubernetes.io/name: collector
   endpoints:
   - port: prom-metrics
-    interval: 10s
```
lyanco, line 29: let's keep the self-pod-monitoring at 10s - 1. it's uncommonly used and doesn't include a lot of metrics so w/e and 2. Danny thinks it's set this way because you'd want to know the status of prometheus collectors in between scrapes in case there's a scraping issue

examples/self-pod-monitoring.yaml
```diff
@@ -26,6 +26,6 @@ spec:
...
+    interval: 30s
   - port: cfg-rel-metrics
-    interval: 10s
```
lyanco, line 31: ditto

lyanco (changes requested):

examples/self-pod-monitoring.yaml
yqlu: Got it, reverted

lyanco (approved):

PR #370, opened 2022-11-22
Target status reporting - fix kubelet handling and unit test

pintohutch:
kubelet scrapePools are hardcoded by the operator and don't use a PodMonitoring/ClusterPodMonitoring. We want to be able to handle these without erroring.

- The expanding function in target_status_test.go was a little buggy, so I fixed that.
- Some light renaming and refactoring.

### DIFF
--- a/pkg/operator/target_status.go
+++ b/pkg/operator/target_status.go
@@ -130,7 +130,7 @@ func (r *targetStatusReconciler) Reconcile(ctx context.Context, request reconcil
 
 	now := time.Now()
 	if err := pollAndUpdate(ctx, r.logger, r.opts, r.getTarget, r.kubeClient); err != nil {
-		r.logger.Error(err, "polling targets")
+		r.logger.Error(err, "poll and update")
 	} else {
 		// Only log metrics if target polling was successful.
 		duration := time.Since(now)
@@ -317,6 +317,7 @@ func updateTargetStatus(ctx context.Context, logger logr.Logger, kubeClient clie
 		return err
 	}
 
+	var patchErr error
 	for job, endpointStatuses := range endpointMap {
 		// Kubelet scraping is configured through hard-coding and not through
 		// a PodMonitoring. As there's no status to update, we skip.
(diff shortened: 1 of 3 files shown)


pkg/operator/target_status.go
```diff
@@ -309,23 +309,28 @@ func patchPodMonitoringStatus(ctx context.Context, kubeClient client.Client, obj
...
 		podMonitoringStatusContainer, err := buildPodMonitoring(job)
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "building podmonitoring: %s", job)
 		}
 		podMonitoringStatusContainer.GetStatus().EndpointStatuses = endpointStatuses
 
 		if err := patchPodMonitoringStatus(ctx, kubeClient, podMonitoringStatusContainer, *podMonitoringStatusContainer.GetStatus()); err != nil {
-			logger.Error(err, "error patching job", "job", job)
+			return errors.Wrapf(err, "patching job: %s", job)
```
TheSpiritXIII, line 333: I think the reason I made this a log and not a hard error is in case the PodMonitoring is removed between when we do the poll and we patch the status. We expect the patch to fail but we don't mind. Do you know what the implications for this are?

pkg/operator/target_status_test.go
```diff
@@ -92,29 +90,35 @@ func expandPopulateTargetsTestCases(testCases []populateTargetsTestCase) []popul
...
+			targets:        tc.targets,
+			podMonitorings: tc.podMonitorings,
+			expErr:         tc.expErr,
 		}
-		dataPodMonitorings := populateTargetsTestCase{
-			name:              data.name + "-pod-monitoring",
-			prometheusTargets: data.prometheusTargets,
-			podMonitorings:    data.podMonitorings,
+		dataFinal = append(dataFinal, dataPodMonitorings)
+		if len(clusterPodMonitorings) == 0 {
```
TheSpiritXIII, line 105: It looks like this was originally checking the length of `podMonitorings` and skipping the test, but it looks like you moved this down here because you want to append at least one for the kubelet tests. Does it make sense for the kublet tests to be suffixed with "-pod-monitoring" when there is none? In other words, why don't we move the check back up, change it back to check for `podMonitoring` and if there are none, then we could add the test case maybe without a suffix or with a `-no-pod-monitoring` suffix?

TheSpiritXIII:
Good find! In my original PR, I remember I had some special handling with kublets but I never had a test case for them so I lost it because I wasn't sure how they worked. I left a couple comments for you!

TheSpiritXIII (changes requested):

pkg/operator/target_status_test.go
pintohutch: Yea - that's cleaner than what I did. Done.

TheSpiritXIII (approved):
Looks good. Thanks!

PR #376, opened 2022-11-28
[DO NOT MERGE] Apply Kubernetes YAML files via kustomize

TheSpiritXIII:
This change is a precursor to a few changes:
1. Templating images: Kustomize lets us define images maps. You give Kustomize the original image and then tell Kustomize what to replace it with.
2. Deploying the operator in a Kind cluster. This is really easy to do with #1 -- just make a template to point images to a local repository (`localhost`) and `kubectl apply` that template.
3. Removing `manifest/` altogether. Kustomize can take Github URLs so we can update our public docs to point to our `kustomization.yaml` and have Kustomize combine our YAML files.


pintohutch (changes requested):
Can you try a rebase against `main` to refresh the diff and fix merge conflicts?

hack/presubmit.sh
```diff
@@ -67,15 +67,13 @@ update_codegen() {
...
-  cat "${SCRIPT_ROOT}"/hack/boilerplate.txt > $DEST_YAML
-  printf "\n# NOTE: This file is autogenerated.\n" >> $DEST_YAML
-  sed -s '$a---' $SOURCE_YAMLS | sed -e '$ d' -e '/^#/d' -e '/^$/d' >> $DEST_YAML
+  local source_dir=$1
+  local dest_yaml=$2
+
+  mkdir -p $(dirname $dest_yaml)
+  cat "${SCRIPT_ROOT}"/hack/boilerplate.txt > $dest_yaml
+  printf "\n# NOTE: This file is autogenerated.\n" >> $dest_yaml
+  kustomize build --reorder=none "${source_dir}" >> $dest_yaml
```
pintohutch, line 76: Interesting, it seems kustomize is re-ordering the output despite specifying `--reorder=none`. I suppose this isn't the biggest deal since the generated manifests should only ever be read-only.

pintohutch: I don't have a strong opinion, but running a final `kustomize cfg fmt` on the output files may get them back in order, but I think that can be optional.

pintohutch (approved):
% ordering of the generated files. Up to you if you want to re-format them or just use the generated kustomize output

hack/presubmit.sh
TheSpiritXIII: The reorder is for reording resources, not fields. For example, kustomize tries to put things like namespaces or webhooks first.

pintohutch: Ahh got it. Makes sense.

pintohutch (approved):

TheSpiritXIII:
Closing, as we are now using helm to accomplish this.

PR #383, opened 2022-11-30
GKE Security Posture configuration fixes

TheSpiritXIII:
This fixes the main configurations as well as a few examples. One of the prominent examples I have not covered here is the node-exporter because that has a few issues and I am not familiar with the exporter enough to determine if it would break anything. To be more specific, they have issues with using the host network and volume mounts which I don't believe we could remove.

### DIFF
--- a/cmd/operator/deploy/operator/11-rule-evaluator.yaml
+++ b/cmd/operator/deploy/operator/11-rule-evaluator.yaml
@@ -49,11 +49,16 @@ spec:
                 operator: In
                 values:
                 - linux
-      serviceAccountName: collector
       initContainers:
       - name: config-init
         image: gke.gcr.io/gke-distroless/bash:20220419
         command: ['/bin/bash', '-c', 'touch /prometheus/config_out/config.yaml']
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - all
+          privileged: false
         volumeMounts:
         - name: config-out
           mountPath: /prometheus/config_out
@@ -89,9 +94,6 @@ spec:
             drop:
             - all
           privileged: false
-          runAsGroup: 1000
-          runAsNonRoot: true
-          runAsUser: 1000
         volumeMounts:
         - name: config-out
           mountPath: /prometheus/config_out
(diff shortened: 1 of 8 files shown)


cmd/operator/deploy/operator/11-rule-evaluator.yaml
```diff
@@ -49,11 +49,19 @@ spec:
                 operator: In
                 values:
                 - linux
-      serviceAccountName: collector
```
pintohutch, line 52: You removed this as it's [already part](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/c4f3239d50e536b60ba730e91fce1f9fa3069c93/cmd/operator/deploy/operator/11-rule-evaluator.yaml#L52) of the template spec, correct?

examples/node-exporter.yaml
```diff
@@ -69,6 +69,15 @@ spec:
           requests:
             cpu: 102m
             memory: 180Mi
+        securityContext:
```
pintohutch, line 72: With the exception of the `frontend`, I'm tempted to leave the examples directory out of this change for now.

They're not officially a part of the managed infrastructure and I'd rather not risk breaking anything which our [docs](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/node_exporter) point to.

pintohutch (commented):
Nice! It looks like it was mostly `initContainers` anyway.

I'd say the only other thing to try out - maybe it would be simpler to consolidate `securityContext` at the [pod-leve](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/c4f3239d50e536b60ba730e91fce1f9fa3069c93/cmd/operator/deploy/operator/05-deployment.yaml#L96-L98) instead of the container level?

Though that could always be done later, so no strong opinion.

cmd/operator/deploy/operator/11-rule-evaluator.yaml
TheSpiritXIII: Yes.

pintohutch (approved):
LGTM!

PR #387, opened 2022-12-02
Use kustomize to sort and format yaml manifests

pintohutch:
For reviewing, the only relevant commit is https://github.com/GoogleCloudPlatform/prometheus-engine/pull/387/commits/8e7c400fc3ac2ed97b375249426676efc8b2b0c8.

The second commit https://github.com/GoogleCloudPlatform/prometheus-engine/pull/387/commits/a2bf046d8b216ae6dde35ca9e4af296a1dfa6679 is the actual reformatting.

I left the `examples/` directory out for now, but I suppose we could add that later.


TheSpiritXIII (approved):
I think this looks good! I'm surprised how the field reording turned out. I expected it to be alphabetical but it looks arbitrary.

pintohutch:
Yep - the field ordering is mentioned [here](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/commands/fmt.md), which isn't necessarily alphabetical.

PR #391, opened 2022-12-08
support all sd configs in rule-evaluator

### LINKED ISSUE #385: Issue with rule-evaluator with self deploy alertmanager
When I was following this steps :- https://cloud.google.com/stackdriver/docs/managed-prometheus/rules-unmanaged#eval-rules-unmanaged

I made changes in the `rule-evaluator` configmap 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: rule-evaluator
  namespace: monitoring
  labels:
    app.kubernetes.io/name: rule-evaluator
data:
  config.yaml: |
    global:
      external_labels: 
        test: global-rule-evaluator
      evaluation_interval: 30s
    alerting: 
      alertmanagers:
        - dns_sd_configs:
          - names:
            - alertmanager.monitoring.svc.cluster.local
          port: 9093
          type: A
    rule_files:
    - "/etc/rules/*.yaml"
```

After this changes i deploy the rule-evaluator deployment so that if any alert will trigger it will go to my self deploy alertmanager 

After this deployment i was seeing this error in the logs of `rule-evaluator` pods 

```
level=info ts=2022-12-01T13:06:27.137865623Z caller=main.go:450 msg="Loading configuration file" filename=/etc/config/config.yaml
level=error ts=2022-12-01T13:06:27.138282653Z caller=main.go:375 msg="Error reloading config" err="couldn't load configuration (--config.file=\"/etc/config/config.yaml\"): parsing YAML file /etc/config/config.yaml: yaml: unmarshal errors:\n  line 7: field dns_sd_configs not found in type config.AlertmanagerConfig\n  line 10: field port not found in type config.AlertmanagerConfig\n  line 11: field type not found in type config.AlertmanagerConfig"
```

Here it says that its parsing error and `dns_sd_configs not found in type config.AlertmanagerConfig` but as per the prometheus documentation :- https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config
The configuration is there as i missing something here

image that i am using for the rule-evaluator :- `gke.gcr.io/prometheus-engine/rule-evaluator:v0.5.0-gke.0`

if you want to see the deployment file as well let me know i can add that file here as well but its the same which is mention over here :- https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/manifests/rule-evaluator.yaml

pintohutch:
Tentatively fixes #385

### DIFF
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -48,7 +48,7 @@ import (
 	"github.com/prometheus/prometheus/discovery"
 
 	// Import to enable 'kubernetes_sd_configs' to SD config register.
-	_ "github.com/prometheus/prometheus/discovery/kubernetes"
+	_ "github.com/prometheus/prometheus/discovery/install"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/notifier"
 	"github.com/prometheus/prometheus/promql"
--- a/go.mod
+++ b/go.mod
@@ -39,19 +39,40 @@ require (
 )
 
 require (
+	github.com/Azure/azure-sdk-for-go v63.0.0+incompatible // indirect
+	github.com/Azure/go-autorest v14.2.0+incompatible // indirect
+	github.com/Azure/go-autorest/autorest v0.11.25 // indirect
+	github.com/Azure/go-autorest/autorest/adal v0.9.18 // indirect
+	github.com/Azure/go-autorest/autorest/date v0.3.0 // indirect
+	github.com/Azure/go-autorest/autorest/to v0.4.0 // indirect
+	github.com/Azure/go-autorest/autorest/validation v0.3.1 // indirect
+	github.com/Azure/go-autorest/logger v0.2.1 // indirect
+	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
+	github.com/Microsoft/go-winio v0.5.1 // indirect
 	github.com/PuerkitoBio/purell v1.1.1 // indirect
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137 // indirect
+	github.com/armon/go-metrics v0.3.3 // indirect
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d // indirect
 	github.com/aws/aws-sdk-go v1.43.31 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
+	github.com/cncf/xds/go v0.0.0-20211011173535-cb28da3451f1 // indirect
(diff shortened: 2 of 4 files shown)


StevenYCChou:
First, I believe in #385 , static config is fine in the end (see https://github.com/GoogleCloudPlatform/prometheus-engine/issues/385#issuecomment-1343977338 ). Therefore, we don't need to support all sd configs, including  `dns_sd_configs` mentioned in #385 . 

Second, I am hesitant adding any new service discovery libraries, as I don't know whether Cloud Monitoring APIs can support target labels beyond the one owned by kubernetes, given the [`prometheus_target`](https://cloud.google.com/monitoring/api/resources#tag_prometheus_target) only has labels related to k8s. I'd like to learn more if our backend support other kinds of labels in a good way - so I suggest that we take a step back and understand it first before keeping working/reviewing this PR.

StevenYCChou:
Also, I am not sure if #385 is really a bug (if it doesn't work for the existing cases), or if we categorize it as a feature request to expand the functionality of `rule-evaluator` from discover k8s targets only to other targets.

pintohutch:
Good callout on categorization. I relabeled it from `bug` to `enhancement`.

The primary reason I went and imported everything was so the rule-evaluator could be run in any environment supported by Prometheus and discover Alertmanagers there. It's likely pretty long-tail in terms of actual usage, as I imagine `static_config` and `kubernetes_sd_config` covers nearly all current use cases. Though I'd imagine we'd want to _at least_ support `gce_sd_config` for any users running the `rule-evaluator` with an Alertmanager in GCE. But at that point, it's almost more work to downselect which environments we support for Alertmanager discovery. cc @lyanco 

The `relabel_configs` that concern the rule-evaluator configuration are filtering Alertmanager targets, not for ingesting time series data. Labels for ingested time series to the Cloud Monitoring API are configured on the [rule spec](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#rule), which would be hardcoded into the config, or [templated](https://prometheus.io/docs/prometheus/latest/configuration/template_examples/) using [labels](https://prometheus.io/docs/prometheus/latest/configuration/template_reference/) already-existing on queried time series.

Unless I'm misunderstanding something?

StevenYCChou:
We synced offline, it's lower priority so we'll hold off reviewing/working on this PR.

My comment for the latest comment is: 
- When we need `gce_sd_config`, then we add it then.
- Adding too many unnecessary service discovery config may make us harder to maintain in the long run, so we should hold off until there is a real need to include a service discovery config.
- I think I misunderstood how service discovery is used here, so apologize for comment related to metrics ingestion as they seem not relevant. 
- I think the `relabel_configs` is orthogonal to the alertmanager's service discovery config here - were you commenting on my misunderstanding to the metrics ingestion? feel free to reply or wait until we need to revive this PR later.

pintohutch:
> I think the relabel_configs is orthogonal to the alertmanager's service discovery config here - were you commenting on my misunderstanding to the metrics ingestion? feel free to reply or wait until we need to revive this PR later.

Yea - I was highlighting how the Alertmanager `relabel_configs` are not used to relabel time series for ingestion. It is used as documented [here](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config):
> Additionally, relabel_configs allow selecting Alertmanagers from discovered entities and provide advanced modifications to the used API path, which is exposed through the __alerts_path__ label.

PR #394, opened 2022-12-08
Use lower requests

pintohutch:
Specify the workload requests of CPU/RAM to reflect what they use at idle.

### DIFF
--- a/cmd/operator/deploy/operator/05-deployment.yaml
+++ b/cmd/operator/deploy/operator/05-deployment.yaml
@@ -58,7 +58,7 @@ spec:
           containerPort: 18080
         resources:
           limits:
-            memory: 256M
+            memory: 512M
           requests:
             cpu: 8m
             memory: 32M
--- a/cmd/operator/deploy/operator/05-deployment.yaml
+++ b/cmd/operator/deploy/operator/05-deployment.yaml
@@ -56,6 +56,12 @@ spec:
           # Note this should match the --metrics-addr flag passed in to the operator args.
           # Default is 18080.
           containerPort: 18080
+        resources:
+          limits:
+            memory: 256M
+          requests:
+            cpu: 8m
+            memory: 32M
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
--- a/cmd/operator/deploy/operator/10-collector.yaml
+++ b/cmd/operator/deploy/operator/10-collector.yaml
@@ -118,7 +118,7 @@ spec:
             memory: 2G
           requests:
             cpu: 16m
-            memory: 32M
+            memory: 128M
         volumeMounts:
         - name: storage
           mountPath: /prometheus/data
--- a/cmd/operator/deploy/operator/10-collector.yaml
+++ b/cmd/operator/deploy/operator/10-collector.yaml
@@ -115,7 +115,7 @@ spec:
           value: "25"
         resources:
           limits:
-            memory: 3G
+            memory: 2G
           requests:
             cpu: 16m
             memory: 32M
--- a/cmd/operator/deploy/operator/10-collector.yaml
+++ b/cmd/operator/deploy/operator/10-collector.yaml
@@ -72,7 +72,7 @@ spec:
           limits:
             memory: 32M
           requests:
-            cpu: 2m
+            cpu: 8m
             memory: 16M
         volumeMounts:
         - name: config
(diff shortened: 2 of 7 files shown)


pintohutch:
Sure - lemme add some documentation. I added comments to a few containers, but not all of them.

The method I used was very unscientific. I looked at the max CPU/RAM usage over the past 6 weeks for all of the containers across several clusters (both at idle and during usage), and adjusted the numbers here to match what I saw.

Ultimately, we should have some benchmarking, particularly for the collectors, to guide these numbers in the future.

cmd/operator/deploy/operator/05-deployment.yaml
```diff
@@ -56,6 +56,12 @@ spec:
           # Note this should match the --metrics-addr flag passed in to the operator args.
           # Default is 18080.
           containerPort: 18080
+        resources:
```
StevenYCChou, line 59: Setting resource limits/requests here is great and nice.

cmd/operator/deploy/operator/10-collector.yaml
```diff
@@ -117,8 +117,10 @@ spec:
           limits:
             memory: 3G
           requests:
-            cpu: 100m
-            memory: 200M
+            # Resource requests were empirically determined to be in this range
+            # when running idle (i.e. no scraping).
+            cpu: 2m
```
StevenYCChou, line 122: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run

says

```
The CPU request typically defines a weighting. If several different containers (cgroups) want to run on a contended system, workloads with larger CPU requests are allocated more CPU time than workloads with small requests.

```

So I am concerned that if we set CPU request for the IDLE phase, the CPU may be taken away by other workloads.

cmd/operator/deploy/operator/05-deployment.yaml
pintohutch: Cool!

pintohutch:
For CPU requests, to address Yen's review [comment](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/394#discussion_r1046355704), I took the p90 of CPU utilization for each container and rounded up to the nearest power of 2. This should help mitigate under-allocation during contention.

For memory requests, I took the p90 of memory utilization of each container and rounded up to the nearest power of 2.

For the memory limits, I took the p999 of utilization and rounded up to the nearest power of 2. This ended up lowering most of the limits.

cmd/operator/deploy/operator/10-collector.yaml
pintohutch: See https://github.com/GoogleCloudPlatform/prometheus-engine/pull/394#issuecomment-1347421158

shuaich:
What is the impact if we keep the requests configuration unchanged?

pintohutch:
@shuaich - it likely wouldn't affect too things too much.

This was initially a conservative effort to
- make expected idle resource usage explicit in our manifests
- ease scheduling concerns on resource-strained K8s nodes
 
but practically speaking, it probably doesn't affect too many use cases.

cmd/operator/deploy/operator/10-collector.yaml
```diff
@@ -115,10 +115,10 @@ spec:
           value: "25"
         resources:
           limits:
-            memory: 3G
+            memory: 2G
```
bwplotka, line 118: It would be quite amazing to depend that to machine type.. do we have plans for it? (TODO/ticket is fine)

bwplotka (approved):
LGTM, modulo one question - not a regression though.

cmd/operator/deploy/operator/10-collector.yaml
TheSpiritXIII: @bwplotka can you elaborate on how memory usage requirements might vary across different machine types?

shuaich (approved):
Thanks for these efforts.

cmd/operator/deploy/operator/10-collector.yaml
pintohutch: I'm honestly not sure how you would do this dynamically other than maybe using a [VPA](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler)?

TheSpiritXIII (approved):

PR #386, opened 2022-12-01
[DO NOT MERGE] Run e2e tests against a deployed operator

TheSpiritXIII:
This change makes a few fixes to allow running E2E tests against a live/deployed operator, and this capability is added through the `--live-operator` flag.

Notable changes include:
1. The operator typically runs on `gmp-system`. Users don't typically deploy resources on the operator namespace, so the tests were updated to put custom resources in a custom namespace and access the operator resources from the operator namespace.
2. The operator currently has 6 webhooks but the tests create 2. More were dynamically added so they match up when run on a deployed operator.
3. Minor typos and fixes.

PR #395, opened 2022-12-09
Target status e2e test

TheSpiritXIII:
This PR builds on top of #386

Due to the nature of `kind` tests running the operator within the test itself, target status cannot be done without port forwarding so for non-local operators, the tests were disabled.


TheSpiritXIII:
Closing. Superseded by https://github.com/GoogleCloudPlatform/prometheus-engine/pull/604.

PR #400, opened 2022-12-12
prepare v0.6.0 cut

pintohutch:
Bump the export library version in code as well as the anticipated manifest image versions.

### DIFF
--- a/cmd/operator/deploy/operator/05-deployment.yaml
+++ b/cmd/operator/deploy/operator/05-deployment.yaml
@@ -42,7 +42,7 @@ spec:
       priorityClassName: gmp-critical
       containers:
       - name: operator
-        image: gke.gcr.io/prometheus-engine/operator:v0.5.0-gke.0
+        image: gke.gcr.io/prometheus-engine/operator:v0.6.0-gke.0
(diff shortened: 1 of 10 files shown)


realschwa (approved):
Do we also need to update this in the following files?:

https://github.com/GoogleCloudPlatform/prometheus-engine/blob/0c2fcf209332a542e4af5326c413d2eb179f0984/examples/frontend.yaml#L46

https://github.com/GoogleCloudPlatform/prometheus-engine/blob/fbbff287ea59500bc9b8739a2fc8d027c2cae6ed/pkg/operator/e2e/alertmanager.yaml#L90

https://github.com/GoogleCloudPlatform/prometheus-engine/blob/bf78334acb31ac470cd08e482ebe8a44a7d0d270/pkg/operator/e2e/collector.yaml#L55

https://github.com/GoogleCloudPlatform/prometheus-engine/blob/bf78334acb31ac470cd08e482ebe8a44a7d0d270/pkg/operator/e2e/rule-evaluator.yaml#L56

pintohutch:
Oh those are symlinks - so should be good.

PR #408, opened 2023-01-05
examples: cert-manager podmonitoring manifest

ajayk:
Add an example for the basic cert-manager podmonitoring manifest to scrape the cert-manager metrics

### DIFF
--- a/examples/cert-manager/README.md
+++ b/examples/cert-manager/README.md
@@ -1,0 +1,1 @@
+# Cert-Manager sample manifests
--- a/examples/cert-manager/pod-monitoring.yaml
+++ b/examples/cert-manager/pod-monitoring.yaml
@@ -1,0 +1,28 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


ajayk:
@pintohutch : ^^

yqlu:
Hi @ajayk 

Thanks for the contribution!

Let's flesh out the README a bit more -- the other examples here link to documentation in which we try to guide the user through the whole process of integrating an application with Managed Service with Prometheus. The podmonitoring config is only one part of the puzzle. Can you include:

1. a link to cert-manager itself (https://cert-manager.io/)
2. a sample yaml snippet or instructions on how to configure cert-manager to export prometheus-format metrics. I haven't worked with cert-manager, but I found the following through a quick search:
    - https://cert-manager.io/docs/usage/prometheus-metrics/ , which says you want to set certain configs in the Helm chart or apply a patch to the container
    - the helm chart itself (https://github.com/cert-manager/cert-manager/blob/7c175c5d80fba53915ae6c98c5c0e7cde8f27615/deploy/charts/cert-manager/templates/deployment.yaml#L117) seems to suggest that the named port `http-metrics` (9402) is always exposed no matter what. Setting additional configs only matter for interoperability with prometheus-operator, not GMP with managed collection
3. guidance on how to verify (2) is working -- for example, our [Airflow](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/airflow#prerequisites) instructions document how to run `kubectl port-forward ...` and access the port to check that metrics are available at that endpoint.
4. podmonitoring -- you have this, great!
5. guidance on how to verify that metrics are correctly ingested into GMP (again, refer to our [published docs](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/airflow#verify-config) that ask the user to navigate to Metrics Explorer and run a given query.

PR #411, opened 2023-01-20
Update couchdb sample snippet

yqlu:
Eliminate stray indentation


lyanco (approved):
extremely controversial but i'll allow it

PR #413, opened 2023-01-23
Fix deprecated, missing error and type cast lint checks

TheSpiritXIII:
This fixes all lints warnings in the operator package and a few in other packages.

### DIFF
--- a/pkg/operator/collection.go
+++ b/pkg/operator/collection.go
@@ -371,12 +371,6 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 		cfgs, err := pmon.ScrapeConfigs(projectID, location, cluster)
 		if err != nil {
 			msg := "generating scrape config failed for PodMonitoring endpoint"
-			cond = &monitoringv1.MonitoringCondition{
-				Type:    monitoringv1.ConfigurationCreateSuccess,
-				Status:  corev1.ConditionFalse,
-				Reason:  "ScrapeConfigError",
-				Message: msg,
-			}
 			logger.Error(err, msg, "namespace", pmon.Namespace, "name", pmon.Name)
 			continue
 		}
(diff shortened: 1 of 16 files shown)


pkg/operator/collection.go
```diff
@@ -371,12 +371,6 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 		cfgs, err := pmon.ScrapeConfigs(projectID, location, cluster)
 		if err != nil {
 			msg := "generating scrape config failed for PodMonitoring endpoint"
-			cond = &monitoringv1.MonitoringCondition{
```
pintohutch, line 374: I actually think this is a bug. We never update the status on a bad config.

Can you add a TODO to fix with your deletion?
```go
// TODO: surface ScrapeConfigError to PodMonitoring status here.
```

pkg/operator/collection.go
```diff
@@ -410,12 +404,6 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 		cfgs, err := cmon.ScrapeConfigs(projectID, location, cluster)
 		if err != nil {
 			msg := "generating scrape config failed for PodMonitoring endpoint"
-			cond = &monitoringv1.MonitoringCondition{
```
pintohutch, line 413: Same with this, this should be surfaced to users instead of nacked.

pkg/operator/e2e/context.go
```diff
@@ -256,23 +262,20 @@ func (tctx *testContext) createBaseResources(ctx context.Context) ([]metav1.Owne
 		if err != nil {
 			return nil, errors.Wrap(err, "deserializing alertmanager manifest")
 		}
-		switch obj.(type) {
+		switch obj := obj.(type) {
```
pintohutch, line 265: nice.

pkg/operator/target_status_test.go
```diff
@@ -1251,9 +1251,11 @@ func TestPolling(t *testing.T) {
 	expectStatus := func(t *testing.T, description string, expected []monitoringv1.ScrapeEndpointStatus) {
 		// Must poll because status is updated via other thread.
 		var err error
-		if pollErr := wait.Poll(100*time.Millisecond, 2*time.Second, func() (done bool, err error) {
+		if pollErr := wait.Poll(100*time.Millisecond, 2*time.Second, func() (bool, error) {
 			var podMonitorings monitoringv1.PodMonitoringList
-			kubeClient.List(ctx, &podMonitorings)
+			if err := kubeClient.List(ctx, &podMonitorings); err != nil {
+				return false, nil
```
pintohutch, line 1257: nit: By returning `nil` here, we're effectively nacking any errors when doing the list. Is that the desired behavior? Or should we error and exit immediately in the test?

pintohutch (commented):
Btw, we have a make target [`make lint`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/46daf9aee156643e76916973cf99fd5fc89828b6/Makefile#L40-L43) that runs the default golangci-lint checkers.

I don't think your changes here resolve 100% everything there, but if you want to add linting to the Github workflows, that could also be something we enforce continually.

PR #417, opened 2023-01-31
Add Kubernetes version label to Pods with images

TheSpiritXIII:


pintohutch (commented):
can you add to the 11-rule-evaluator.yaml as well and `make presubmit`?

pintohutch (approved):

PR #419, opened 2023-02-02
Update the RabbitMQ sample PodMonitoring manifest

yqlu:
There are two common ways to install RabbitMQ which result in pods with different label selectors. Acknowledge both options in the example.

This manifest is embedded in https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/rabbitmq#exporter-podmonitoring which also includes a paragraph reminder to check this against the labels on the RabbitMQ pods.


macxamin (approved):
Looks good.

PR #360, opened 2022-11-09
go-1.18:bump go to 1.18 and controller-gen v0.10.0

ajayk:
- Update to go 1.18
- Upgrade controller-gen to v0.10.0 , cleans up the crd status field deletion hacks

PR #424, opened 2023-02-16
Build with Golang 1.20

TheSpiritXIII:


pintohutch:
You'll want to bump the go version in all the binaries' `Dockerfile`s as well like in this [PR](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/360/files?file-filters%5B%5D=No+extension&show-deleted-files=true&show-viewed-files=true).

PR #429, opened 2023-02-22
allow frontend to find ProjectID from metadata

smaftoul:
Pretty simple, and it seems to me, most people will configure frontend for the current project.

Does it make sense ? Any tests to add ?
Also, it probably needs a bit of doc ?

### DIFF
--- a/cmd/frontend/main.go
+++ b/cmd/frontend/main.go
@@ -33,6 +33,7 @@ import (
 	"syscall"
 	"time"
 
+	"cloud.google.com/go/compute/metadata"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/oklog/run"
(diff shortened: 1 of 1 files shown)


lyanco:
I believe this is not the behavior we want, actually, and it could cause a lot of confusion.

We recommend that you have one frontend per metric scope, and we also recommend that a metric scope should be a project with no resources within it. So if someone follows all the recommendations, and they have a multi-project metric scope, then they actually should **always** have to specify a different ProjectID than the one the frontend is running in.

For customers who don't use scopes, choosing the current project ID should be fine, but I'd rather require people to read the documentation and make a decision instead of relying on built-in default behavior that could be wrong.

smaftoul:
@lyanco thanks for the prompt answer ;)

Actually, on the linked page I read this:

>Change the queried project

>The frontend deployment **uses the cluster's Google Cloud project as the scoping project**. If this project is the scoping project of a [multi-project metrics scope](https://cloud.google.com/monitoring/settings#concept-scope), then it can read metrics from all projects in the metrics scope.
>You **can** specify a **different** project by using the --query.project-id flag.
>Typically, you use a dedicated project as a scoping project, and this project is not the same project the frontend deployment runs in. To let the deployment read a different target project, you must do the following: [...]

If I understand correctly, my PR implements this behaviour, which is not the current case.

I am actually configuring a multi-project scope and will have to change the projectID I am using, but have already deployed the frontend and found boring to configure that, while usually, when I don't specify any project it default to the current project if applicable, in my experience.

We maybe can add an info log so it's clear when the service starts ?

lyanco:
Nice catch on the docs, thanks! I will update those.

FWIW one of the biggest failure points is people not configuring Grafana to connect to the APIs, instead leaving it connected to the local Prometheus. It's maybe like 1/8th of users miss this, so I do want to keep the friction here to help ensure people set Grafana up in a way that actually gets the full benefit of the system. Especially as you should only be setting up Grafana and the frontend only once or a handful of times, it hopefully shouldn't be too painful... but yes, boring :-)

lyanco:
The real solution here is for Grafana to implement https://github.com/grafana/grafana/discussions/40844#discussioncomment-4332200 but that's out of our control.

PR #431, opened 2023-02-24
Re-introduce example-app

pintohutch:
This change reverts 7b103d1f59d0819a3813e08f2b32776895266b49 and adds toy exemplar instrumentation to a histogram metric, that can be controlled through a sampling flag.

Kudos @realschwa for most of the changes.


macxamin (approved):
LGTM

PR #433, opened 2023-02-26
Add Anthos hybrid distributions to UserAgentMode

### LINKED ISSUE #432: Add anthos distributions to the `user-agent-mode` enum
right now the `user-agent-mode` has 3 supported values: `gke`, `kubectl`, `unspecified`, ref: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.6.1/pkg/export/setup/setup.go#L134

By adding anthos distributions to the enum (e.g. one additional enum item `anthos` or further breakdowns like `anthos-vmware`, `anthos-baremetal`, `anthos-aws`, `anthos-azure`, `anthos-attached`), we'll be able to further distinguish the traffic coming from different type of clusters.

leonzz:
This resolves #432

### DIFF
--- a/pkg/export/setup/setup.go
+++ b/pkg/export/setup/setup.go
@@ -48,6 +48,8 @@ const (
 	UAModeGKE         = "gke"
 	UAModeKubectl     = "kubectl"
 	UAModeUnspecified = "unspecified"
+	UAModeAVMW        = "on-prem"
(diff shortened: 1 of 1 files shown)


bwplotka (approved):
Thanks, LGTM 👍🏽

PR #435, opened 2023-02-27
Added examples/go for example Go instrumentation (metrics, logs & tracing) and sample GMP configuration.

bwplotka:


bwplotka:
This is a little bit overlapping with https://github.com/GoogleCloudPlatform/prometheus-engine/tree/main/cmd/example-app - perhaps it's wise to merge those in this PR.

I would propose putting https://github.com/GoogleCloudPlatform/prometheus-engine/tree/main/cmd/example-app inside `examples/go` dir, WDYT? Ideally we also decide which one we want:

a. client/server with middleware setting exemplars
b. mem util based with direct exemplars - or both? 🤔 

cc @realschwa @pintohutch

pintohutch:
Honestly, we could just supersede the current example app with this one.

We want a simple app that:
- exposes Prometheus metrics and exemplars
- has an accompanying pod monitoring
- we can use those signals ingested into GMP to run e2e tests against
- we can refer to in our documentation that can help users get started

The fact that this app is less "toyish" in how it's instrumented and provides boilerplate for OT traces and logs as well is great!

PR #345, opened 2022-10-14
Remove Autopilot hack

TheSpiritXIII:

PR #438, opened 2023-03-03
remove aggressive tolerations on non-DS workloads

pintohutch:


TheSpiritXIII (commented):
We can get rid of the hack on the DaemonSet: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/pintohutch/rm-tolerations/manifests/operator.yaml#L680

However, that is optional for you. We could leave it and remove it in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/345

TheSpiritXIII (approved):

PR #446, opened 2023-03-21
Security patches

TheSpiritXIII:
Includes security patches from the `main` branch.

Cherry-picked:
- https://github.com/GoogleCloudPlatform/prometheus-engine/commit/9930d0648735c2b3dc626dc29ba2e3ec9e6390aa
- https://github.com/GoogleCloudPlatform/prometheus-engine/commit/46b559329d53c98d50d861e538b70c1d5e1850a7
- https://github.com/GoogleCloudPlatform/prometheus-engine/commit/b7802536933248f9ed8246dc9d13ad546c97bace

And manually did partial https://github.com/GoogleCloudPlatform/prometheus-engine/pull/360 because Go 1.17 introduced [more control over the module graph](https://go.dev/doc/go1.17). However, we have to go to Go 1.18 and not 1.17 because 1.17 breaks `go mod tidy`, which is needed for `make regen`.


pintohutch:
Don't we just need https://github.com/GoogleCloudPlatform/prometheus-engine/commit/9930d0648735c2b3dc626dc29ba2e3ec9e6390aa cherry-picked on top of https://github.com/GoogleCloudPlatform/prometheus-engine/commit/14c288e605eebbc1179ffe97d77be1344cc7a1e8?

StevenYCChou:
> And manually did partial #360 because Go 1.17 introduced [more control over the module graph](https://go.dev/doc/go1.17). However, we have to go to Go 1.18 and not 1.17 because 1.17 breaks `go mod tidy`, which is needed for `make regen`.

Can you elaborate more on this comment as I don't understand what is broken if you do not do partial #360. It seems that we'll upgrade to G0 1.20, so I am curious what is introduced here for your partial #360 , so I can have a proper review on the last commit.

TheSpiritXIII:
> Can you elaborate more on this comment as I don't understand what is broken if you do not do partial #360. It seems that we'll upgrade to G0 1.20, so I am curious what is introduced here for your partial #360 , so I can have a proper review on the last commit.

We _build_ with Go 1.20 but the code is _written for_ Go 1.18. Previously, our code was written for Go 1.16 but this lacks the ability to specify transitive dependency versions. Go 1.17 introduced that ability (hence the new section in `go.mod` on that commit). However, Go 1.17 introduced a special flag when you call `go mod tidy` and you either have to specify to use the new `go.mod` format or old `go.mod` format. In Go 1.18 they removed this special flag. This parallels our `main` branch setup.

PR #449, opened 2023-03-23
Add option to default the project_id label for exemplars

realschwa:


pkg/export/export.go
```diff
@@ -170,6 +170,12 @@ type ExporterOpts struct {
 	Location  string
 	Cluster   string
 
+	// If true, automatically populate the project id in an exemplar labelset.
+	PopulateExemplarProjectID bool
```
bwplotka, line 174: I would say let's avoid bools flag. They are not flexible and scalable. What about empty `ExemplarProjectID` stopping this to populate that extra label?

pkg/export/setup/setup.go
```diff
@@ -133,6 +133,12 @@ func FromFlags(a *kingpin.Application, userAgentProduct string) func(log.Logger,
 	a.Flag("export.label.project-id", fmt.Sprintf("Default project ID set for all exported data. Prefer setting the external label %q in the Prometheus configuration if not using the auto-discovered default.", export.KeyProjectID)).
 		Default(opts.ProjectID).StringVar(&opts.ProjectID)
 
+	a.Flag("export.exemplars.populate-exemplar-project-id", "If true, automatically populate the 'project_id' label in an exemplar labelset.").
```
bwplotka, line 136: ditto

pkg/export/setup/setup.go
```diff
@@ -133,6 +133,12 @@ func FromFlags(a *kingpin.Application, userAgentProduct string) func(log.Logger,
...
+		Default("true").BoolVar(&opts.PopulateExemplarProjectID)
+
+	a.Flag("export.exemplars.project-id", "If non-empty and export.exemplars.populate-exemplar-project-id is true, use this project_id as the project_id for exemplar labelsets.").
```
bwplotka, line 139: let's also explain the dependency on existing project_id label AND trace and span existance?

pkg/export/export.go
realschwa: I think the semantics of this would get confusing since this is supposed to be both overridable and disableable. 

An empty `ExemplarProjectID` seems to make sense as the case where you would not autopopulate the `project_id`, but not passing something in seems to imply that it is the default state which is not what we want. We want the default state to be to autopopulate the exemplar projectID. 

Unless you mean empty as in the literal empty string (`""`) which also seems to be semantically confusing.

I'll defer to you, @pintohutch and @lyanco  on this one.

PR #453, opened 2023-03-27
Add example setups for argo, etcd, hadoop, hbase, jetty, mongo, scyll…

yqlu:
…a, solr, tomcat, varnish

### DIFF
--- a/examples/mongodb/exporter.yaml
+++ b/examples/mongodb/exporter.yaml
@@ -1,0 +1,39 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mongodb-exporter
+  labels:
+    app.kubernetes.io/name: mongodb-exporter
+spec:
+  replicas: 1
+  selector:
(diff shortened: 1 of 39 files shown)


mtabasko (approved):
LGTM

examples/mongodb/exporter.yaml
```diff
@@ -0,0 +1,39 @@
...
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: mongodb-exporter
+    spec:
+      containers:
+      - name: exporter
+        image: percona/mongodb_exporter:0.20
+        args:
+        - --mongodb.uri=mongodb://user:password@mongodb-service-name:27017
```
lyanco, line 35: Doesn't this line need a +? user:password should need to be changed

lyanco (changes requested):

examples/mongodb/exporter.yaml
mtabasko: Probably -- which means we'll have to change this to a "snippet" file here and in the import. (No biggie, just noting it).

yqlu: The rule is slightly different: we've been using `+` to mark when you need to *add* lines to your *existing* application yaml to add a sidecar or turn on exporting by setting env variables.

There are two prior cases where we've instructed you to install a brand new standalone exporter -- we haven't been using `+`, but instead spell out in a paragraph following the yaml how the exporter should talk to the application.

[Elasticsearch](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/elasticsearch)

> The Elasticsearch exporter is configurable with [flags](https://github.com/prometheus-community/elasticsearch_exporter#configuration), which can be set as container args. You can adjust your exporter's configuration to suit your Elasticsearch configuration. Update the `--es.uri` flag to match your environment's username, password, and clusterIP service name.

[Kafka](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/kafka)

> The Kafka exporter is configurable with [flags](https://github.com/danielqsj/kafka_exporter#flags), which can be set as container args. This example assumes Kafka is available as a ClusterIP service named `kafka` on port `9092`.

lyanco (approved):

PR #454, opened 2023-03-28
Add support for insecure TLS scraping

### LINKED ISSUE #238: support TLS config: insecure_skip_verify for PodMonitoring config
with promethus config:

```
    - job_name: 'vault-pods'
      metrics_path: "/v1/sys/metrics"
      params:
        format: ['prometheus']
      scheme: https
      tls_config:
        insecure_skip_verify: true
```

Did not find any way to set the `insecure_skip_verify: true` from the [scrapeendpoint doc]( https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.4.1/doc/api.md#scrapeendpoint)

Error message:
```
2022-06-01T17:27:06.271Z [INFO]  http: TLS handshake error from 10.1.2.125:50352: remote error: tls: bad certificate
```

ip: 10.1.2.125,   is the pod ip of the collector.

```
❯ kubectl get po -n gmp-system
NAME                              READY   STATUS    RESTARTS      AGE
collector-jsjx9                   2/2     Running   0             25h
```

TheSpiritXIII:
Fixes #238! This PR also updates the example-app to support starting up with a TLS configuration, so I was able to test this new flag.

This PR builds on top of #570.

Internal doc: go/managed-prometheus:minimal-scrape-endpoint-authorization


cmd/example-app/main.go
```diff
@@ -187,10 +221,79 @@ func main() {
 		)
 	}
 	{
-		server := &http.Server{Addr: *addr}
+		var tlsConfig *tls.Config = nil
+		if flag.Lookup("self-signed") != nil || *tlsCert != "" {
```
bwplotka, line 225:
```suggestion
		if selfSigned != nil || *tlsCert != "" {
```

Why not variable?

bwplotka (commented):
Looks good so far. I assume we want to finalize and approve design first, right?

TheSpiritXIII:
@bwplotka the only detail we're finalizing with this PR is the fact that the authentication configuration will be inline, which I think we've already settled on. Let me know if you disagree or want us to discuss some more!

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -997,6 +1012,14 @@ type ScrapeEndpoint struct {
 	// instance, or __address__) are not permitted. The labelmap action is not permitted
 	// in general.
 	MetricRelabeling []RelabelingRule `json:"metricRelabeling,omitempty"`
+	// Prometheus HTTP client configuration.
+	HTTPClientConfig `json:",inline"`
+}
+
+// HTTPClientConfig stores HTTP-client configurations.
+type HTTPClientConfig struct {
```
bwplotka, line 1020: I think this might be different to what was in the design doc so far (https://docs.google.com/document/d/16erW70zg-F8CE0Kf3GHMASRsdNnMUNfdCI5YvfnNXxw/edit?resourcekey=0-N8ZdIwLMf7uNdm1XmckMaQ#heading=h.sfylv0d4e4lc). Let's make it consistent as it is the change we have to be careful here (:

bwplotka (commented):
Nice, we synced offline, we want to unblock this issue with simple PR that should go along early design doc ideas we had - however I wanted to double check the API we are forming here. In the doc we had:

```
Authorization:
  # Sets the `Authorization` header on ever scape reques wth the
  # configured username and password.
  BasicAuth: BasicAuth

  # Sets the `Authorization` header on every scrape request with
  # the configured credentials.
  Authorization: Authorization

  # Optional OAuth 2.0 configuration.
  OAuth2: OAuth2

  # Optional TLS configuration.
  TLS: TLS
```

pkg/operator/apis/monitoring/v1/types.go
bwplotka: Maybe I took the design outline too literally?

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -695,6 +695,21 @@ func endpointScrapeConfig(id, projectID, location, cluster string, ep ScrapeEndp
 		httpCfg.ProxyURL.URL = proxyURL
 	}
 
+	if ep.HTTPClientConfig.TLS != nil {
+		unsupportedFormat := "Setting %q is not supported. See https://github.com/GoogleCloudPlatform/prometheus-engine/issues/450 for more details"
+		if ep.HTTPClientConfig.TLS.CA != nil {
```
bwplotka, line 700: This is odd a little bitt. We are adding fields we want in future, sure, but we produce documentation and ability to parse this only to tell user that it's not possible to use (since propagation mechanism has to be designed and implemented) - and it's not even documented anywhere, just in error message.

I understand the reusability of it, but wonder if this will not confuse users and upset. Perhaps we could create temporary type with only skip field? (should be easy to replace with TLSConfig later on?)

bwplotka (commented):
I am fine with type flattened now, but I want to double check another thing - risk of user confusion here, see comment below - WDYT?

pkg/operator/apis/monitoring/v1/types.go
pintohutch: hrm - yea tricky one. My hesitancy with a temporary type is we'll have to handle [CRD versioning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#overview), which isn't the most fun thing.

You could also create a duplicate type that we slowly add to over time so it's backwards compatible. That seems a little messy, but does avoid dealing with versions/conversions.

At the very least, you could add a comment about lack of support [here](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/454/files#diff-918b2f2eda7a36f6f638775958962c3aeeaa1272fa163f6f4a3633da478f667fR1015), which should make it's way to the API docs with `make presubmit`.

TheSpiritXIII: Going with the duplicate type idea, since when we generate the CRD metadata, it does not show up as a separate type because it's an inline field. Thanks!

pintohutch:
I'm going to wait to review until we have consensus on the design. Please ping again if you're ready for a round!

cmd/example-app/main.go
TheSpiritXIII: Marking resolved. This was split out into #570. Thanks!

bwplotka (approved):
Sorry for lag! Let's go 💪🏽

PR #458, opened 2023-04-14
Use static target for managed Alertmanager

pintohutch:
Since the managed Alertmanager service is known, we can rely on [K8s DNS resolution](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services) of services to configure a static target to route alerts to.

This is an expected improvement over relying on K8s SD as it reduces the watch load in OOTB deployments.

### DIFF
--- a/pkg/operator/e2e/main_test.go
+++ b/pkg/operator/e2e/main_test.go
@@ -316,6 +316,14 @@ func testRuleEvaluatorConfig(ctx context.Context, t *testContext) {
         external_key: external_val
 alerting:
     alertmanagers:
+        - follow_redirects: true
+          enable_http2: true
+          scheme: http
+          timeout: 10s
+          api_version: v2
+          static_configs:
+            - targets:
+                - alertmanager.{namespace}.svc.cluster.local:9093
         - authorization:
             type: Bearer
             credentials_file: /etc/secrets/secret_{pubNamespace}_alertmanager-authorization_token
@@ -337,29 +345,6 @@ alerting:
               regex: (.+):\d+
               target_label: __address__
               replacement: $1:19093
-              action: replace
-          kubernetes_sd_configs:
-            - role: endpoints
-              kubeconfig_file: ""
-              follow_redirects: true
-              enable_http2: true
-              namespaces:
-                own_namespace: false
-                names:
-                    - {namespace}
-        - follow_redirects: true
-          enable_http2: true
-          scheme: http
-          timeout: 10s
-          api_version: v2
-          relabel_configs:
-            - source_labels: [__meta_kubernetes_endpoints_name]
-              regex: alertmanager
-              action: keep
-            - source_labels: [__address__]
-              regex: (.+):\d+
-              target_label: __address__
-              replacement: $1:9093
               action: replace
           kubernetes_sd_configs:
             - role: endpoints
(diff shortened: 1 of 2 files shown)


pkg/operator/operator_config.go
```diff
@@ -465,18 +464,20 @@ func (r *operatorConfigReconciler) makeAlertmanagerConfigs(ctx context.Context,
 		if ports := amSvc.Spec.Ports; len(ports) > 0 {
 			// Assume first port on service is the correct endpoint.
 			port := ports[0].Port
-			alertManagers = append(alertManagers, monitoringv1.AlertmanagerEndpoints{
-				Name:      amNamespacedName.Name,
-				Namespace: amNamespacedName.Namespace,
-				Port: intstr.IntOrString{
-					Type:   intstr.Int,
-					IntVal: port,
+			svcDNSName := fmt.Sprintf("%s.%s.svc.cluster.local:%d", amSvc.Name, amSvc.Namespace, port)
```
StevenYCChou, line 467: It seems that `cluster.local` is adjustable by adjusting cluster-domain mentioned in [1]. It seems that GKE allows adjusting the cluster domain by customizing DNS [2].

Suggest using "%s.%s:%d" as it seems that it will resolve to the same address as long as a DNS service is set up [3]. If a DNS is not set up, I would assume "%s.%s.svc.cluster.local:%d" won't be usable either - hence the suggestion.

[1] https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#introduction
[2] https://cloud.google.com/kubernetes-engine/docs/how-to/custom-kube-dns
[3] https://kubernetes.io/docs/concepts/services-networking/service/#dns

StevenYCChou (changes requested):

pkg/operator/operator_config.go
pintohutch: Fantastic point. Thanks for catching this! I made the suggested change.

pkg/operator/e2e/main_test.go
```diff
@@ -347,29 +355,6 @@ alerting:
                 own_namespace: false
                 names:
                     - {namespace}
```
StevenYCChou, line 357: I couldn't comment here but I want to know if the `kubernetes_sd_configs` here is constructed by default - or if it's  something added by the users additionally.

pintohutch: It's in the test because we configure an [extra alertmanager](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/7106c5142e22edb3ff337a98fdfb525d1609e258/pkg/operator/e2e/main_test.go#L209) in the test harness.

I'm leaving non-managed PAMs on k8s SD. I suppose we could change that in the future as well, but probably not necessary right now.

StevenYCChou: Agreed - I also read up [1] and [2] and understand now it's for configuring additional alertmanager endpoints. Agree that we can leave them out of this PR and decide later as it can be a separate decision.

Thanks for the clarification and working it through with me

[1] https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/doc/api.md#alertmanagerendpoints
[2] https://cloud.google.com/stackdriver/docs/managed-prometheus/rules-managed#self-deployed_alertmanager

StevenYCChou (approved):

PR #461, opened 2023-04-18
Disable target status reporting by default

pintohutch:
Introduce a field in the OperatorConfig to toggle target status reporting as an opt-in feature.

The motivation for this is to minimize cpu and memory usage of the operator in cases where people don't care to use the feature.


TheSpiritXIII (approved):
Look good!

lyanco:
lgtm

macxamin (approved):
LGTM, should we remove this flag https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/operator/operator.go#L122

I don't see it being set and it makes the code confusing.

pintohutch:
> 

Good call! Done

PR #462, opened 2023-04-18
use higher limits for operator and rule-evaluator

pintohutch:
While extremely rare, we have seen cases where memory of these components can bleed up to the limits specified here.

### DIFF
--- a/cmd/operator/deploy/operator/05-deployment.yaml
+++ b/cmd/operator/deploy/operator/05-deployment.yaml
@@ -59,7 +59,7 @@ spec:
           containerPort: 18080
         resources:
           limits:
-            memory: 512M
+            memory: 2G
           requests:
             cpu: 8m
             memory: 32M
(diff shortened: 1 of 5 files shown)


shuaich (approved):

shuaich:
Should we also bump the memory limit for operator here in https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/manifests/operator.yaml#L264 ?

pintohutch:
> Should we also bump the memory limit for operator here in https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/manifests/operator.yaml#L264 ?

I believe that's [done](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/462/files#diff-961a34fc7262fda9004a88b40f9b7ad3f0974a6b6a37ab5e173603057dec27bfR264), no?

macxamin (approved):

shuaich:
Yes. Thanks for clarifying this.

PR #464, opened 2023-04-21
Adjust rule-evaluator memory limits

macxamin:
Increase evaluator memory limit to 1G
reset config-reloader to 32M from 1G


pintohutch (approved):
LGTM!

PR #465, opened 2023-04-21
Add Anthos MultiCloud distributions to UserAgentMode

### LINKED ISSUE #432: Add anthos distributions to the `user-agent-mode` enum

leonzz:
#433 only added the 2 Anthos Hybrid distributions, this one adds 3 MultiCloud distributions to fully resolve #432.

### DIFF
--- a/pkg/export/setup/setup.go
+++ b/pkg/export/setup/setup.go
@@ -50,6 +50,9 @@ const (
 	UAModeUnspecified = "unspecified"
 	UAModeAVMW        = "on-prem"
 	UAModeABM         = "baremetal"
+	UAModeAWS         = "aws"
+	UAModeAzure       = "azure"
+	UAModeAttached    = "attached"
 )
 
 // Environment variable that contains additional command line arguments.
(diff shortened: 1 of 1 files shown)


pkg/export/setup/setup.go
```diff
@@ -50,6 +50,9 @@ const (
 	UAModeUnspecified = "unspecified"
 	UAModeAVMW        = "on-prem"
 	UAModeABM         = "baremetal"
+	UAModeAWS         = "aws"
```
pintohutch, line 53: We have both a `UserAgentMode` and `UserAgentEnv`, where the former is generally reserved for the mode of installation and the latter for the platform environment. Notably, the former is user-modifiable via a flag, whereas the latter is hard-coded into the binary and auto-populated at runtime.

Is there any way to auto-detect whether we're running in AWS or Azure using a go library that we could add to the logic in [`setup.go`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/4e7649c1e50dd7ea326d2760b954258a41f53349/pkg/export/setup/setup.go#L93-L118)?

For on-prem or baremetal, autodetection is not possible due to the heterogeneity of those environments, so using "mode" as a way to distinguish these makes sense, albeit it's getting double-usage out of that flag.

pkg/export/setup/setup.go
```diff
@@ -50,6 +50,9 @@ const (
 	UAModeUnspecified = "unspecified"
 	UAModeAVMW        = "on-prem"
```
pintohutch, line 51: Could you add comments on what each of these modes means? e.g. what's the difference between "on-prem" and "baremetal"?

pkg/export/setup/setup.go
```diff
@@ -50,6 +50,9 @@ const (
...
 	UAModeABM         = "baremetal"
+	UAModeAWS         = "aws"
+	UAModeAzure       = "azure"
+	UAModeAttached    = "attached"
```
pintohutch, line 55: What is "attached"?

pkg/export/setup/setup.go
leonzz: That's a good point. @ericroman920 could you comment on the auto detection for anthos on aws & azure?

pkg/export/setup/setup.go
leonzz: [Attached clusters](https://cloud.google.com/anthos/clusters/docs/multi-cloud/attached) could be any k8s cluster that customers want to connect to Google.

pintohutch: Gotcha - could you add a comment?

PR #466, opened 2023-04-21
add self-monitoring for mgd rule-evaluator

pintohutch:


lyanco (approved):
We should add instructions to change gmp-system to gmp-gke-system in AP, but that's probably a docs change, not one relevant here

yqlu (approved):

PR #469, opened 2023-05-08
Add a tsdb-removing InitContainer.

StevenYCChou:


bwplotka (approved):
This looks fair to me.

Can we add clear commentary why it's safe and why we need this?

(consistency - the best effort ephemeral storage is not something we want to rely on, so we expect no persistence. If we want persistence - we need something more specialised and documented)

bwplotka (changes requested):
Actually let's be more careful and double check the code. It seems if the WAL is not used, it should be safe, but would need to verify this.

pintohutch:
Here's a [diff](https://github.com/GoogleCloudPlatform/prometheus/pull/52/commits/2b83f193654bddd18a54bdd840bb952331e14e74#diff-c3ae701c35bb1de34870207029d1fb3390613fbba65e6f2f9bffe059bff89001R495) with a good summary of go and non-vendor changes the GMP fork incorporates.

You can see the exporting of data (and thus the feeding to the GMP-specific "WAL") is done upon `Commit()` to the `headAppender`.

IIRC the WAL doesn't persist series or type metadata, so we can't reconstruct GCM TimeSeries samples from it anyway, so this should be safe.

pintohutch:
@StevenYCChou the presubmit is failing because the change should go in cmd/operator/deploy/operator/10-collector.yaml and then running `make presubmit` locally will mirror that to the global manifests dir (which is autogenerated generally).

pintohutch:
I meant the OSS Prometheus WAL.

Interesting the spec has it! I believe we assumed this wasn't the case when setting up the GMP WAL.

We currently [inject](https://github.com/GoogleCloudPlatform/prometheus/pull/52/commits/2b83f193654bddd18a54bdd840bb952331e14e74#diff-4535d6172e0a9ab4e0879d71ef67ab69023f6eda9b44100343a6afa04fc8925fR303) the metadata from the scrape loop context and [pass](https://github.com/GoogleCloudPlatform/prometheus/pull/52/commits/2b83f193654bddd18a54bdd840bb952331e14e74#diff-c3ae701c35bb1de34870207029d1fb3390613fbba65e6f2f9bffe059bff89001R495) that along to the GMP WAL during append commit.

Quickly source-diving, I don't [see](https://github.com/GoogleCloudPlatform/prometheus/blob/release-2.35.0-gmp/tsdb/record/record.go#L230-L245) type information persisted to the WAL on write, but I may be misunderstanding the code here.

pintohutch:
friendly ping @bwplotka @StevenYCChou - any other hesitations here?

If nothing else, we can verify this manually to make sure by deploying this alongside a non-modified collector and having them scrape and export the same data (we can inject an extra label or something to differentiate)

bwplotka:
I have now some time to verify this, if @StevenYCChou is ok with that.

bwplotka:
I looked together on our export and I think it's totally fine to remove WAL for our export cases. Export literally hooks into append: https://github.com/GoogleCloudPlatform/prometheus/blob/81bf2e3878b2e4be5b56e75d06ed0a95ee842b3d/tsdb/head_append.go#L972 and I don't think it saves anything.

However I wonder if it makes sense to just add such removal of WAL to our fork instead of scripts like this. However, the advantage with such script is that it will work once we will move someday to native Prometheus. But then we might want to reconsider persistence (:  Generally, some proposal would be nice where we want to go from here (persistence/no persistence, local querying disabled/enabled). 

For now it's kind of sad we block users and pay for WAL and TSDB when we don't use it (all is batched to export logic IIUC). However paying for it now, so we can also pay for it in future (once we would want persistence) is also not a bad idea.

StevenYCChou:
> I looked together on our export and I think it's totally fine to remove WAL for our export cases. Export literally hooks into append: https://github.com/GoogleCloudPlatform/prometheus/blob/81bf2e3878b2e4be5b56e75d06ed0a95ee842b3d/tsdb/head_append.go#L972 and I don't think it saves anything.

Got it - feel free to merge this PR if the change makes sense. If there are any more changes needed, feel free to create a new PR to add other changes so I am not blocking anything :)

> However I wonder if it makes sense to just add such removal of WAL to our fork instead of scripts like this. However, the advantage with such script is that it will work once we will move someday to native Prometheus. But then we might want to reconsider persistence (: Generally, some proposal would be nice where we want to go from here (persistence/no persistence, local querying disabled/enabled).
> 
> For now it's kind of sad we block users and pay for WAL and TSDB when we don't use it (all is batched to export logic IIUC). However paying for it now, so we can also pay for it in future (once we would want persistence) is also not a bad idea.

bernot-dev:
I would consider using a [PostStart hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) instead of an initContainer. `rm` is available in the existing container, so the hook may be a little less overhead than spinning up a new container.

bwplotka:
> If there are any more changes needed, feel free to create a new PR to add other changes so I am not blocking anything :)

Perfect, thanks for starting this @StevenYCChou  💪🏽 Closing this for now, I think we need to adjust - I like the PostStart hook, but honestly I would double check the fork change to make things less hacky.

Do you want to take a try in another PR @bernot-dev ?

PR #474, opened 2023-05-12
tolerate ready-url errors and retry

### LINKED ISSUE #472: config-reloader Restarts Multiple Times on Boot
I've been investigating why the `config-reloader` pod occasionally restarts multiple times while booting, and I've narrowed it down to [this line](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/3d1c0fbf4211bd34b955079297bece2e738da3c0/cmd/config-reloader/main.go#L91). I'm seeing the following in the logs:

```
Get \"http://localhost:19090/-/ready\": dial tcp 127.0.0.1:19090: connect: connection refused
```

Per the comment on line 73, it seems like the intent would be to continue polling in this situation, but instead the process exits and has to be restarted.

pintohutch:
Tolerate http request errors when calling the ready url.

During start-up, Prometheus could refuse connections, resulting in the returned error here. This should be tolerated and retried to mitigate container restarts.

Fixes #472

### DIFF
--- a/cmd/config-reloader/main.go
+++ b/cmd/config-reloader/main.go
@@ -87,9 +87,10 @@ func main() {
 				os.Exit(0)
 			case <-ticker.C:
 				resp, err := http.DefaultClient.Do(req)
+				// Log but tolerate the error here as a "connection refused"
+				// error from Prometheus is a retryable event during startup.
 				if err != nil {
 					level.Error(logger).Log("msg", "polling ready-url", "err", err)
-					os.Exit(1)
 				}
 				if resp.StatusCode == http.StatusOK {
 					level.Info(logger).Log("msg", "ready-url is healthy")
--- a/cmd/config-reloader/main.go
+++ b/cmd/config-reloader/main.go
@@ -90,7 +90,7 @@ func main() {
 				// Log but tolerate the error here as a "connection refused"
 				// error from Prometheus is a retryable event during startup.
 				if err != nil {
-					level.Error(logger).Log("msg", "polling ready-url", "err", err)
+					level.Error(logger).Log("msg", "polling ready-url", "status", resp.StatusCode, "err", err)
 				}
 				if resp.StatusCode == http.StatusOK {
 					level.Info(logger).Log("msg", "ready-url is healthy")


cmd/config-reloader/main.go
```diff
@@ -87,9 +87,10 @@ func main() {
 				os.Exit(0)
 			case <-ticker.C:
 				resp, err := http.DefaultClient.Do(req)
+				// Log but tolerate the error here as a "connection refused"
+				// error from Prometheus is a retryable event during startup.
 				if err != nil {
 					level.Error(logger).Log("msg", "polling ready-url", "err", err)
-					os.Exit(1)
```
TheSpiritXIII, line 92: @pintohutch do we maybe want to log non-OK statuses in the same spirit as we log here? In that case, would we maybe want to `break` here?

As an aside, I'm not sure why we're using a goroutine here -- we could have just `break`ed the infinite `for` loop within the `select`.

bwplotka (changes requested):
Thanks, but I think we need to be careful here.

Current SLIs are checking if pod is available (running and exposing "some" metrics) and if the `reloader_reloads_failed_total / reloader_reloads_total < X`. If we e.g misconfigure prometheus URL we would essentially never catch this in SLO, yet configuration would be never reconciled.

We are resisting configuration SLOs, so let's discuss there what behavior & SLO updates we want.

pintohutch:
IIUC the config-reloader metrics won't be emitting anything at this point in the code because it takes place before the reloader is initiated and started.

However, you do make a good point around misconfiguration of the `ready-url` and not being able to catch issues there.

What do you think about creating and incrementing a metric here? Strawman: `reloader_ready_url_retries_total{status="success|fail"}`?

bwplotka:
Nice, discussing this with @macxamin

pintohutch:
Friendly ping @macxamin @bwplotka - maybe this will help reducing noise in config-reloader SLOs?

bwplotka:
Aware of that, on it.

bwplotka:
Proposed different solution: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/526 - closing this for now. We can always reopen.

pintohutch:
@bwplotka WDYT about re-opening this considering we're going to be reliant on the config-reloader for the time being?

We're still seeing transient restart spikes, and I think this may help.

bwplotka (changes requested):
Thanks! I personally would vote to NOT do it, but if the team wants to go with its consequences I am happy to disagree and commit. My preference is [removing readiness logic and fixing our alerts and SLO for failed reloads metric
](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/526)

### My rationales

If we want this we have to accept that any misconfiguration to **ready-URL** (or readiness handling on target components) will cause silent errors (error log only) not captured by anything (alert, SLO, user facing error). Even further, it will block reloader from actually generating right configuration e.g. on Prometheus startup (we decompress and re-generate main configuration file in target path that Prometheus waits to read after init container created empty file).

Is that better then crashlooping config-reloader on crash looping Prometheus? On scale, I would argue not being able to understand when we misconfigure reloader can have bigger consequences.

### Alternatives

One alternative is to increment metric, but that would have similar issue as our flakiness of failed reloads metric. We could finally update our alerts to exclude cases of Prometheus (or any other target component) unready/down, but then removing readiness check https://github.com/GoogleCloudPlatform/prometheus-engine/pull/526 and relying only on failed reloads is simpler and has same effect IMO.

Second alternative is to alert on lack of reloads. This also requires some exclusion of Prometheus down states. This has also problem of relying on reloads happening periodically or at least once on start.

Third alternative is to only sort crashes on start up (e.g Prometheus starting too slow vs config-reloader. The same PR but with some 5m timeout would be safe to accept IMO. Won't solve crashlopping Prometheus case, but it would potentially remove the startup phase flakiness.

pintohutch:
Apologies for the delayed response here @bwplotka. I agree that this may be hurting more than helping. I'm going to close this and we can iterate on a longer term solution, starting with #526.

PR #475, opened 2023-05-15
bump to using go 1.20.4

pintohutch:


TheSpiritXIII:
Do we want this? Simply rebuilding the images today would rebuild as 1.20.4. Tomorrow, if 1.20.5 comes out, we can rebuild again to get 1.20.5, but after this change, we'd have to make another code change to get 1.20.5.

Issue #479, opened 2023-05-19
[bug] Unable to specify endpoints with the same port name


Zebradil:
I convert a ServiceMonitor resource rendered from a helm chart into a PodMonitoring resource according to the [GMP documentation](https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-managed#gmp-servicemonitor).

<details><summary>The initial ServceMonitor resource</summary>
<p>

```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  # Removed
  #annotations:
  labels:
    app.kubernetes.io/component: keycloak
    app.kubernetes.io/instance: keycloak
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: keycloak
    helm.sh/chart: keycloak-15.1.1
  name: keycloak
  namespace: keycloak
spec:
  endpoints:
    - interval: 30s
      path: /metrics
      port: http
    - interval: 30s
      path: /realms/master/metrics
      port: http
  namespaceSelector:
    matchNames:
      - keycloak
  selector:
    matchLabels:
      app.kubernetes.io/component: metrics
      app.kubernetes.io/instance: keycloak
      app.kubernetes.io/name: keycloak
```

</p>
</details> 

<details><summary>The resulting PodMonitoring resource</summary>
<p>

```yaml
apiVersion: monitoring.googleapis.com/v1
kind: PodMonitoring
metadata:
  # Removed
  #annotations:
  labels:
    app.kubernetes.io/component: keycloak
    app.kubernetes.io/instance: keycloak
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: keycloak
    helm.sh/chart: keycloak-15.1.1
  name: keycloak
  namespace: keycloak
spec:
  endpoints:
    - interval: 30s
      path: /metrics
      port: http
    - interval: 30s
      path: /realms/master/metrics
      port: http
  selector:
    matchLabels:
      app.kubernetes.io/instance: keycloak
      app.kubernetes.io/name: keycloak
```

</p>
</details>

<details><summary>The difference</summary>
<p>

```diff
--- servicemonitor-keycloak.yaml	2023-05-19 13:43:45.708860839 +0200
+++ podmonitoring-keycloak.yaml	2023-05-19 13:47:15.243825855 +0200
@@ -1,31 +1,27 @@
-apiVersion: monitoring.coreos.com/v1
-kind: ServiceMonitor
[...]

PR #480, opened 2023-05-19
Include Endpoint index in job_name (#479)

Zebradil:
This fixes the issue described in #479.

### Summary

- A ServiceMonitor resource from the monitoring.coreos.com/v1 API can contain multiple endpoints with the same port name but different paths in its spec.
- Scraping job names generated from ServiceMonitor resource contain endpoint indexes (e.g. `ServiceMonitor/ns1/name1/0`, `ServiceMonitor/ns1/name1/1`, see the [implementation](https://github.com/prometheus-operator/prometheus-operator/blob/b5327e7e66e818b8b924332191d4888d45d270e4/pkg/prometheus/promcfg.go#L1075)).
- A PodMonitoring resource from the monitoring.googleapis.com/v1 API can contain multiple endpoints with the same port name but different paths in its spec.
- Scraping job names generated from PodMonitoring resource *does not* contain endpoint indexes (e.g. `PodMonitoring/ns1/name1/http`, `PodMonitoring/ns1/name1/http`, see the [implementation](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/b353ebdee8236c5daa7c6cc93d0d23c0d3c1cd13/pkg/operator/apis/monitoring/v1/types.go#L244)). This way job names get duplicated.
- Prometheus does not allow duplicated job names in a scraping configuration.

This PR includes endpoint indexes into job names to fix the issue. However, the resulting job names' format differs from the upstream format (which does not include a port name):

```diff
- ServiceMonitor/ns1/name1/0
+ PodMonitoring/ns1/name1/http/0
```

I'm not entirely sure if the format should match the upstream or not.

### DIFF
--- a/pkg/operator/apis/monitoring/v1/types.go
+++ b/pkg/operator/apis/monitoring/v1/types.go
@@ -244,6 +244,10 @@ func (p *PodMonitoring) GetKey() string {
 	return fmt.Sprintf("PodMonitoring/%s/%s", p.Namespace, p.Name)
 }
 
+func (p *PodMonitoring) GetEndpointKey(index int) string {
+	return fmt.Sprintf("%s/%s/%d", p.GetKey(), &p.Spec.Endpoints[index].Port, index)
+}
+
 func (p *PodMonitoring) GetStatus() *PodMonitoringStatus {
 	return &p.Status
 }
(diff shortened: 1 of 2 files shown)


pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -698,7 +706,7 @@ func endpointScrapeConfig(id, projectID, location, cluster string, ep ScrapeEndp
 	scrapeCfg := &promconfig.ScrapeConfig{
 		// Generate a job name to make it easy to track what generated the scrape configuration.
 		// The actual job label attached to its metrics is overwritten via relabeling.
-		JobName:                 fmt.Sprintf("%s/%s", id, &ep.Port),
+		JobName:                 id,
```
bwplotka, line 709: Unfortunately we have to be more careful here in terms of breaking change, so let's see in https://github.com/GoogleCloudPlatform/prometheus-engine/issues/479 if we want to pursue this and how 👍🏽

Zebradil: Yes, I agree.

I'm keeping this PR open as a draft until a better way is found in #479. Then we'll see if this PR can be improved or if it's better to make a new one.

Or, if you think it makes sense to just close it now to get it out of the list, feel free :-)

PR #481, opened 2023-05-19
Automatically detect export library version for user agent

TheSpiritXIII:
Tested by creating a new Go binary:

```go
package main

import (
	"fmt"

	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/export"
)

func main() {
	version, err := export.Version()
	fmt.Printf("Hello! %q %s", version, err)
}
```

Works as expected by extracting from `go.mod`!

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -35,6 +35,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      # Fetch the git tree so we can call `git` commands to validate the tag.
+      # TODO(TheSpiritXIII): Remove this when debug.ReadBuildInfo() is fixed.
+      with:
+        fetch-depth: 0
     - name: Run unit tests
       run: make test
   e2e-test:
--- a/pkg/export/export.go
+++ b/pkg/export/export.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"math"
+	"os"
 	"os/exec"
 	"runtime/debug"
 	"strings"
@@ -506,17 +507,25 @@ func Testing() bool {
 // restrictions. While testing, the static version is validated for correctness.
 func Version() (string, error) {
 	if Testing() {
-		// TODO(TheSpiritXIII): After https://github.com/golang/go/issues/50603 just return empty string here.
+		// TODO(TheSpiritXIII): After https://github.com/golang/go/issues/50603 just return empty
+		// string here. For now, use the opportunity to confirm that the static version is correct.
+		// We manually get the closest git tag if the user is running the unit test locally, but
+		// fallback to the GIT_TAG environment variable in case the user is running the test via
+		// Docker (like `make test` does by default).
 		cmd := exec.Command("git", "describe", "--tags", "--abbrev=0")
-		var stdout, stderr bytes.Buffer
+		var stdout bytes.Buffer
 		cmd.Stdout = &stdout
-		cmd.Stderr = &stderr
 
+		version := ""
 		if err := cmd.Run(); err != nil {
-			return "(devel)", nil
+			version = strings.TrimSpace(os.Getenv("GIT_TAG"))
+			if version == "" {
+				return "", errors.New("unable to detect git tag, please set GIT_TAG env variable")
+			}
+		} else {
+			version = strings.TrimSpace(stdout.String())
 		}
 
-		version := strings.TrimSpace(stdout.String())
 		if version != mainModuleVersion {
 			return "", fmt.Errorf("export version %q does not match tag %q", mainModuleVersion, version)
 		}
@@ -524,6 +533,9 @@ func Version() (string, error) {
 		// To prevent breaking in unit tests.
 		return version, nil
 	}
+
+	// TODO(TheSpiritXIII): Due to https://github.com/golang/go/issues/50603 we must use a static
+	// string for the main module (when we import this function locally for binaries).
 
 	bi, ok := debug.ReadBuildInfo()
 	if !ok {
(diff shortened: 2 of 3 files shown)


pkg/export/export.go
```diff
@@ -475,12 +482,44 @@ func (e *Exporter) triggerNext() {
...
+		return "", nil
+	}
+
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return "", fmt.Errorf("unable to retrieve build info")
+	}
+
+	var exportDep *debug.Module
+	for _, dep := range bi.Deps {
```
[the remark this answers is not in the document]
TheSpiritXIII: Much clean! Thanks! Made this change.

pintohutch:
Actually - double-checking this works in the rule-evaluator binary, which is housed in this repo.

TheSpiritXIII:
@pintohutch please take another look. I editted the code so that unit tests fail if the latest tag does not match the static version in the binary after determining that passing in the version information into Docker was not worth it because we would need to update our internal tooling which directly calls docker build instead of our Makefile from our internal repository. I felt this was easier when Golang supports stamping git tags into the main binary. This would require least effort to update to.

pkg/export/export.go
```diff
@@ -475,12 +484,66 @@ func (e *Exporter) triggerNext() {
...
+	moduleName = "github.com/GoogleCloudPlatform/prometheus-engine"
 )
 
+// Testing returns true if running within a unit test.
+// TODO(TheSpiritXIII): Replace with https://github.com/golang/go/issues/52600
+func Testing() bool {
+	return flag.Lookup("test.v") != nil
+}
+
+// Version is used in the User Agent.
```
bwplotka, line 503: Can we add more commentary on behavior of version function? Sounds like it does a lot depending if it's rule evaluator, prometheus engine or run in testing mode?  It's not trivial.

pkg/export/export.go
```diff
@@ -475,12 +484,66 @@ func (e *Exporter) triggerNext() {
...
 }
 
-// ClientName and Version are used to identify to User Agent. TODO(maxamin): automate versioning.
 const (
+	// ClientName is used to identify the User Agent.
 	ClientName = "prometheus-engine-export"
-	Version    = "0.6.3"
+	// Align with git tag.
+	// TODO(TheSpiritXIII): Remove with https://github.com/golang/go/issues/50603
+	exportVersion = "v0.7.0-rc.0"
```
bwplotka, line 492: sounds like Version can give different version then this - can we elaborate what version it is then with comment or var name? Default? Rule Eval one?

bwplotka: Looks like it is a main module version (not a version used if export is a dependency)

bwplotka (commented):
Nice, some readability suggestions only.

pkg/export/export.go
TheSpiritXIII: Renamed to `mainModuleVersion`.

pkg/export/export.go
TheSpiritXIII: Done!

TheSpiritXIII:
@bwplotka thanks for the initial review! I made changes requested further feedback from you! I also added a Makefile change so that the unit test could run within Docker (and within GitHub).

.github/workflows/presubmit.yml
```diff
@@ -35,6 +35,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
```
pintohutch, line 39: Just curious what this does

pkg/export/export.go
```diff
@@ -475,12 +485,74 @@ func (e *Exporter) triggerNext() {
...
+// this function is imported as a library. However, the version is statically
+// set if this function is used in a binary in prometheus-engine due to Golang
+// restrictions. While testing, the static version is validated for correctness.
+func Version() (string, error) {
+	if Testing() {
+		// TODO(TheSpiritXIII): After https://github.com/golang/go/issues/50603 just return empty string here.
+		cmd := exec.Command("git", "describe", "--tags", "--abbrev=0")
+		var stdout, stderr bytes.Buffer
+		cmd.Stdout = &stdout
+		cmd.Stderr = &stderr
```
pintohutch, line 514: Is initiating a `stderr` necessary? I don't see it used anywhere.

pkg/export/export.go
```diff
@@ -475,12 +485,74 @@ func (e *Exporter) triggerNext() {
...
+
+		version := ""
+		if err := cmd.Run(); err != nil {
```
pintohutch, line 517: Can you add a comment describing what this is doing? e.g. `// Prefer version from GIT_TAG env var, otherwise fallback to closest git tag`

pkg/export/export.go
```diff
@@ -475,12 +485,74 @@ func (e *Exporter) triggerNext() {
...
+		// To prevent breaking in unit tests.
+		return version, nil
+	}
+
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return "", fmt.Errorf("unable to retrieve build info")
+	}
+
+	if bi.Main.Path == mainModuleName {
```
pintohutch, line 539: Same here wrt comment, e.g. `// If binary is within repo, fallback to hardcoded version.`

pintohutch (approved):
% nits and questions

bwplotka (approved):
Quite convoluted especially interesting the special case for testing. Probably shows we we do something too tricky on testing side (: 

But needed for now, so LGTM, thanks for great docs on those 💪🏽

TheSpiritXIII:
Convuluted indeed, but the toil may be something that automation can improve.

PR #482, opened 2023-05-19
Replace github.com/pkg/errors with standard library methods

TheSpiritXIII:
For ease of code review, see individual commits!

Invalid usages were fixed:
- Any usages where there was no formatting but a format was specified, e.g. Wrapf or Errorf, was changed to New.
- Any usages that formatted an error with %v or %s was changed to %w.

### DIFF
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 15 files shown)


bwplotka (approved):
Thanks for this! My personal preference is to use https://github.com/efficientgo/core/tree/main/errors - I generally don't like the `%w` non typed pattern and lack of stack trace, but std wrapping is lightweight for our work here, so fine 💪🏽

PR #483, opened 2023-05-19
Bump up minimum export library to Prometheus 2.41

TheSpiritXIII:


bwplotka (approved):
Thanks!

go.mod
```diff
@@ -147,4 +149,4 @@ exclude (
 
 // Go modules keeps resetting the required v0.35.0 to a random unversioned commit.
 // So this is required for unknown reasons.
-replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.35.0
+replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.41.0
```
pintohutch, line 152: nit: curious if this is still necessary

go.mod
```diff
@@ -3,118 +3,120 @@ module github.com/GoogleCloudPlatform/prometheus-engine
 go 1.18
 
 require (
-	cloud.google.com/go/compute v1.5.0
-	cloud.google.com/go/monitoring v1.5.0
-	github.com/go-kit/log v0.2.0
+	cloud.google.com/go/compute/metadata v0.2.2
```
pintohutch, line 6: Were all these changes auto-bumped as a result of the Prometheus upgrade followed by z `make presubmit`?

go.mod
TheSpiritXIII: What makes you curious? We need the version bump because of the API changes to build with Prometheus 2.41. Unless you're suggesting we could update other Prometheus client libraries?

pintohutch: This `replace` directive was put in as a hack because go modules was doing something funky. In theory, we should be able to point to `v0.41.0` above where the original Prometheus [dependency](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/fa652aea7a407d3528e06d80c7df20171ac8683f/go.mod#L19) is configured.

TheSpiritXIII: Double checked this!

Looks like without the replace directive, after you run `go mod tidy`, you get `github.com/prometheus/prometheus v1.8.2-0.20211119115433-692a54649ed7` which confirms you still need the directive (sadly).

pintohutch (approved):

PR #484, opened 2023-05-25
Log an error when unable to fetch Google Cloud metadata

TheSpiritXIII:
This is an issue we observed on a single internal cluster. This logging would help us debug why this happened.

### DIFF
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -71,7 +71,11 @@ func main() {
 
 	var defaultProjectID string
 	if metadata.OnGCE() {
-		defaultProjectID, _ = metadata.ProjectID()
+		var err error
+		defaultProjectID, err = metadata.ProjectID()
+		if err != nil {
+			level.Error(logger).Log("msg", "Unable to fetch Google Cloud project", "err", err)
+		}
 	}
 
 	reg := prometheus.NewRegistry()
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"net/http"
 	"os"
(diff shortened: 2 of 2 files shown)


cmd/rule-evaluator/main.go
```diff
@@ -71,7 +71,11 @@ func main() {
 
 	var defaultProjectID string
 	if metadata.OnGCE() {
-		defaultProjectID, _ = metadata.ProjectID()
+		var err error
+		defaultProjectID, err = metadata.ProjectID()
```
macxamin, line 75: Is there a library we can use to verify these are valid?

TheSpiritXIII: Valid in which way? Formatting or correctness?

In terms of formatting, users can run outside of GKE. I may be wrong but project would just a label plain and could have any arbitrary name.

In terms of correctness, we could build an integration test that when running GMP in a GKE project, the exported metric's project label matches the GKE project it's running on. Currently we do not check the label value.

We do at the very least ensure that the project ID is non-empty.

Let me know your thoughts!

macxamin: Valid as it exists in GCP and is reachable. Not sure how to do it. Then we can alert the user to fix. Is it possible we may have projects showing up on alerts and SLO that are not using the correct projectID, cluster, etc?

pintohutch (approved):

cmd/rule-evaluator/main.go
TheSpiritXIII: I'm not sure if that's desirable because:
1. If it's empty, that's already an error.
2. If the user does nothing, this code that I'm updating already detects the correct one.
3. The user may set the `--project-id` flag purposefully to something different.

#2 would be the only time we'd want to validate, and I can't imagine when it would be wrong. An integration test could maybe be created for this case, but it would rarely fail.

Either way, it's out of scope for this PR so I will be resolving. Thank you!

PR #492, opened 2023-06-16
Set default interval for ScrapeEndpoint

### LINKED ISSUE #491: [bug/docs] The operator requires endpoint's scrape interval, but docs say it's optional
According to the docs, `ScrapeEndpoint.interval` is optional (`required == false`, [ref](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.7.0/doc/api.md#scrapeendpoint)). However, I can't apply the following resource:

```yaml
kind: PodMonitoring
metadata:
  name: argocd-metrics
  namespace: argocd
spec:
  endpoints:
    - port: metrics
  selector:
    matchLabels:
      app.kubernetes.io/name: argocd-metrics
```

The validation error:

> admission webhook "validate.podmonitorings.gmp-operator.gmp-system.monitoring.googleapis.com" denied the request: invalid definition for endpoint with index 0: invalid scrape interval: empty duration string

I guess, the docs should be corrected.

bernot-dev:
Fix #491

Also adding validation for the field, based on the [regex provided in the Prometheus duration spec](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file).

### DIFF
--- a/pkg/operator/apis/monitoring/v1/types.go
+++ b/pkg/operator/apis/monitoring/v1/types.go
@@ -988,6 +988,8 @@ type ScrapeEndpoint struct {
 	// Proxy URL to scrape through. Encoded passwords are not supported.
 	ProxyURL string `json:"proxyUrl,omitempty"`
 	// Interval at which to scrape metrics. Must be a valid Prometheus duration.
+	// +kubebuilder:validation:Pattern="^((([0-9]+)y)?(([0-9]+)w)?(([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?(([0-9]+)ms)?|0)$"
+	// +kubebuilder:default="1m"
 	Interval string `json:"interval,omitempty"`
 	// Timeout for metrics scrapes. Must be a valid Prometheus duration.
 	// Must not be larger then the scrape interval.
--- a/cmd/operator/deploy/crds/monitoring.googleapis.com_clusterpodmonitorings.yaml
+++ b/cmd/operator/deploy/crds/monitoring.googleapis.com_clusterpodmonitorings.yaml
@@ -90,7 +90,9 @@ spec:
                       x-kubernetes-int-or-string: true
                     interval:
                       type: string
+                      default: 1m
                       description: Interval at which to scrape metrics. Must be a valid Prometheus duration.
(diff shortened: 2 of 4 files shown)


pkg/operator/apis/monitoring/v1/types.go
StevenYCChou, line 1: Can you add tests to the corresponding path pkg/operator/apis/monitoring/v1/types_test.go on the default value and maybe a simple smoke test for the validation pattern?

pkg/operator/generated/clientset/versioned/clientset.go
StevenYCChou, line 1: Are the code under `pkg/operator/generated/clientset/versioned/clientset.go` generated based on the changes in  `pkg/operator/apis/monitoring/v1/types.go` and you didn't manually write any code here? If that's the case, please just resolve this comment.

StevenYCChou (changes requested):

pkg/operator/generated/clientset/versioned/clientset.go
bwplotka: Yea I think it feels like not releated changes to the main focus of this PR 🤔

bwplotka (commented):
Thanks! 

I think I am fine with types.go and yaml changes, but clientset and factory changes feels unrelated? 🤔 Maybe some different generator version? Should we change that in separate PR?

bernot-dev:
> clientset and factory changes feels unrelated? 🤔 Maybe some different generator version? Should we change that in separate PR?

From what I can tell, those files are related somehow...the clientset/factory generated files don't change when I run `make regen` before adding the annotations. Exactly how the generated files are related to the actual change I made is beyond me. But it doesn't seem like the code generator thinks they are unrelated.

pkg/operator/apis/monitoring/v1/types.go
StevenYCChou: [bernot-dev](https://github.com/bernot-dev) reached out to me offline that we probably need to add a e2e test running on KIND, not `pkg/operator/apis/monitoring/v1/types_test.go` - and I agree as the change is on the `apiextensions.k8s.io/v1`. 

Thanks for bringing this up, Adam!

bernot-dev:
@bwplotka You were right, it was skipping codegen when it didn't see that the files had been updated. #497 updates the clientset and factory generated files separately.

bwplotka:
So this PR has to be rebased now, right?

bwplotka (approved):
Thanks 💪🏽

PR #497, opened 2023-06-21
Update generated files with controller-gen v0.12.0

bernot-dev:
Update clientset and factory generated files for changes introduced by upgrading `controller-gen`.


bwplotka (approved):
Thanks! Is there a way we can pin the version in our tooling? or at least document?

bwplotka:
We could use https://github.com/bwplotka/bingo (just saying).

PR #506, opened 2023-06-29
ui: Changed script to remove known unused and insecure packages.

bwplotka:
We really don't want to vendor packages we don't to build our UI

Unfortunately it's not so easy. Simply doing `npm ci --omit=dev` or `npm install --omit=dev` that in theory should only install required deps to build, still does not install enough deps and blocks npm build. I think (not an expert here), this is because some package authors misunderstand what devDeps vs normal deps are and put production packages to "dev" e.g. [lezer-promql](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/third_party/prometheus_ui/base/web/ui/module/lezer-promql/package.json#L32). I tried to install deps with --omit=dev, and manually install those deps for lezer-promql, but similar problem occurs on other packages.

For now we manually remove known, vulnerable package from this repo. Note that we have same problem to solve on https://github.com/GoogleCloudPlatform/prometheus/tree/release-2.35.0-gmp/web/ui/node_modules/rst-selector-parser


bwplotka:
Tested changes (make frontend and manual browse of UI on built container)

TheSpiritXIII (approved):
Nice job!

PR #510, opened 2023-07-05
Refactor cleanup of old resources

bernot-dev:
- Leverage the client.IgnoreNotFound function to improve readability of some error handling.
- Wrap errors from controller-runtime.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -397,12 +397,12 @@ func (o *Operator) Run(ctx context.Context, registry prometheus.Registerer) erro
 func (o *Operator) cleanupOldResources(ctx context.Context) error {
 	// Delete old ValidatingWebhookConfiguration that was installed directly by the operator
 	// in previous versions.
-	err := o.client.Delete(ctx, &arv1.ValidatingWebhookConfiguration{
+	validatingWebhookConfig := arv1.ValidatingWebhookConfiguration{
 		ObjectMeta: metav1.ObjectMeta{Name: "gmp-operator"},
-	})
-	if err != nil && !apierrors.IsNotFound(err) {
+	}
+	if err := o.client.Delete(ctx, &validatingWebhookConfig); client.IgnoreNotFound(err) != nil {
 		o.logger.Error(err, "deleting legacy ValidatingWebhookConfiguration")
-		return err
+		return fmt.Errorf("delete legacy ValidatingWebHookConfiguration: %w", err)
 	}
 
 	// If cleanup annotations are not provided, do not clean up any further.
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -412,34 +412,36 @@ func (o *Operator) cleanupOldResources(ctx context.Context) error {
 
 	// Cleanup resources without the provided annotation.
 	// Check the collector DaemonSet.
-	var ds appsv1.DaemonSet
-	if err := o.client.Get(ctx, client.ObjectKey{
+	dsKey := client.ObjectKey{
```
bwplotka, line 415: Is this really necessary? Inline looks better IMO. Also extra variable tells us usually that this value is reused, but it's not, so it's bit misleading (: 

See the rationales on the style guide I wrote for Thanos project: https://github.com/thanos-io/thanos/blob/main/docs/contributing/coding-style-guide.md#avoid-defining-variables-used-only-once

pkg/operator/operator.go
```diff
@@ -412,34 +412,36 @@ func (o *Operator) cleanupOldResources(ctx context.Context) error {
...
 			o.logger.Error(err, "cleaning up collector DaemonSet")
-			return err
+			return fmt.Errorf("delete collector DaemonSet: %w", err)
 		}
 	}
 
 	// Check the rule-evaluator Deployment.
-	var deploy appsv1.Deployment
-	if err := o.client.Get(ctx, client.ObjectKey{
+	deployKey := client.ObjectKey{
```
bwplotka, line 432: ditto

pkg/operator/operator.go
```diff
@@ -412,34 +412,36 @@ func (o *Operator) cleanupOldResources(ctx context.Context) error {
...
-	var ds appsv1.DaemonSet
-	if err := o.client.Get(ctx, client.ObjectKey{
+	dsKey := client.ObjectKey{
 		Name:      NameCollector,
 		Namespace: o.opts.OperatorNamespace,
-	}, &ds); err != nil && apierrors.IsNotFound(err) {
+	}
+	var ds appsv1.DaemonSet
+	if err := o.client.Get(ctx, dsKey, &ds); apierrors.IsNotFound(err) {
 		o.logger.Error(err, "Getting collector DaemonSet failed")
```
bwplotka, line 421: I know it's not your problem, but do we really want to double handle the error? It's kind of big anti-pattern (the returned error is likely logged as well or handled in some other way 🤔 )

bwplotka (approved):
Nice! Some nits, otherwise LGTM (:

pkg/operator/operator.go
bernot-dev: I generally agree that defining a local variable to only be used once is not ideal. My thought was that a local variable allows for less visual clutter in the function call. Because the client.ObjectKey is the second of three parameters in the Get method, the arguments end up separated over several lines of code and (I think) more difficult to read. My eyes have to go searching for the third argument after the inline definition of the second argument. As a colleague once pointed out to me, this issue would be less pronounced if the key was the final parameter instead of the middle parameter of controller-runtime. But that bell likely cannot be un-rung, at this point.

Go has made several choices, like the [omission of ternary expressions](https://go.dev/doc/faq#Does_Go_have_a_ternary_form), which intend to make code more readable. In my mind, I extend that idea to trying to trying to make sure that not too much logic is crammed into a single line of code.

Between the inline object definition and the compound conditional with a short assignment, the existing code crossed my personal threshold for complexity in a single line.

pkg/operator/operator.go
bernot-dev: Good point. I suspect that returning the error would be sufficient.

bwplotka: Do you want to check it, before merging?

bernot-dev: Verified that these errors bubble up to the top level of the application, where they are [logged before the program exits](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/cmd/operator/main.go#L164). Extra logging messages removed.

PR #511, opened 2023-07-07
add updateversions to presubmit

pintohutch:
This will ensure our manifests are auto-updated during `presubmit`


bwplotka (approved):
LGTM!

PR #514, opened 2023-07-13
export: Exposed internal shard + buffer sizing flags as exorted options.

bwplotka:
This is useful to perform e2e benchmarks to fine tune more. It also allows advanced configuration for customers in unique environments.


TheSpiritXIII (approved):
Thanks!

pintohutch (approved):
Nice!

PR #516, opened 2023-07-14
Update kube-state-metrics.yaml ClusterPodMonitoring object to include…

cdmello-g:
… persistentvolumes

Currently ClusterPodMonitoring regex drops PersistentVolumes, this will now include this object type and prevent it from being dropped.

Essentially this also fixes empty GMP PV/HPA dashboards.

### DIFF
--- a/examples/kube-state-metrics/kube-state-metrics.yaml
+++ b/examples/kube-state-metrics/kube-state-metrics.yaml
@@ -318,7 +318,7 @@ spec:
     interval: 30s
     metricRelabeling:
     - action: keep
-      regex: kube_(daemonset|deployment|pod|namespace|node|statefulset|persistentvolumes)_.+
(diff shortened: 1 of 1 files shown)


cdmello-g:
@lyanco & @yqlu

yqlu (approved):

cdmello-g:
@yqlu sure thing i'll update there as well!

cdmello-g:
@yqlu I've added HPA as well now - I'll also start on updating the docs

lyanco:
@yqlu looks like managed ksm already includes PV and HPA metrics

yqlu:
Right, exactly, so because of that we should include them in this copy-pastable config too so that this is a superset of the managed config and will populate the OiC dashboards correctly.

lyanco:
Ah, ok, I misread what you wrote, yes it all makes sense to me.

PR #518, opened 2023-07-14
Add postStart hook to remove WAL data

### LINKED ISSUE #459: Prometheus Crash Looping on startup
Screen Shot 2023-04-17 at 11 16 03

* Related to https://github.com/prometheus/prometheus/issues/6934

I'm trying to change the resource limits but since it is managed, I do not have the permissions to edit / modify anything in the `gke-gmp-system` namespace.

How can I reset the pod so that it wipes the wal directory. There was an issue where it was scraping too many timeseries, and caused it to get into this crash looping situation. I limited the number of timeseries but now the collector is stuck and can't progress.

Any recommendations?

bernot-dev:
Fix #459

### DIFF
--- a/cmd/operator/deploy/operator/10-collector.yaml
+++ b/cmd/operator/deploy/operator/10-collector.yaml
@@ -140,6 +140,10 @@ spec:
             port: 19090
             path: /-/ready
             scheme: HTTP
+        lifecycle:
+          postStart:
+            exec:
+              command: ["sh", "-c", "rm -r /prometheus/data/wal/*"]
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
--- a/manifests/operator.yaml
+++ b/manifests/operator.yaml
@@ -646,6 +646,10 @@ spec:
             port: 19090
             path: /-/ready
             scheme: HTTP
+        lifecycle:
+          postStart:
+            exec:
+              command: ["sh", "-c", "rm -r /prometheus/data/wal/*"]
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:


bernot-dev:
I'm wondering if disabling the WAL may be a cleaner approach, if we're always going to drop the data anyways.
https://github.com/prometheus/prometheus/blob/main/tsdb/db.go#L95

bwplotka:
Thanks. 

The only missing thing is the ask from @lyanco make this configurable from the operator side. This is due to the fact that persistent and proper WAL handling is great for local debuggabilitity. We want customers to enable/disable it on the fly.

For the same reason we don't want to disable WAL from the Prometheus code. There are a few other reasons:
* We want to be prepared for resource usage when we will decide to maintain persistence with remote write. This is to manage expectations going forward.
* We don't want to diverge too much with the fork, especially given plan to switch to native remote write at some point.

bwplotka:
Do we have precedense on customizing DaemonSet yaml based on operator config @pintohutch @TheSpiritXIII ?

bernot-dev:
@bwplotka Based on our discussions, my opinion is that the best option is to proceed with implementing the postStart lifecycle hook (the current PR), rather than disabling the WAL. This solution will continue to allow customers to access local data in most circumstances (except immediately after a restart), and is expected to resolve the reported issue.

In additional, the actual window of data loss should be expected to be small, as most data is being sent to the GCP backend quickly - on the order of seconds. So, the scenario where this change would impact customers is only when the collector crashes/restarts and either:

- Data of interest was collected immediately preceding a restart, between the time that it was written to the WAL and when it is replicated to GCP.
- Customer is interested in debugging **local** data only, preceding a restart. Data stored in GCP is not a suitable alternative.

In the more somewhat more common case where a customer is interested in looking at local data and the collector has not crashed recently, there would effectively be no difference for their use case, based on this change.

We may want to re-evaluate creating a configuration option to disable the WAL in the future if we have additional reasons to do so.

Let me know if you have any additional comments/concerns.

bwplotka:
Thanks. What you said makes sense, it is hypothetically rare. However I see also a use case where users are debugging why metric was not exported (e.g misconfigured filtering). I still see your point that crash/restart reset might be fine if you are debugging metric which is constantly scraped, but the problem might be when user ingested something 1h ago and is debugging now where sample steam is blocked (is sample not scraped? Scraped but not exported? Not exported why?).
Another unknown is self-deployed OSS users. Are we sure we know all their cases? Ideally we let them opt-in to "kill your data on start" - in some way we are not managing their setups to decide about it. By adding this hook, we or user will not have much control over any of this.

This is why I would vote for adding opt-out of force data removal setting for managed GMP from the start. For unmanaged it should be opt-in if possible. Thanks to @pintohutch [mention](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/518#issuecomment-1642300774) it should be relatively simple to do. We could add [OperatorConfig Collector setting](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/ff0246ffd6f057601607da2e768dd9da7e5d7b4d/pkg/operator/apis/monitoring/v1/types.go#L97C9-L97C10) "ForceEmptyStorageOnStart bool" (with the right commentary that it's only "local" persistence, for export it does not work yet - might in future). By default **false** for self-deployed, **true** by default on GMP managed. (Calling this setting "Persistence bool" might be misleading - even with true export is not truly persistent). Essentially, given we are theoretically imagining certain cases, it's likely we are wrong and we miss important use case.

This setting can control postHook (by templating collector DaemonSet), but I am not sure I am a fan of this. It's generally [racy](https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#discussion) and we have full control of Prometheus to do it there. We could add hidden flag `--storage.debug.remove-data-on-start` (default false) flag to our Prometheus fork that will explicitly remove all files from "data" directory. Alternatively we can try playing with WAL disable flag (perhaps it discards any WAL?), but remove data flag will be safer and faster (less tests to make), as users might have already corrupted or large WALs. Thanks of Compression precedence passing flag, this should be trivial in operator.

However, we could also skip adding custom code to Prometheus fork and we can simply [change collector's DaemonSet data dir to memory](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) when "ForceEmptyStorageOnStart" is true. The only minor nit is that it will improve perf characteristics -- might be not a bad thing as well (: Another nice thing about this is that there is no need to explain or debug any logic of synthetic removal. The memory is ephemeral by design, no extra logic to maintain.

To sum up: 

* I would still expect a setting in OperatorConfig that controls this behaviour. Data consistency is important for users. CC @lyanco opinion would be valuable (: 
  * Ideally we don't remove data dir for self-deploy users
* PostStart Hook might be not the ideal solution, why not changing emptyDir medium to memory? Or collector custom flag.

bernot-dev:
In my testing, if the container hit an OOM with `emptyDir.medium = Memory`, the data persisted. The data is not persisted through a **node** restart, but it did seem to survive a **container** restart. So, that alone would not solve #459.

bernot-dev:
Regarding `ForceEmptyStorageOnStart`, it does not seem like configuration to me, per se. As I understand it, the only use case to opt in for that config option would be because you're running into a bug. But removing the data on startup would not actually result in any runtime differences in the program. So, putting that option alongside other configuration options that all correspond to runtime functionality feels strange to me.

As I think about it, I think the most narrowly tailored solution for the problem that was reported simply allows a customer using a GKE Autopilot cluster a mechanism to delete the data and restart the container as a troubleshooting step. It is likely this would only need to be done once, rather than every time the container starts.

For unmanaged instances, we could probably just add an appropriate `kubectl exec -- rm` command to the troubleshooting runbook.

Alternatively (in managed/autopilot environments), we could watch for OOM events and use that event as a trigger to reset the data in the container. It's theoretically possible to have some OOM events that are not related to this issue, but it would have greater accuracy than unconditionally wiping the data every startup.

bwplotka:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/518#issuecomment-1648331138 - wow nice, good to know, thanks!

Cool, all of those are nice ideas, all have some trade-offs.  I agree runtime/environment impression around this configuration, the alternative is to configure patch Daemon Set based on that config. It is weird, but other options (Autopilot node restart and `kubectl exec -- rm`) are far less scalable and more painful for support and customers. We know it's safe to kill it so why not doing it always until we have full persistence?

bwplotka:
We need more eyes on this too, so short proposal outlining option might be amazing if we want to progress on it, WDYT?

Thanks for elaborating 💪🏽

PR #519, opened 2023-07-14
Update kind used in kindtest target

bernot-dev:
Update from v0.11.1 to v0.20.1. This resolves a bug on my local development environment where kind cluster fails to create.


TheSpiritXIII (approved):
Thank you!

PR #526, opened 2023-07-25
config-reloader: Remove readiness check

bwplotka:
This PR is part of larger redesign of config-reloader alerting and SLI pipeline discused internally.

TL;DR: To denoise our internal alerts and SLI we remove the config-reloader readiness check. There is no benefit in checking readiness only at the container start compared to trying to reload and fail reload HTTP request straight away. Furthermore, we don’t check readiness after the initial check (target component can restart anytime), so we increment reloader_reloads_failed_total anyway (contradicting initial [rationale](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/380)).

Removal will give us:
* Consistency and predictable failure mode for config reloader (all problematic error cases will consistently increment reloader_reloads_failed_total).
* We stop restarting the container due to a broken target.
* No extra readiness configuration item to maintain and set.

### DIFF
--- a/cmd/config-reloader/main.go
+++ b/cmd/config-reloader/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/thanos-io/thanos/pkg/reloader"
 )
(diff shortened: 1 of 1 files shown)


cmd/config-reloader/main.go
```diff
@@ -56,8 +59,8 @@ func main() {
 
 	metrics := prometheus.NewRegistry()
 	metrics.MustRegister(
-		prometheus.NewGoCollector(),
-		prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}),
+		collectors.NewGoCollector(),
```
bwplotka, line 62: Moved to non-deprecated structs - the same behaviour.

bwplotka:
Thanks, responded in the document, let's keep discussion in one place.

bwplotka:
NOTE: We are discussing the next steps internally.

bwplotka:
Changing the front as discussed in the proposal. We propose to upstream this sidecar functionality to help not only ourselves, but also wider community: https://github.com/prometheus/prometheus/issues/9783#issuecomment-1669341273

bwplotka:
Since then we went back - we might need sidecar anyway. This PR is still relevant to avoid unnecessary restarts which only blurs what's happening. Let's only increment errors.

bwplotka:
> I'm not sure if this is the right solution. I would have preferred either:
> To go the other way, where we use the readyness check on restarts (in addition to startups).

I explained how this is non trivial to implement (race conditions and also expensive) in our design doc.

> Keep initial readiness check to prevent the initial error total from incrementing at startup but use a metric that only increments once per-error (since current reloader_reloads_failed_total is noisy). 

Hm, is it really helpful though? `reloader_reloads_failed_total` is noisy, so it's the restart signal these days (on healthy Prometheus we restart at least one config-reloader (!)). With this change we at least move to one noisy metric and we can iterate later further.

The only benefit of pre check is it to treat `reloader_reloads_failed_total` metric as situation when we "definitely" configured Prometheus well, just it's not healthy (because it responded once) - not sure if this is helpful for us though.

Also, any other metric will require new alert and new plumbing too.

Let's think about it more, moving to draft.

bwplotka:
We didn't get consensus on this, so closing for now.

PR #532, opened 2023-07-27
Pass PR number to privileged job.

bwplotka:


bwplotka:
Tmp fork until https://github.com/pozil/auto-assign-issue/pull/115 is merged.

.github/workflows/auto-assign-issue.yml
```diff
@@ -0,0 +1,16 @@
+name: (Privileged) Auto Assign
```
TheSpiritXIII, line 1:
```suggestion
name: Auto Assign Issue
```
This one isn't privileged.

bwplotka: It still is (:

bwplotka: but yea, no biggie

.github/workflows/auto-assign-pr.yml
```diff
@@ -0,0 +1,53 @@
...
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+    - name: 'Download parameter'
+      uses: actions/github-script@v3.1.0
+      with:
+        script: |
+          var artifacts = await github.actions.listWorkflowRunArtifacts({
```
TheSpiritXIII, line 22: I would use `const` here in place of all the `var`s.

`var` isn't really used in JavaScript anymore (`let` is for mutable references).

TheSpiritXIII (approved):
Thanks!

.github/workflows/auto-assign-pr.yml
bwplotka: Let me fix this later, want to first try if this is a right path

Issue #536, opened 2023-07-28
examples: Add resource.requests to all containers


bwplotka:
This is due to https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#defaults which causes our example pods to not work in the typical GKE autopilot cluster as unrealistic requests is assigned (2GiB mem).

![image](https://github.com/GoogleCloudPlatform/prometheus-engine/assets/6950331/ae9ddd3e-faab-49e1-93c0-6a55fc455e67)

PR #537, opened 2023-07-28
Added requests resources to examples/example-app.yaml

bwplotka:
Related to https://github.com/GoogleCloudPlatform/prometheus-engine/issues/536

### DIFF
--- a/examples/example-app.yaml
+++ b/examples/example-app.yaml
@@ -43,6 +43,7 @@ spec:
                 values:
                 - linux
       containers:
+      # Instrumentation source: https://github.com/nilebox/prometheus-example-app
       - image: nilebox/prometheus-example-app@sha256:dab60d038c5d6915af5bcbe5f0279a22b95a8c8be254153e22d7cd81b21b84c5
         name: prom-example
         ports:
@@ -52,3 +53,7 @@ spec:
         - "/main"
         - "--process-metrics"
         - "--go-metrics"
+        resources:
+          requests:
+            cpu: 100m
+            memory: 100Mi


examples/example-app.yaml
```diff
@@ -52,3 +53,7 @@ spec:
         - "/main"
         - "--process-metrics"
         - "--go-metrics"
+        resources:
+          requests:
+            cpu: 100m
```
TheSpiritXIII, line 58: This is significantly higher than other applications we have. Is the example app really more resource intensive [than prometheus](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/manifests/operator.yaml#L630)?

examples/example-app.yaml
```diff
@@ -52,3 +53,7 @@ spec:
...
+            memory: 100Mi
```
TheSpiritXIII, line 59: Same here. We give our [operator `32M`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/manifests/operator.yaml#L267C15-L267C15). Did you check actual usage?

TheSpiritXIII (approved):
Thanks! Left some comments but I know we want to improve the example app so I'm not going to be too picky here.

examples/example-app.yaml
bwplotka: Not really, but not sure if it matters as long as it's not something ridiculous by default (500m)

examples/example-app.yaml
bwplotka: No, will check, but I don't think that's important as long as it's not 2GBi (default on my autopilot cluster)

pintohutch (approved):

PR #540, opened 2023-08-02
Updated Consul exporter yaml snippet

algchoo:
## Changes
* [updated the image in the example to reflect the latest version of consul](https://github.com/GoogleCloudPlatform/prometheus-engine/commit/94545649be6d28030f4ba07ce93ceafd6e92d062)

## Details

For consistency, where the test results for currency tell us:
* The dashboards/alerts are backwards compatible with new version of the application and the exporter

We are opting to keep the `minimum_supported_version` as it is, while the example in the documentation will reflect the latest version of the app/exporter.

### DIFF
--- a/examples/consul/exporter.yaml.snippet
+++ b/examples/consul/exporter.yaml.snippet
@@ -27,7 +27,7 @@ spec:
     spec:
       containers:
       - name: consul
-        image: consul:1.13.3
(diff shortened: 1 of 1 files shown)


yqlu (approved):

bwplotka (approved):
Thanks!

PR #544, opened 2023-08-11
Add syntax validation for GlobalRules

macxamin:
Currently there is no syntax validation for GlobalRules and this causes customers to apply rule files with syntax errors. 

Syntax validation happen in [rules.Scope.](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/3016092cbcfec6181adb274e06aa5ea06503a500/pkg/rules/rules.go#L85) which is called by [GenerateRules and GenerateClusterRules.](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/3016092cbcfec6181adb274e06aa5ea06503a500/pkg/operator/rules.go#L213-L241)

This commit contains:
1. added `rules.Scope` to `GenerateGlobalRules`
2. Added unit test cases for `GenerateRules`, `GenerateClusterRules` and `GenerateGlobalRules`.

### DIFF
--- a/pkg/operator/rules_test.go
+++ b/pkg/operator/rules_test.go
@@ -1,0 +1,248 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package operator
+
+import (
+	"testing"
+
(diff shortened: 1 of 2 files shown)


pkg/operator/rules_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+import (
+	"testing"
+
+	monitoringv1 "github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator/apis/monitoring/v1"
+	"github.com/google/go-cmp/cmp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_generateRules(t *testing.T) {
```
pintohutch, line 11: nit: we typically use the [recommended](https://go.dev/doc/effective_go#mixed-caps) `camelCase` over `snake_case` in our repo.

pkg/operator/rules_test.go
```diff
@@ -0,0 +1,234 @@
+package operator
```
pintohutch, line 1: You'll want to add a [header](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/hack/boilerplate.go.txt).

pintohutch (approved):
LGTM - nice catch!

PR #548, opened 2023-08-17
Bump recommended KSM version

### LINKED ISSUE #547: example metrics with 2.3.0 will no longer suitable for gke 1.25 and later
As GKE dashboard shows, we need to migrate to validate version after gke upgrade to 1.25, as I checked with our own app, all APIs seems fine, but we can see all the api invalidate calls from kube-state-metrics with 2.3.0. we need to separate the version for before 1.25, or we just update the docs directly

yqlu:
v2.8.2 is what our managed equivalent is currently using. This version bump resolves #547 because newer KSM invokes v1/CronJob API endpoints instead of v1beta1/CronJob endpoints, which have been deprecated since v1.21+ and unavailable in v1.25+. See https://github.com/kubernetes/kube-state-metrics/blob/main/CHANGELOG.md#v241--2022-02-10 for more context.

This switch breaks compatibility with K8s <v1.21, but this is not a concern given earliest version in GKE stable release channel is v1.23 and default is currently v1.27.

### DIFF
--- a/examples/kube-state-metrics/kube-state-metrics.yaml
+++ b/examples/kube-state-metrics/kube-state-metrics.yaml
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2023 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


TheSpiritXIII (approved):
Thanks for the quick turnaround (and for scanning the Issues list)!

PR #550, opened 2023-08-18
Update Debian to latest stable release

macxamin:
Need to update to golang:1.20.7 to fix security vulnerabilities.

golang 1.20-buster supports only until golang 1.20.5, it's also the oldoldstable of debian.

golang 1.20-bookworm is the current stable release. https://wiki.debian.org/DebianReleases


TheSpiritXIII (approved):
Looks good. Thanks!

PR #553, opened 2023-08-22
add more buckets to grpc_client_handling_seconds

macxamin:
Currently [grpc_client_handling_seconds](https://github.com/grpc-ecosystem/go-grpc-prometheus/blob/a5446fa2489078304671617eeebd4c5e457dcf07/client_metrics.go#L80-L83) uses [`prom.DefBuckets`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/b07c5f3b326f793f8e411862a2b7e40961cb69f9/vendor/github.com/prometheus/client_golang/prometheus/histogram.go#L261-L265) which has a maximum bucket of 10. We need to add more granularity for our SLO. I added `15, 20, 30, 40, 50, 60`. I will adjust based on usage.

Tested and confirmed working.


bwplotka (approved):
LGTM, a quite many buckets, but should be fine for us. Native histograms will fix it soon (:

PR #555, opened 2023-08-28
Patch client-go

TheSpiritXIII:
A customer reported that this issue happens on GMP and causes crash looping: https://github.com/kubernetes/kubernetes/issues/118361 b/297601248


macxamin (approved):
LGTM

PR #557, opened 2023-08-28
also rm max-count frm go-synthetic.yaml

JuddRogers:
max-count arg was removed by PR #539


TheSpiritXIII (approved):
Thanks Judd! Noticed this too!

PR #563, opened 2023-09-05
Lower resource requests to idle usage

pintohutch:
Lower resource requests to match idle usage

### DIFF
--- a/cmd/operator/deploy/operator/11-rule-evaluator.yaml
+++ b/cmd/operator/deploy/operator/11-rule-evaluator.yaml
@@ -70,8 +70,8 @@ spec:
           limits:
             memory: 32M
           requests:
-            cpu: 8m
-            memory: 16M
+            cpu: 1m
+            memory: 4M
         volumeMounts:
         - name: config
           readOnly: true
@@ -103,8 +103,8 @@ spec:
           limits:
             memory: 1G
           requests:
-            cpu: 8m
-            memory: 32M
+            cpu: 1m
+            memory: 16M
         volumeMounts:
         - name: config-out
           readOnly: true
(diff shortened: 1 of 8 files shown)


cmd/operator/deploy/operator/11-rule-evaluator.yaml
```diff
@@ -103,8 +103,8 @@ spec:
           limits:
             memory: 1G
           requests:
-            cpu: 8m
-            memory: 32M
+            cpu: 1m
+            memory: 16M
```
TheSpiritXIII, line 107: Should we maybe keep this consistent with `cmd/operator/deploy/rule-evaluator/01-deployment.yaml`?

cmd/operator/deploy/operator/12-alertmanager.yaml
```diff
@@ -82,8 +82,8 @@ spec:
           limits:
             memory: 128M
           requests:
-            cpu: 8m
-            memory: 32M
+            cpu: 1m
+            memory: 16M
```
TheSpiritXIII, line 86: Should we maybe keep this consistent with `examples/alertmanager.yaml`?

TheSpiritXIII (approved):
LGTM! Just added some minor optional nits!

cmd/operator/deploy/operator/12-alertmanager.yaml
pintohutch: Thx - done!

cmd/operator/deploy/operator/11-rule-evaluator.yaml
pintohutch: Thx - done!

PR #566, opened 2023-09-06
Update prometheus common library to match prometheus client library

TheSpiritXIII:
Something we missed when updating Prometheus to 2.41. As far as I know, there are no known consequences of not having already done this. I only discovered this because I wanted to add support for HTTP client authentication in a separate change and I was getting some marshalling errors without this.


pintohutch:
What marshaling errors were you seeing?

Looking at https://github.com/GoogleCloudPlatform/prometheus/blob/release-2.41.0-gmp/go.mod#L47, it seems we use 0.39 there too, which is good.

bwplotka (approved):
Still, should be fine to upgrade, no?

LGTM (:

PR #567, opened 2023-09-14
Add BasicAuth support into the example app

TheSpiritXIII:
Example usage:
```
go run ./examples/instrumentation/go-synthetic/ --basic-auth-username=admin --basic-auth-password=pw
curl localhost:8080/metrics -u "admin:pw"
```


pintohutch:
Gonna let @macxamin handle this - dropping myself.

macxamin:
Can you add instructions on how to use it to the README? So we can have a record on how to run with all the different auth modes?

macxamin (approved):
Thanks! LGTM!

PR #568, opened 2023-09-14
Add Authorization header support into the example app

TheSpiritXIII:
Example with bearer token:
```bash
go run ./examples/instrumentation/go-synthetic/ --auth-scheme=Bearer --auth-parameters=xyz
curl -H "Authorization: Bearer xyz" localhost:8080/metrics
```

This PR builds on top of #567.


pintohutch:
I'm going to let @macxamin take point on the review here. I will drop myself.

macxamin (approved):
Thanks! LGTM!

PR #569, opened 2023-09-14
Add OAuth 2.0 support into the example app

TheSpiritXIII:
Example usage:
```bash
go run ./examples/instrumentation/go-synthetic/ --oauth2-client-id=abc --oauth2-client-secret=xyz
curl "localhost:8080/token?grant_type=client_credentials&client_id=abc&client_secret=xyz"
# Fetch access token from above and use as bearer token example below:
curl -H "Authorization: Bearer DZ~9UYwD" localhost:8080/metrics
```

This PR builds on top of #568.


pintohutch:
I'm going to let @macxamin take point on the review here and will drop myself.

bwplotka (approved):
client_credentials grant is bit simplistic for real use cases, but for test purposes - sure (:

PR #570, opened 2023-09-15
Add mTLS support into the example app

TheSpiritXIII:
Example usage:
```bash
go run ./examples/instrumentation/go-synthetic/ --tls-create-self-signed=true
curl -k https://localhost:8080/metrics
```

This PR builds on top of #569.

### DIFF
--- a/examples/instrumentation/go-synthetic/auth.go
+++ b/examples/instrumentation/go-synthetic/auth.go
@@ -1,17 +1,35 @@
 package main
 
 import (
+	"crypto/ed25519"
+	cryptorand "crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"fmt"
+	"math/big"
 	"math/rand"
+	"net"
 	"net/http"
+	"os"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 )
 
+const (
+	defaultRSABits      = 4096
+	keyAlgorithmRSA     = "rsa"
+	keyAlgorithmEd25519 = "ed25519"
+)
+
+// isFlagSet returns true if the flag was explicitly set in the command line by the user.
 func isFlagSet(name string) bool {
 	found := false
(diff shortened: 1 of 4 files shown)


examples/instrumentation/go-synthetic/auth.go
```diff
@@ -22,6 +41,170 @@ func isFlagSet(name string) bool {
...
+			return nil, fmt.Errorf("unable to create self-signed certificate: %w", err)
+		}
+		certPem := pem.EncodeToMemory(&pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: certBytes,
+		})
+
+		privateKeyBytes, err := x509.MarshalPKCS8PrivateKey(privateKey)
+		if err != nil {
+			log.Println("Unable to marshal private key", err)
```
bwplotka, line 181: That's a bit odd, let's return error with wrap?

bwplotka: same below

bwplotka (changes requested):
LGTM, thanks!

Just error handling is off.

examples/instrumentation/go-synthetic/main.go
```diff
@@ -230,12 +230,24 @@ func main() {
...
-			Addr:    *addr,
-			Handler: mux,
+			Addr:      *addr,
+			Handler:   mux,
+			TLSConfig: tlsConfig,
 		}
 
 		g.Add(func() error {
+			if tlsConfig != nil {
+				fmt.Printf("Starting server on %q with TLS\n", *addr)
```
macxamin, line 247: nit log.Println seems to be used in this file. Should we keep the standard?

examples/instrumentation/go-synthetic/main.go
```diff
@@ -230,12 +230,24 @@ func main() {
...
+				return server.ListenAndServeTLS("", "")
+			}
+			fmt.Printf("Starting server on %q\n", *addr)
```
macxamin, line 250: nit same as above

macxamin (commented):
LGTM

macxamin (approved):
LGTM

bwplotka (approved):
Thanks!

Issue #549, opened 2023-08-17
Operator runs out of ConfigMap space for collectors with ~> 450 PodMonitoring CRs


optix2000:
We're running into an issue where we have a lot of PodMonitoring CR's and the collector is failing to scrape new namespaces.
After some debugging we traced it down to the gmp-operator failing to update the Configmap as the config has gotten too big for it.

Our use-case here is we want to have different scrape intervals between different namespaces as they have different levels of importance, so ClusterPodMonitoring doesn't work for us.

```
% kubectl get podmonitorings.monitoring.googleapis.com -A | wc -l
453
```

Operator logs are full of:
```
[...]

PR #571, opened 2023-09-15
Add flag to compress the Prometheus ConfigMap

TheSpiritXIII:
This PR is a simple short-term solution for #549.

Simply specify:
```bash
kind: OperatorConfig
features:
  config:
    compression: gzip
```

Verified the config-reloader decompresses this config file automagically in-place:
```bash
{"caller":"reloader.go:373","cfg_in":"/prometheus/config/config.yaml","cfg_out":"/prometheus/config_out/config.yaml","level":"info","msg":"Reload triggered","ts":"2023-09-15T20:21:45.101155762Z","watched_dirs":""}
```


cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -50,6 +50,12 @@ spec:
                 enum:
                 - none
                 - gzip
+              configCompression:
```
pintohutch, line 53: IMO this makes more sense as a `.features` rather than in `.collection`.

The reason being that `.collection` fields relate more to ingress/egress of metric data, not the config itself.

The fact that the config is compressed isn't really a concern of the collectors at all actually (at least not as of today).

pintohutch: Another thought I had.

If we want to make this more generalizable and future-proof, where we could provide sharding down the line. Maybe something like:
```yaml
kind: OperatorConfig
metadata:
  namespace: gmp-public
  name: config
features:
  config:
    # Eventually deprecate and have a defaulting webhook omit it.
    compression: gzip
    # Add later when/if config-reloader is obsolete.
    shardCount: 5
```

pintohutch: Un-resolving for thoughts on new API structure.

Though I am happy to bow out here and let @bwplotka take over 😄

bwplotka: Nice. Do we really need option at all? Always gzip is not a bad idea.

Another point is that we will have a harder time to implement https://github.com/prometheus/prometheus/issues/9783#issuecomment-1669341273 if we depend on gzip. But so for any other sharding solution.

TheSpiritXIII:
Thanks! Kudos to @pintohutch for finding the auto-gzip-unzip.

> Do we need extra manual config option? Why not always gzip?
It would be easier to backport this in the future if we don't change defaults. The new feature is protected by a feature flag that customers can optionally enable. We can always change defaults in a future PR.

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
bwplotka: Thanks for the answer on https://github.com/GoogleCloudPlatform/prometheus-engine/pull/571#issuecomment-1725867848

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -93,6 +93,16 @@ spec:
             type: object
             description: Features holds configuration for optional managed-collection features.
             properties:
+              config:
+                type: object
+                description: Configuration of the Prometheus configuration.
```
bwplotka, line 98: Bit weird wording, maybe 
```suggestion
                description: Settings for the collector configuration propagation.
```

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -93,6 +93,16 @@ spec:
...
+                properties:
+                  configCompression:
+                    type: string
+                    description: Compression enables compression of the config data
```
bwplotka, line 102:
```suggestion
                    description: Compression enables compression of the config data propagated by operator to collectors. It's recommended to use gzip option for the large number of Cluster/PodMonitoring CRs.
```

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
```diff
@@ -93,6 +93,16 @@ spec:
             type: object
             description: Features holds configuration for optional managed-collection features.
             properties:
+              config:
+                type: object
+                description: Configuration of the Prometheus configuration.
+                properties:
+                  configCompression:
```
bwplotka, line 100: Perhaps just compression is enough?

```suggestion
                  compression:
```

bwplotka: since we are in .config property.

pkg/operator/collection.go
```diff
@@ -316,9 +330,24 @@ func (r *collectionReconciler) ensureCollectorConfig(ctx context.Context, spec *
 			Namespace: r.opts.OperatorNamespace,
 			Name:      NameCollector,
 		},
-		Data: map[string]string{
+	}
+
+	switch compression {
+	case monitoringv1.CompressionGzip:
```
bwplotka, line 336: Let's add comment like 
```suggestion
        // Thanos config-reloader detects gzip compression automatically, so no sync with config-reloaders is needed when switching between those.
	case monitoringv1.CompressionGzip:
```

bwplotka: as it might be a bit surprising for future readers.

cmd/operator/deploy/crds/monitoring.googleapis.com_operatorconfigs.yaml
pintohutch: We could omit it and have it transparently always-on as well.

Per rollout policy, this could be considered a blanket "performance improvement" as it allows managed collection to scale better.

However, presumably CPU usage of the config-reloaders would increase. Maybe it's worth doing a quick benchmark to see what the cost is for a relatively large configmap?

bwplotka (approved):

bwplotka:
Thanks 💪🏽

TheSpiritXIII:
Hi @shinebayar-g the easiest way to get this change is to switch to the GKE rapid release channel on the latest GKE minor version.

In general, after we tag a release, it takes a couple of months for the change to go from release -> regular -> stable and at each stage we rollout to a small percentage of customers to ensure our changes don't break everyone. We are not allowed to make any progress towards release promotion during US holidays so expect some delays for this release.

shinebayar-g:
We're on Stable channel which has 1.25/1.26. Looks like Rapid channel is at 1.27. I afraid we're not allowed to use Rapid channel in our environment during this time as well.
Underlying [terraform resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#managed_prometheus) has only boolean flag to turn on and off the Prometheus addon. Is there any plan to version these addons? So these changes are not tied to unknown GKE versions.

Is there any other workaround for the moment? I guess if we manually apply the new version it may get reverted back by GKE control plane.

pintohutch:
Hi @shinebayar-g - what version of GKE are you on? It is possible to upgrade to a cluster minor in STABLE to get `v0.8.0`.

For example I was able to upgrade a test cluster to `1.25.15-gke.1115000` via:
```
 gcloud container clusters upgrade my-test-cluster --cluster-version 1.25.15-gke.1115000 --master
```
to get the latest tag.

shinebayar-g:
That's great to hear! I just checked our clusters.
We're on `1.25.10-gke.2700`, `1.25.12-gke.500` respectively. They're all on `components.gke.io/component-version: 0.7.0-gke.0`

pintohutch:
Hey @shinebayar-g - you can see more available versions to you by running
```
gcloud container get-server-config
```

PR #572, opened 2023-09-20
Fix lint warnings in e2e package

TheSpiritXIII:
My original take on this is in #413 but they potentially discovered real bugs so I split out the e2e ones and ran the latest linter.


bwplotka (approved):
Thanks!

PR #573, opened 2023-09-21
Add ability to run individual tests

TheSpiritXIII:
Provides a uniform way of running a single test via our Makefile, no matter what other environment variables are used.


bwplotka (approved):
LGTM, thanks!

Ideally we could leverage IDE with clickables, but we would need to push the setup framework in to unit test (:

PR #575, opened 2023-09-21
Replace deprecated ioutil module usage

TheSpiritXIII:
This is effectively a no-op. The ioutils package today simply call the os packages. This removes some warnings in the codebase.


macxamin (approved):
LGTM

PR #577, opened 2023-09-25
Use subtest labels for cleanup

TheSpiritXIII:
This PR is the first part of multiple for allowing tests to be able to run run isolated from each other. Currently some subtests are broken because they rely on configurations to get created from prior subtests.

Using labels for tests allows us to remove resources in subtests in the future to ensure that all subtests are completely independent. This also helps prevent name collisions in resources and allows tests to be more robust.  The existing tests already used labels for namespaces, but this expands them by making labels more explicit.

### DIFF
--- a/e2e/alertmanager_test.go
+++ b/e2e/alertmanager_test.go
@@ -40,7 +40,10 @@ route:
   receiver: "foobar"
 `
 	secret := &corev1.Secret{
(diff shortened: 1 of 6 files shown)


bwplotka (approved):
Thanks 👍🏽

PR #578, opened 2023-09-25
Cleanup resources between subtests and make subtests isolated

TheSpiritXIII:
Part 2 of #577. This fixes the tests so that each test can be run isolated via #573.

New tests fail if they are not independent, since resources are now cleaned up between subtests.


bwplotka (approved):
Much cleaner, thanks!

PR #579, opened 2023-09-26
style: linter error fixes

bernot-dev:
A few random linter error fixes.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -1,5 +1,3 @@
-// Copyright 2022 Google LLC
-//
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


bernot-dev:
> I don't see anything in our `Makefile` to run this. Do you think you can add a command?

Maybe we should try this: https://github.com/golangci/golangci-lint-action

pkg/operator/operator.go
```diff
@@ -1,5 +1,3 @@
-// Copyright 2022 Google LLC
```
macxamin, line 1: did the tool mean to remove the copyright?

bernot-dev: I think that was probably just me fat fingering the deletion of the line. Good catch. I'll add that back.

PR #582, opened 2023-09-26
Add HTTP middleware to rule-evaluator

macxamin:
This code accomplishes three goals:
1. Remove grpc measurements since grpc is not used by the rule-evaluator.
2. Expose the export metrics collected by export.go. In the future, we may want to measure latencies of CreateTimeSeries.
3. Wrap Query requests sent to GCM API with an instrument RoundTripper to count return codes and measure latencies. This will give GMP more control over their SLO's and Alerts. 

This PR will resolve b/300157911


bwplotka (approved):
Looks solid & simple!

Thanks! 👍🏽

PR #585, opened 2023-09-27
Allow e2e tests to run in parallel

TheSpiritXIII:
Webhooks are preventing e2e tests from running in parallel. This change makes the webhook server choose a random port per test.

Original description:
```
Disable the webhook server by default on e2e tests

This allows us to run e2e tests in parallel after #578 is merged. 🎉 

Disabling it is easier than assigning a new port per test.
```

### DIFF
--- a/e2e/operator_context_test.go
+++ b/e2e/operator_context_test.go
@@ -48,6 +48,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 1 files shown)


bernot-dev:
Why does the webhook server need to be disabled?

TheSpiritXIII:
> Why does the webhook server need to be disabled?

Updated description.

TheSpiritXIII:
Had an offline discussion with @bernot-dev who thinks that this change would affect the fidelity of our integration tests.

TheSpiritXIII:
@bernot-dev please take another look. I figured out how to enable webhook servers to be initialized on random ports.

bernot-dev (approved):
Update PR title/description to reflect the actual change, but LGTM.

PR #586, opened 2023-09-28
Fix bench library error

TheSpiritXIII:
Fixes:
```
pkg/export/bench/server.go:33:6: main redeclared in this block
        pkg/export/bench/example_app.go:143:6: other declaration of main
```


bwplotka (approved):
Sweet, thanks!

PR #587, opened 2023-09-29
Automatically add subtest labels for cleanup on e2e tests

TheSpiritXIII:
Builds on #580 and #577 to make integration tests easier to write.

### DIFF
--- a/e2e/kubeclient.go
+++ b/e2e/kubeclient.go
@@ -1,0 +1,194 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


e2e/kubeclient.go
```diff
@@ -0,0 +1,194 @@
...
+
+func newDelegatingWriteClient(c client.Client, w WriterClient) DelegatingClient {
+	return &delegatingWriteClient{
+		base:   c,
+		writer: w,
+	}
+}
+
+// labelWriterClient adds common labels to all objects mutated by this client.
+type labelWriterClient struct {
```
bernot-dev, line 129: Interesting idea to create a client that writes the labels on every write operation. It does seem a bit heavy-handed, but I think it works for test code. And it prevents the test writer from having to worry about missing any objects they create, which is nice.

bernot-dev (approved):
LGTM

PR #596, opened 2023-10-03
Parallelize webhook tests and replace clientsets

TheSpiritXIII:


bernot-dev (approved):
I should be removing CA bundle injection altogether soon. But this works for now. LGTM.

PR #600, opened 2023-10-03
Add explicit directories for updates

wespanther:
In https://github.com/GoogleCloudPlatform/prometheus-engine/pull/589, I added a dependabot config that just looks at the root directory. This just explicitly lists the directories we want to look at.

Let me know if we want to include '/hack' or not.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,7 +1,23 @@
 version: 2
 updates:
 - package-ecosystem: docker
-  directory: /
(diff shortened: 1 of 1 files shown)


TheSpiritXIII (approved):
`hack` isn't used for production images but let's update everything altogether.

wespanther:
@TheSpiritXIII - Thank you, sounds good! I don't have permission to merge, could you do it for me?

PR #601, opened 2023-10-03
Add ability to pass cert files to operator

bernot-dev:
Currently, the GMP Operator accepts a Certificate for the webhook server
and a CA Cert for clients as arguments on the command line. This is not
the most secure way of passing the server certificate, as noted by
GKE Security (b/297043491#comment2).

This change allows these certificates to be passed as files instead of
command line arguments and prioritizes the files over the command line
args (args will be ignored if files exist).

Expected usage is that a Kubernetes Secret (`webhook-cert`) will be created with the server cert and
a ConfigMap (`webhook-ca`) will be created with the Client CA, which can then be
mounted to the operator as a projected volume at the default path of
`/etc/tls/private/`, using standard filenames `tls.crt`, `tls.key`, and
`ca.crt`, respectively.

The paths can be customized using the new flags, if desired. These flags
to the operator are `--cert-file`, `--key-file`, and `--client-ca-file`,
respectively.

The `--tls-cert-base64`, `--tls-key-base64`, and `--ca-cert--base64`
flags should be considered deprecated and are planned to be removed in
the future.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -38,6 +38,13 @@ import (
 	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator"
 )
 
+const (
+	defaultTLSDir       = "/etc/tls/private"
+	defaultCertFile     = defaultTLSDir + "/tls.crt"
+	defaultKeyFile      = defaultTLSDir + "/tls.key"
+	defaultClientCAFile = defaultTLSDir + "/ca.crt"
+)
+
 func unstableFlagHelp(help string) string {
 	return help + " (Setting this flag voids any guarantees of proper behavior of the operator.)"
 }
@@ -68,10 +75,13 @@ func main() {
 		publicNamespace = flag.String("public-namespace", operator.DefaultPublicNamespace,
 			"Namespace in which the operator reads user-provided resources.")
 
-		tlsCert     = flag.String("tls-cert-base64", "", "The base64-encoded TLS certificate.")
-		tlsKey      = flag.String("tls-key-base64", "", "The base64-encoded TLS key.")
-		caCert      = flag.String("ca-cert-base64", "", "The base64-encoded certificate authority.")
-		webhookAddr = flag.String("webhook-addr", ":10250",
+		tlsCert      = flag.String("tls-cert-base64", "", "The base64-encoded TLS certificate.")
+		tlsKey       = flag.String("tls-key-base64", "", "The base64-encoded TLS key.")
+		caCert       = flag.String("ca-cert-base64", "", "The base64-encoded certificate authority.")
+		certFile     = flag.String("cert-file", defaultCertFile, "Path to TLS certificate for webhook server.")
+		keyFile      = flag.String("key-file", defaultKeyFile, "Path to TLS key for webhook server.")
+		clientCAFile = flag.String("client-ca-file", defaultClientCAFile, "Client CA certificate to trust the webhook server.")
+		webhookAddr  = flag.String("webhook-addr", ":10250",
 			"Address to listen to for incoming kube admission webhook connections.")
 		metricsAddr = flag.String("metrics-addr", ":18080", "Address to emit metrics on.")
 
@@ -110,6 +120,9 @@ func main() {
 		TLSCert:           *tlsCert,
 		TLSKey:            *tlsKey,
 		CACert:            *caCert,
+		KeyFile:           *keyFile,
+		CertFile:          *certFile,
+		ClientCAFile:      *clientCAFile,
 		ListenAddr:        *webhookAddr,
 		CleanupAnnotKey:   *cleanupAnnotKey,
 	})
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -84,6 +84,9 @@ const (
 
 	// The level of concurrency to use to fetch all targets.
 	defaultTargetPollConcurrency = 4
+
+	// certDir is the directory where TLS certificates are stored
+	certDir = "/etc/tls/private"
 )
 
 // Operator to implement managed collection for Google Prometheus Engine.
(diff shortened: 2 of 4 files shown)


pkg/operator/operator.go
```diff
@@ -495,6 +550,14 @@ func (o *Operator) ensureCerts(ctx context.Context, dir string) ([]byte, error)
 	return caData, nil
 }
 
+func fileExists(f string) bool {
+	_, err := os.Stat(f)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
```
TheSpiritXIII, line 558: Is it true that the file exists when there's an error? Have you considered return `bool, error` that way you could early exit on the operator startup?

pkg/operator/operator.go
```diff
@@ -256,7 +259,56 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
...
+						}
+					} else {
+						logger.Error(err, "unable to read client CA for webhook server, skipping")
+						return
+					}
+				}
+				ca, err = x509.ParseCertificate(rawCA)
+				if err != nil {
+					logger.Error(err, "unable to parse client CA for webhook server, skipping")
+					return
```
TheSpiritXIII, line 282: Is it okay that we skipped? We could do error handling outside this method and just pass in the certificate here.

pkg/operator/operator.go
```diff
@@ -256,7 +259,56 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
...
+						cert, tlsErr = tls.X509KeyPair(tlsCert, tlsKey)
+						if tlsErr != nil {
+							logger.Error(tlsErr, "invalid cert or key provided")
+						}
+					} else {
+						logger.Error(err, "unable to load TLS certificate/key pair for webhook server, skipping")
+						return
+					}
+				}
+				c.Certificates = []tls.Certificate{cert}
```
TheSpiritXIII, line 309: Should we maybe append here?

TheSpiritXIII (commented):
Great start! Please answer my questions and I'll re-review and approve.

TheSpiritXIII (changes requested):
(still learning how to use GitHub's UI)

pintohutch:
Can you elaborate on the motivation and use-cases for this in the PR or commit description?

And can you add some instructions on how to use these new flags?

Specifically, how are files passed to the operator? Does it require changing the manifest?

pkg/operator/operator.go
bernot-dev: I'll change this to take any error as a false for fileExists because I think the other possible errors would indicate the file isn't going to be readable.

pkg/operator/operator.go
bernot-dev: Good idea. That ends up being a bit cleaner, I think.

pkg/operator/operator.go
bernot-dev: Shouldn't matter in this case, but I'll make the change.

bernot-dev:
@pintohutch I added details in the description and commit message.

cmd/operator/main.go
```diff
@@ -68,10 +75,13 @@ func main() {
...
 			"Namespace in which the operator reads user-provided resources.")
 
-		tlsCert     = flag.String("tls-cert-base64", "", "The base64-encoded TLS certificate.")
-		tlsKey      = flag.String("tls-key-base64", "", "The base64-encoded TLS key.")
-		caCert      = flag.String("ca-cert-base64", "", "The base64-encoded certificate authority.")
-		webhookAddr = flag.String("webhook-addr", ":10250",
+		tlsCert      = flag.String("tls-cert-base64", "", "The base64-encoded TLS certificate.")
+		tlsKey       = flag.String("tls-key-base64", "", "The base64-encoded TLS key.")
+		caCert       = flag.String("ca-cert-base64", "", "The base64-encoded certificate authority.")
+		certFile     = flag.String("cert-file", defaultTLSDir+"tls.crt", "Path to TLS certificate for webhook server.")
```
TheSpiritXIII, line 81: What if we have non-webhook server TLS certificates in the future. Maybe we should be more detailed here with the flag name.

pkg/operator/operator.go
```diff
@@ -189,6 +192,46 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
...
+	if fileExists(opts.ClientCAFile) {
+		rawCA, err := os.ReadFile(opts.ClientCAFile)
+		if err != nil {
+			return nil, fmt.Errorf("read CA file: %w", err)
+		}
+		ca, err = x509.ParseCertificate(rawCA)
+		if err != nil {
+			return nil, fmt.Errorf("parse CA file: %w", err)
+		}
+	} else if opts.CACert != "" {
```
TheSpiritXIII, line 206: Does it make sense for the user to provide both flags if they're mutually exclusive? Do we not want to show an error or a warning if they put both since 1 is effectively useless? Or will deprecate the inline certs in the near future?

TheSpiritXIII (approved):
I left a couple small minor nits in comments but I am approving! Thank you!

pkg/operator/operator.go
bernot-dev: Good idea. Yeah, I will log a warning.

But yes, the flags for inline certs should disappear soon after the necessary changes are submitted/released on the GKE side.

pintohutch:
Thanks @bernot-dev.

What happens if Secret `webhook-tls` or ConfigMap `webhook-ca` don't exist?

bernot-dev:
> What happens if Secret `webhook-tls` or ConfigMap `webhook-ca` don't exist?

Then the operator won't find them. They are marked as optional, so they won't prevent the operator from starting. Because it still is accepting the literal args as well, for now.

PR #602, opened 2023-10-04
Add Kubernetes clients which tracks objects created in e2e tests

TheSpiritXIII:
Tracked objects are used for cleanup. This dramatic improves e2e test performance while ensuring conformity of tests.

On my machine before this change, e2e took 4 minutes. After this change, it only takes 7 seconds.


pintohutch:
Can you explain more about what this change is doing? And why the speed up?

IIUC before, we deleted a transient test namespace, which maybe took more time than caching created objects and deleting them explicitly. But that may be incorrect.

Can be here in or in-code.

TheSpiritXIII:
@pintohutch The problem with subtests before is that they were not isolated.

For example let's say you're writing or updating a subtest. In Go, you can specify which subtest to run and not run any of the others. You may run one such as `TestCollector.self-pod-monitoring`. One issue here, for example, might be that the `PodMonitoring` for this subtest is created in a preceding subtest. If you only run this one subtest and not the subtest(s) prior, you don't have all the data so your subtest will fail. Hence, this subtest is not what I would call isolated.

To resolve the issue of dependent subtests, I created a series of PRs that used labels instead of owner references #578. Then, I created a PR that adds these labels automatically #587. All of this came with new cleanup code that was run in between each subtest.

Although I added parallelization to tests, that cleanup code is slow. It finds all Kubernetes resources and deletes all resources with the subtest label. This easily becomes hundreds of delete calls, because it has to be repeated for each namespace. Hence why the sudden performance increase -- hundreds of blocking delete calls becomes just a dozen. Originally, I tried parallelizing the delete calls but Kubernetes starts rate-limiting them so I arrived at this solution.

TheSpiritXIII:
Going to close this. After #630, this change does not bring much value, since we now leak resources by default.

PR #604, opened 2023-10-04
Add target status integration tests

TheSpiritXIII:
Slightly different approach from #395. The problem is that when we run an operator within the test, the HTTP client cannot resolve Pod IP addresses. The resolve this, I created a custom HTTP client which can do this.

This PR builds on top of #613 since we now check if collectors are ready before checking whether target status is working.


pintohutch:
cc @bwplotka

e2e/network.go
```diff
@@ -0,0 +1,197 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package e2e contains tests that validate the behavior of gmp-operator against a cluster.
+// To make tests simple and fast, the test suite runs the operator internally. The CRDs
+// are expected to be installed out of band (along with the operator deployment itself in
+// a real world setup).
+package e2e
```
bwplotka, line 19: wowow - this is quite next level and quite innovating. Probably good for fun blog post / separate OSS lib / contribution to k8s testing libs or so 🙈

bwplotka (approved):
Looks solid, thanks for this work! 💪🏽 A bit a lot of low level code to maintain, but well readable & for good reasons, thanks!

PR #613, opened 2023-10-06
Allow Prometheus to run GitHub Actions in e2e tests

TheSpiritXIII:
Our current e2e tests are not deploying the Prometheus collector pods correctly on GitHub Actions due to 2 reasons:

1. GitHub Actions does not support "localhost". It only supports the loopback address directory.
2. GitHub Actions does not have Google Cloud credentials.

These issues cause the collector pods to crash loop indefinitely in e2e tests. cc @pintohutch

### DIFF
--- a/cmd/operator/deploy/operator/10-collector.yaml
+++ b/cmd/operator/deploy/operator/10-collector.yaml
@@ -57,8 +57,8 @@ spec:
         args:
         - --config-file=/prometheus/config/config.yaml
         - --config-file-output=/prometheus/config_out/config.yaml
-        - --reload-url=http://localhost:19090/-/reload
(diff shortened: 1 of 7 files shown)


pintohutch:
@TheSpiritXIII did you ever figure out what the issue was with the metricsClient auth disabling issue?

TheSpiritXIII:
> @TheSpiritXIII did you ever figure out what the issue was with the metricsClient auth disabling issue?

Yes -- it was my fault. The arg wasn't originally taking because I was modifying the value in the for loop, which was being copied and not actually applied to the original object. In this PR, the for loop explicitly takes a reference to the container object to prevent that issue.

pintohutch (approved):

PR #615, opened 2023-10-08
Move proxy-related fields to proxy config object

TheSpiritXIII:
This matches the Prometheus data structure: 
- https://github.com/prometheus/common/blob/7043ea0e691b6da9ecbd08e7ae41e9cf28898e98/config/http_config.go#L1196

This is then used in the root HTTP Client and OAuth2 configs in Prometheus:
- https://github.com/prometheus/common/blob/7043ea0e691b6da9ecbd08e7ae41e9cf28898e98/config/http_config.go#L226
- https://github.com/prometheus/common/blob/7043ea0e691b6da9ecbd08e7ae41e9cf28898e98/config/http_config.go#L289


doc/api.md
```diff
@@ -104,7 +105,7 @@ Authorization specifies a subset of the Authorization struct, that is safe for u
 | Field | Description | Scheme | Required |
 | ----- | ----------- | ------ | -------- |
 | type | Set the authentication type. Defaults to Bearer, Basic will cause an error | string | false |
-| credentials | The secret's key that contains the credentials of the request | *[v1.SecretKeySelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#secretkeyselector-v1-core) | false |
+| credentials | The secret's key that contains the credentials of the request | *corev1.SecretKeySelector | false |
```
bernot-dev, line 108: Interesting that the link got dropped. Core is just an alias...I wonder why that got carried forward here.

pkg/operator/apis/monitoring/v1/http.go
```diff
@@ -37,3 +38,41 @@ func (c *TLS) ToPrometheusConfig() (*config.TLSConfig, error) {
 		MaxVersion:         maxVersion,
 	}, nil
 }
+
+func (c *ProxyConfig) ToPrometheusConfig() (config.URL, error) {
+	proxyURL, err := url.Parse(c.ProxyURL)
+	if err != nil {
+		return config.URL{URL: nil}, fmt.Errorf("invalid proxy URL: %w", err)
```
bernot-dev, line 45: Any reason to set `URL: nil` here? Should just be an empty literal, right?

bernot-dev (approved):
Non-blocking comments. LGTM.

doc/api.md
TheSpiritXIII: Huh, probably a bug in the underlying doc generator, which seems like it has been deleted and no longer maintained.

PR #616, opened 2023-10-09
Run make regen.

bwplotka:
Something went off - build fails and it's not clear what's exactly not regenerated (https://github.com/GoogleCloudPlatform/prometheus-engine/actions/runs/6458765809/job/17533143057?pr=612), but the `make regen` indicates this vendor code. 

Sounds like dependanbot messed up. 🤔


bwplotka:
I though it's because vendor is false, but it should be still fine: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#vendor

TheSpiritXIII (approved):

PR #621, opened 2023-10-10
Detect running GKE cluster on e2e tests

TheSpiritXIII:
Quality of life improvement: This is something I wanted while trying to running the tests in my debugger. This change lets you just start your debugger (or even the test directly) without tinkering with any CLI args.


bwplotka (approved):
👍🏽

PR #622, opened 2023-10-10
add initial support for alertmanager external-url

### LINKED ISSUE #352: no way to set `web.external-url` on Managed Alertmanager
Hey there! Thanks very much for this project, it's been very simple to set up.

With self-managed Prometheus/Alertmanager, regardless of the helm chart / operator / managed process approach, there's usually a way to set the [`-web.external-url` runtime flag](https://github.com/prometheus/alertmanager/blob/28c3df01739b3e8b5e33004a776db53fd576e7b8/cmd/alertmanager/main.go#L205).  I'm having some challenges [finding a way to inject this with the GMP operator](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/manifests/operator.yaml#L906-L911).

Without being able to set this backlink, our Alertmanager notifications won't contain some way for our responders to link back to the AM alert itself, since like many, we're running Alertmanager from inside of a private k8s cluster

pintohutch:
Resolves #352


pintohutch:
Just making sure the test fails properly before fixing.

pintohutch:
Hey @bwplotka PTAL

pintohutch:
cc @bernot-dev as this may be relevant to default-off alertmanager/rule-evaluator

e2e/alertmanager_test.go
```diff
@@ -108,15 +110,15 @@ func testAlertmanager(ctx context.Context, t *OperatorContext, spec *monitoringv
...
+		testAlertmanagerDeployed(ctx, t, spec)
 	}))
 	t.Run("config set", t.subtest(func(ctx context.Context, t *OperatorContext) {
 		t.Parallel()
 		testAlertmanagerConfig(ctx, t, secret, key)
 	}))
 }
 
-func testAlertmanagerDeployed(ctx context.Context, t *OperatorContext) {
+func testAlertmanagerDeployed(ctx context.Context, t *OperatorContext, spec *monitoringv1.ManagedAlertmanagerSpec) {
```
bwplotka, line 121:
```suggestion
func testAlertmanagerDeployed(ctx context.Context, t *OperatorContext, config *monitoringv1.ManagedAlertmanagerSpec) {
```

bwplotka: Super small nit, but I was quite confused thinking initially this variable as used down there is Alertmanager spec not alertmanager configuration spec (kind of) - a bit more explicitness might help

e2e/alertmanager_test.go
```diff
@@ -135,6 +137,27 @@ func testAlertmanagerDeployed(ctx context.Context, t *OperatorContext) {
...
+			return true, nil
+		}
+		var wantArgs []string
+		for _, c := range ss.Spec.Template.Spec.Containers {
+			if c.Name != "alertmanager" {
+				continue
+			}
+			// We're mainly interested in the dynamic flags but checking the entire set including
+			// the static ones is ultimately simpler.
+			if externalURL := spec.ExternalURL; externalURL != "" {
```
bwplotka, line 151: Bit confused again - can't we prepare wantArgs before looping through Alertmanager statetful set containers? It feel like you do it inside this loop for a reason, but I can't find any 🤔

manifests/setup.yaml
```diff
@@ -986,6 +986,9 @@ spec:
                 required:
                 - key
                 x-kubernetes-map-type: atomic
+              externalURL:
+                type: string
+                description: ExternalURL is the external URL the managed Alertmanager will be available under. This is used for generating links back to the Alertmanager itself in fired alerts.
```
bwplotka, line 991: I wonder if it wouldn't be easier to put exact flag description. It's essentially more than just alert link, it also fixes UI when reverse proxies are used. WDYT?

```suggestion
                description: The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.
```

bwplotka (approved):
Epic, thanks! Some nits, otherwise LGTM. Sorry for review lag.

pintohutch (commented):
Thanks for the feedback! Pushed up some fixes

PR #627, opened 2023-10-11
Fix e2e tests cannot run locally after webhook cert dir changes

TheSpiritXIII:
#601 introduced a small regression where the operator fails when running e2e tests locally. The reason is because the original code always created a temp dir while the new code creates a dir at a static location (/etc/tls) which fails on my system. Even if it passed, it would not work if every e2e uses the same directory.

As such, the e2e tests now create and manage a temp dir on their own.

Note, I accidentally pushed this commit to the main branch earlier which I reverted so we could code review this properly.

### DIFF
--- a/e2e/operator_context_test.go
+++ b/e2e/operator_context_test.go
@@ -236,6 +236,16 @@ func newOperatorContext(t *testing.T) *OperatorContext {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


bernot-dev:
I think there will need to be some additional logic to use the "default" directory and maybe fall back to the temp directory.

TheSpiritXIII:
Deleting, as we are in progress of fixing this.

PR #630, opened 2023-10-16
Add options to leak and cleanup resources

TheSpiritXIII:
Provides the ability to leak and cleanup resources for an improved debugging experience, especially when running tests locally.

### DIFF
--- a/e2e/operator_context_test.go
+++ b/e2e/operator_context_test.go
@@ -75,6 +75,8 @@ var (
 	skipGCM           bool
 	gcpServiceAccount string
 	portForward       bool
+	leakResources     bool
+	cleanup           bool
 )
 
 func init() {
@@ -121,6 +123,8 @@ func TestMain(m *testing.M) {
 	flag.BoolVar(&skipGCM, "skip-gcm", false, "Skip validating GCM ingested points.")
 	flag.StringVar(&gcpServiceAccount, "gcp-service-account", "", "Path to GCP service account file for usage by deployed containers.")
 	flag.BoolVar(&portForward, "port-forward", true, "Whether to port-forward Kubernetes HTTP requests.")
+	flag.BoolVar(&leakResources, "leak-resources", false, "If set, prevents deleting resources. Useful for debugging.")
+	flag.BoolVar(&cleanup, "cleanup-resources", false, "If set, cleans resources before running tests.")
 
 	flag.Parse()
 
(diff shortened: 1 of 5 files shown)


bernot-dev:
The pattern I typically see is to "leak" by default, and then cleanup by default before the test starts. This can prevent the case where someone leaks resources that impact subsequent tests.

TheSpiritXIII:
Hi @bernot-dev can you share some examples where you see this convention?

My reasoning for not leaking by default are:

- It doesn't change behavior of tests today.
- It doesn't mess up your cluster if you run it locally by default.
- It makes tests significantly slower. When we clean after a test, we know what to clean (at least after we merge #602). If we clean before each test run, we don't know what to clean. This translates to hundreds of blocking API calls. We have to lookup all Kubernetes API groups, and then individually clean each group. Note, we cannot parallelize these because the Kubernetes API would only throttle us.

bernot-dev:
> * It doesn't mess up your cluster if you run it locally by default.

I guess it's still a little strange to me that E2E tests wouldn't be running in a cluster dedicated to testing. But yes, it does speed things up. Probably a good tradeoff for us.

> * Tests would fail since they expect resources from the prior one to get cleaned up. The `leak` flag would only be useful for subtests right now.

True, if the cleanup didn't happen before running tests.

> * It makes tests significantly slower. When we clean after a test, we know what to clean (at least after we merge [Add Kubernetes clients which tracks objects created in e2e tests #602](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/602)). If we clean before each test run, we don't know what to clean. This translates to hundreds of blocking API calls. We have to lookup all Kubernetes API groups, and then individually clean each group. Note, we cannot parallelize these because the Kubernetes API would only throttle us.

Makes sense. I would think hundreds of queries would be on the order of seconds, though, as long as they don't have to deal with finalizers and such.

TheSpiritXIII:
@bernot-dev please take another look! I flipped the defaults of the flags, as requested. I got around the performance issue by simply defaulting turning the cleanup flag off when running under Docker (which implicitly does cleanup by recreating a new `kind` cluster).

e2e/operator_context_test.go
```diff
@@ -66,7 +67,7 @@ const (
 
 var (
 	startTime    = time.Now().UTC()
-	globalLogger = zap.New(zap.Level(zapcore.Level(-1)))
+	globalLogger = zap.New(zap.Level(zapcore.DebugLevel))
```
bernot-dev, line 70: 👍

bernot-dev (approved):
LGTM

PR #634, opened 2023-10-17
export: Added "promtest" framework allowing tests against GCM and Prometheus; Added counter test.

bwplotka:
This PR:
* Adds `pkg/export/gcm/promtest` pkg that contains "IngestionTest" framework.
* (The only change to export pkg) pkg/export: Adding a config option that allows passing credentials content instead of files.
* Adds `pkg/export/gcm/export_gcm_test.go` that implements a simple counter test that showcases the framework use.

This uses docker the e2e framework [I maintain with some Thanos maintainers](https://github.com/efficientgo/e2e) for Prometheus test and local export code for "GMP" test. The "local export" is so we can use local debugger and test EVERY export change in this repo.

Tests with Prometheus enabled, takes N samples * 5 seconds. We could take scrape interval to 1s, but it might kill our CI CPU. We can also disable Prometheus tests for CI and run it only locally.

**Those tests will also NOT run for contributors as it requires secret.**

In future PRs we can:
* Move to Cloud build for stronger security guarantees for our test SA.
* Add test for new logic for resets.
* Add GMP + GCM test (docker container too)
* Add tests for other types, different queries (e.g. rates) etc - sky is the limit.


bwplotka:
Tests passed without mention about skipping anything, but I see it ran within 3s which is not realistic so assuming it was skipped - investigating.

bwplotka:
All should work, PTAL @TheSpiritXIII @pintohutch

.gitignore
```diff
@@ -1,3 +1,4 @@
 /build
 /**/node_modules/.cache
-.idea/
\ No newline at end of file
+e2e_*
```
TheSpiritXIII, line 3: Where is this used?

Makefile
```diff
@@ -120,14 +120,24 @@ test:        ## Run unit tests. Setting NO_DOCKER=1 runs test on host machine.
...
 	# TODO(TheSpiritXIII): Temporary env variables part of `export.go` unit tests.
 	$(call docker_build, -f ./hack/Dockerfile --target sync -o . -t gmp/hermetic \
 		--build-arg RUNCMD='GIT_TAG="$(shell git describe --tags --abbrev=0)" TEST_TAG=true ./hack/presubmit.sh test' .)
 	rm -rf vendor.tmp
 endif
 
+GCM_SECRET?=''
+.PHONY: test-export-gcm
+test-export-gcm:  ## Run export unit tests that will use GCM if GCM_SECRET is present.
+                  ## TODO(bwplotka): Move to cloud build.
```
TheSpiritXIII, line 134: Can you ensure you have a bug/issue for this TODO? Typically in Google we also add this link to the TODO. See go/todo-style

Makefile
```diff
@@ -120,14 +120,24 @@ test:        ## Run unit tests. Setting NO_DOCKER=1 runs test on host machine.
...
+ifneq ($(GCM_SECRET), '')
+	TEST_TAG=false go test -v ./pkg/export/gcm
```
TheSpiritXIII, line 136: I may be wrong but false is already the default. Maybe we can remove that env var here?

pkg/export/gcm/promtest/promtest.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/timestamp"
+	"github.com/thanos-io/thanos/pkg/runutil"
+)
+
+// GCMServiceAccountOrFail gets the Google SA JSON content from GCM_SECRET
+// environment variable or fails.
+func GCMServiceAccountOrFail(t testing.TB) []byte {
+	// TODO(bwplotka): Move it to https://cloud.google.com/build CI.
+	saJSON := []byte(os.Getenv("GCM_SECRET"))
```
TheSpiritXIII, line 30: Would it maybe be cleaner and more generic to use flags?

pkg/export/gcm/promtest/promtest.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	it := &ingestionTest{
+		t: t,
+		// Use ULID as a unique label per test run.
+		// NOTE(bwplotka): Cardinality is obvious, but even for 100 test runs a day with
+		// 100 cases, 10 samples each, that's "only" 10k series / 0.1 million samples a day.
+		testID: ulid.MustNew(ulid.Now(), rand.New(rand.NewSource(time.Now().UnixNano()))).String(),
```
TheSpiritXIII, line 85: Might be easier for debugging purposes to encode the test name here? The `e2e` tests do that to create the namespace for the test.

pkg/export/gcm/promtest/promtest.go
```diff
@@ -0,0 +1,316 @@
...
+	it       *ingestionTest
+	currTime time.Time
+}
+
+func (ir *ingestionTestExpRecorder) Expect(val float64, metric prometheus.Metric, b Backend) ExpectationsRecorder {
+	m := dto.Metric{}
+	metric.Write(&m)
+	if m.GetSummary() != nil || m.GetHistogram() != nil {
+		// TODO(bwplotka): Implement an alternative.
+		ir.it.t.Fatal("It's not practical to use equalsGCMPromQuery against histograms and summaries.")
```
TheSpiritXIII, line 177: What is `equalsGCMPromQuery`? Was that a function you had here at one point?

pkg/export/gcm/promtest/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func Prometheus(image string) Backend {
+	return &promBackend{image: image}
+}
+
+func (p *promBackend) Ref() string { return "prometheus" }
+
+// recordGatherer is a prometheus.Gatherer capable to "play" the recorded metric state
+// with fixed timestamp to the backfill data into Prometheus compatible system.
```
TheSpiritXIII, line 39: Very cool!

pkg/export/gcm/promtest/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (p *promBackend) injectScrapes(t testing.TB, scrapeRecordings [][]*dto.MetricFamily, timeout time.Duration) {
+	t.Helper()
+
+	p.g.mu.Lock()
+	p.g.i = 0
+	p.g.plannedScrapes = scrapeRecordings
+	p.g.mu.Unlock()
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
TheSpiritXIII, line 164: Why not just pass the ctx into the function args?

pkg/export/gcm/promtest/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+		t.Fatalf("create Prometheus client: %s", err)
+	}
+
+	return v1.NewAPI(cl), map[string]string{"job": "test"}
+}
+
+func newPrometheus(env e2e.Environment, name string, image string, scrapeTargetAddress string, flagOverride map[string]string) *e2emon.Prometheus {
+
+	if image == "" {
+		image = "quay.io/prometheus/prometheus:v2.47.2"
```
TheSpiritXIII, line 101: Would it maybe make sense to make this a flag or env var, that way we can run the test against different Prometheus versions without changing code? Might be useful for quick comparisons.

pkg/export/gcm/export_gcm_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"time"
+
+	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/export/gcm/promtest"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func TestExport_CounterReset(t *testing.T) {
+	const interval = 30 * time.Second
+
+	prom := promtest.Prometheus("quay.io/prometheus/prometheus:v2.47.2")
```
TheSpiritXIII, line 14: Since this method has a default Prometheus image, why not use that to ensure this test and future tests use the same version?

pkg/export/gcm/export_gcm_test.go
```diff
@@ -0,0 +1,100 @@
...
+		c.Add(50)
+		scrape(interval).
+			Expect(60, c, export).
+			Expect(260, c, prom)
+
+		// Reset to 0, then add something that still should indicate "decreased counter"
+		// in absolute values, but our current cannibalization logic is wrong here.
+		// It's perhaps unusual that reset happens without scrape target restart
+		// so, it was easy to forget about this case.
+		counter.Reset()
```
TheSpiritXIII, line 53: Great find on the bug!

It doesn't look like you have a case that just resets without adding. Would it make sense to add one?

.gitignore
bwplotka: https://github.com/efficientgo/e2e creates tmp local e2e_<id> dir for every test run.

pkg/export/gcm/promtest/promtest.go
bwplotka: I like that idea, will add 👍🏽

pkg/export/gcm/promtest/promtest.go
bwplotka: Hm what's the benefit? What being generic here means/gives us? 

I made it an env var as flags in tests are very rare in OSS e.g. thanks of your comments I know it's possible (found [blog posts](https://blog.jbowen.dev/2019/08/using-go-flags-in-tests/) about it). 🙈 

I guess flags are easier to discover perhaps? But not sure how this applies here. Happy to learn about benefits of flags in tests here, so far it feels it does not matter and envvar is more consistent (we store secret in envvar too)

pkg/export/gcm/promtest/promtest.go
bwplotka: yea

Makefile
bwplotka: Added, but I would be careful to not add too many tickets e.g. for other TODOs in this PR 🙈

pkg/export/gcm/promtest/prometheus.go
bwplotka: We could, but that would just move those two lines to the caller - more work for caller essentially. What's the benefit?  (:

pkg/export/gcm/promtest/prometheus.go
bwplotka: Nice! Yea we can do that, but perhaps we could do that in the PR/event where somebody actually would like to do such test?

pkg/export/gcm/export_gcm_test.go
bwplotka: Hm, I don't get you argument. Why not using default would make future tests use different version? 🤔 

I think I will remove default (YAGNI), we don't need implicit versions here.

pkg/export/gcm/export_gcm_test.go
bwplotka: 👍🏽 

Any feedback on test clarity?

Would it be easier to read if we have scenario of scrapes/metrics and expect certain samples PER scrape at the end (instead per scrape)? 🤗

bwplotka:
Address all comments so far @TheSpiritXIII  (:

.gitignore
TheSpiritXIII: Hm, why local? Why not in the OS tmp dir?

Makefile
TheSpiritXIII: Thanks for looking into this!

pkg/export/gcm/export_gcm_test.go
TheSpiritXIII: If it's a static version in each test, I fear they may go out of sync.

pkg/export/gcm/promtest/prometheus.go
TheSpiritXIII: SGTM!

pkg/export/gcm/promtest/prometheus.go
TheSpiritXIII: Just provides more flexibility. I noticed that the k8s client libraries used to do this (pass timeout in arg) but they deprecated those methods and switched to just using ctx (e.g. see their wait package). I guess we can add it if we need that function. Thanks!

pkg/export/export.go
```diff
@@ -936,6 +936,7 @@ func (b batch) send(
 
 			samplesPerRPCBatch.Observe(float64(len(l)))
 
+			fmt.Println("Send", l)
```
TheSpiritXIII, line 939: Leftover debugging, unless you may want to use a logger here?

pkg/export/gcm/export_gcm_test.go
TheSpiritXIII: I like the granularity you have right now! I feel like it's more low-level so we can always create a utility that wraps this one and does checking at the end.

pkg/export/gcm/promtest/promtest.go
TheSpiritXIII: Just thought it would be discoverable and cleaner to pass into the code, but you're also right that it doesn't bring much benefit. We use flags in our current `e2e` tests. Maybe that's more okay there because the flags are global. Thanks!

TheSpiritXIII:
@bwplotka I'm going to approve. Only nit is the `Println`. Please address and then you can merge. 🙂👍

TheSpiritXIII (approved):

.gitignore
bwplotka: That's the improvement on https://github.com/efficientgo/e2e if we want that. Was easier with perms and sharing with docker, but maybe just laziness (:

pkg/export/gcm/promtest/prometheus.go
bwplotka: Yup, luckly we don't need to worry about compatibility here (:

pkg/export/export.go
bwplotka: Ups!

pintohutch:
This is great! We'd love a demo walkthrough of the e2e package at some point 😄 

In general, this PR was a little intimidating in size. Can we try to put up smaller diffs for review?

WDYT about daisy-chaining baby PRs so we don't hinder development while lowering the review burden?

bwplotka:
Thanks @TheSpiritXIII for review and @pintohutch for feedback.

This could be split a bit yea (e.g I could split local export AND e2e Prometheus into two), let's try that next time! The beauty of one atomic PR is that reviewer understands clearly why this complexity and e2e dependency is needed. There is also question PR author time, given chaining PRs in on GitHub is non trivial - definitely it would be good idea to document how we want to do it as a team 🤗

PR #636, opened 2023-10-18
gcm/promtest: Added gmp collector.

bwplotka:
Child PR: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/634

### DIFF
--- a/pkg/export/gcm/promtest/gmp.go
+++ b/pkg/export/gcm/promtest/gmp.go
@@ -1,0 +1,189 @@
+package promtest
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+	"time"
+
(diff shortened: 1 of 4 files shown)


bwplotka:
NOTE: Chained PRs sucks - no tests are ran for them ):

TheSpiritXIII:
Do we want to maintain both collector and export? It sounds like the collector makes the export a bit redundant.

bwplotka:
I am not sure you get the benefit of export yet (: Export gives ability to use local/IDE debugger, 1s-one-click feedback loop when adding new feature/changing export code, and testing export code on every PR CI run.

It's probably overkill to run both in each test, one is enough generally, but for some tests we could run GMP collector as an integration test of Prometheus and export. Perhaps something to add to our fork tests (importing promtest from here)?

bwplotka:
Closing this one, it will be introduced by PRW test suite.

PR #640, opened 2023-10-18
Make e2e network client more resilient to errors

TheSpiritXIII:
Fixes the issue where checking the target status refuses to work if an unrelated container (i.e. config-reloader) fails to start but the collector is fine.


bwplotka (approved):
👍🏽

PR #641, opened 2023-10-18
Add and parallelize target-status ClusterPodMonitoring e2e tests

TheSpiritXIII:
Additionally this change:

- Creates generic utilities for checking if a PodMonitoring is started. This will be used in future PRs.
- Add labels to collectors so that self-scraping tests don't start looking at collectors outside of the current test.


bwplotka (approved):
LGTM, very clear code now, thanks!

Issue #241, opened 2022-06-08
Feature request: Support Basic Authentication in PodMonitoring


zoeponta:
I want to use basicAuth in metrics endpoints of PodMonitoring, but GMP does not support it .

It is already implemented in [GMP's AlertmanagerEndpoint](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/doc/api.md#alertmanagerendpoints) and the [Prometheus Operator's ServiceMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#basicauth).
So, could you please support same feature in PodMonitoring.

Thanks

pintohutch:
Hi @zoeponta,

We agree and are working on authentication support in PodMonitoring. We can update here when this is addressed.

Thanks for your interest.

akettmann-e24:
Just chiming in that I also would love to have the ability to even just specify headers that the request passes along.

Vaitheesh-SS:
Hi,

Does this enhance is in place for Basic Auth on prometheus metrics endpoints? It is much needed now, since made GCP made managed prometheus will be on by default in new GKE standard clusters.

marandalucas:
We are affected by this.

We can not scrape Prometheus metrics in GCP using Managed Prometheus in our GKE cluster.

We are trying to scrape metrics from MongoDB following the following example:
https://github.com/mongodb/mongodb-kubernetes-operator/blob/93c86675b851e7d44d5bfbd65375602dd4b3ad49/docs/prometheus/mongodb-prometheus-sample.yaml#L59-L104

In this scenario, we can not use "authentication", because unfortunately is not allowed in "PodMonitoring or ClusterPodMonitoring" ->
https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.5.0/doc/api.md#scrapeendpoint

pintohutch:
Hi @marandalucas,

Thanks for raising and expressing your interest in this feature. You are correct that we indeed do not support Basic Auth...yet.

We are still scoping out this feature and are very much interested in getting it into your hands! For the time being, you can still leverage GMP in a "self-deployed" model to use the ServiceMonitor verbatim in prometheus-operator. Please see our [docs](https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-unmanaged#prom-managers) for how to set that up.

Hope this helps.

yqlu:
Another suggestion for @marandalucas 

[...]

PR #650, opened 2023-10-24
Add secretless Basic Auth support

TheSpiritXIII:
Partial fix for #241. This change adds Basic Auth support but only usernames for now. Passwords are planned to be represented with secrets in a future PR.

Integration tests are planned for a future PR.


pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -209,6 +209,15 @@ type Authorization struct {
 	Credentials *corev1.SecretKeySelector `json:"credentials,omitempty"`
 }
 
+// BasicAuth sets the `Authorization` header on every scrape request with the
+// configured username.
+//
+// Currently the password is not configurable and always empty.
+type BasicAuth struct {
+	// The username for authentication.
+	Username string `yaml:"username" json:"username"`
```
bwplotka, line 218: issue: Users of CRD might be confused why there is no password. Given no better way to document things, would add big TODO why password is not yet here in commentary. WDYT? (:

bwplotka (approved):
LGTM, just a documentation nit (:

Issue #450, opened 2023-03-23
Protected scrape endpoints


TheSpiritXIII:
Support mTLS, OAuth2, Authorization and BasicAuth on the PodMonitoring ScrapeEndpoint level.

pinkfish:
Is there a way to solve this?  Just ran into this issue myself :(

AgustinRamiroDiaz:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/652 is awesome! Is there an estimate on when will it be available in GMP?

AgustinRamiroDiaz:
https://github.com/GoogleCloudPlatform/prometheus-engine/issues/241 is a subset of this one

PR #651, opened 2023-10-24
Add secretless OAuth 2 support

TheSpiritXIII:
Partial fix for https://github.com/GoogleCloudPlatform/prometheus-engine/issues/450. This change adds OAuth 2 support but without Client Secret for now. Client Secrets are planned to be represented with Kubernetes secrets in a future PR.


TheSpiritXIII:
Prometheus does not handle this correctly. I filed a bug here: https://github.com/prometheus/common/issues/528

TheSpiritXIII:
This change is blocked on a new example app image with #732 included.

TheSpiritXIII:
@bwplotka now that we have an example-app image with OAuth 2 fixed, this should be ready for review now. Can you take a look when you have a chance. Thanks!

bwplotka (approved):
Thanks!

LGTM, sounds like your preference is to merge this before https://github.com/GoogleCloudPlatform/prometheus-engine/pull/776, I am fine equally focusing on the full MVP version as a whole could be a good option too

Let me ask @pintohutch if he is happy with rebasing his e2e PR chain if we merge this 🤗, otherwise good to go!

bwplotka:
Now I am bit surprised we don't use any of "Wait for deployment" etc.. but if you manage to refactor e2e without them, I guess I trust it makes sense 🙈

PR #661, opened 2023-11-13
fix: check for permissions before updating webhooks

bernot-dev:
In preparation for removing permissions from GKE environments, add an explicit check to ensure updating CA Bundle on webhooks is allowed before blindly attempting to do so forever.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -32,6 +32,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	arv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
+	authv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -606,3 +588,63 @@ func (o *Operator) setMutatingWebhookCABundle(ctx context.Context, caBundle []by
...
+		authv1.ResourceAttributes{
+			Group:     arv1.GroupName,
+			Resource:  "ValidatingWebhookConfiguration",
+			Namespace: o.opts.OperatorNamespace,
+			Verb:      "update",
+		},
+	)
+}
+
+func (o *Operator) canI(ctx context.Context, resources ...authv1.ResourceAttributes) (bool, error) {
```
macxamin, line 609: nit: Can you make the name of this function more explicit? maybe merge with `canUpdateWebhooks`.

pkg/operator/operator.go
```diff
@@ -606,3 +588,63 @@ func (o *Operator) setMutatingWebhookCABundle(ctx context.Context, caBundle []by
...
+	for _, resource := range resources {
+		req := authv1.SelfSubjectAccessReview{
+			Spec: authv1.SelfSubjectAccessReviewSpec{
+				ResourceAttributes: &resource,
+			},
+		}
+		if err := o.client.Create(ctx, &req); err != nil {
```
TheSpiritXIII, line 616: **thought (non-blocking):** This is cool! I didn't realize that you could check permissions this way. Do you know how this compares to something like a dry-run client?

pkg/operator/operator.go
```diff
@@ -606,3 +588,63 @@ func (o *Operator) setMutatingWebhookCABundle(ctx context.Context, caBundle []by
...
+			return false, fmt.Errorf("check permissions to %q %q: %w", resource.Verb, resource.Name, err)
+		}
+		if !req.Status.Allowed {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (o *Operator) continuouslySetCABundle(ctx context.Context, caBundle []byte) {
```
TheSpiritXIII, line 626: **note:** link to non-polling solution for this: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/350

TheSpiritXIII (approved):
Thanks!

pkg/operator/operator.go
bernot-dev: canUpdateWebhooks is on `:592`

`canI` is a generic implementation based on the same idea as [kubectl auth can-i](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#checking-api-access).

pkg/operator/operator.go
bernot-dev: `can-i` just checks for permissions without checking for any other kind of API conformity, like a dry-run client might.

PR #663, opened 2023-11-15
fix: webhook TLS certificate loading

bernot-dev:
Create the `certupdater` package to handle the logic around sourcing TLS certificates. Remove all writes to the host filesystem, and store certificates in memory.

Add testdata with sample self-signed certificates for testing purposes only.


TheSpiritXIII (commented):
Thanks!

pkg/operator/certupdater/cert_updater.go
```diff
@@ -0,0 +1,212 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/util/cert"
+)
+
+const TLSCAKey = "ca.crt"
+
+type CertUpdater struct {
+	sync.RWMutex
```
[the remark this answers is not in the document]
bernot-dev: Honestly, I was just copying [certwatcher](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.2/pkg/certwatcher/certwatcher.go#L40) on that.

You're right, I think it makes sense to move it to a private field, though. We probably shouldn't expose the mutex.

pkg/operator/certupdater/cert_updater.go
```diff
@@ -0,0 +1,212 @@
...
+
+type Option func(*CertUpdater)
+
+func WithPolling(d time.Duration) Option {
+	return func(cu *CertUpdater) {
+		cu.pollingInterval = d
+	}
+}
+
+func New(source CertSource, opts ...Option) *CertUpdater {
```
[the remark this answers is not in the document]
bernot-dev: This is used now.

pkg/operator/certupdater/cert_updater.go
```diff
@@ -0,0 +1,212 @@
...
+			case <-ctx.Done():
+				done = true
+			}
+		}
+	}()
+	return nil
+}
+
+func (cu *CertUpdater) poll(ctx context.Context) error {
+	if cu.source == nil {
```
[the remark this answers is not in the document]
bernot-dev: Assuming you're referring to `Start()`?

As a general rule, I don't like spawning long-running processes in a constructor. I feel like that's unexpected behavior and may interfere with some testing scenarios.

pkg/operator/certupdater/cert_updater_test.go
```diff
@@ -0,0 +1,292 @@
...
+		t.Error("Private key is nil")
+	}
+}
+
+func TestSourceBase64(t *testing.T) {
+	type test struct {
+		keyString  string
+		certString string
+		caString   string
+		wantErr    bool
```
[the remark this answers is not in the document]
bernot-dev: I'm a bit conflicted on that point. In this case, the potential direct error type is `base64.CorruptInputError`. Although, indirect errors could occur from the TLS package, which returns untyped errors. And I think [matching on error strings is brittle](https://bitfieldconsulting.com/golang/comparing-errors#string-matching-on-errors-is-fragile).

The tests cover both types of errors (invalid base64 and invalid cert/key), despite not checking the specific error. So, I'd rather not force additional constraints on the implementation details here.

pkg/operator/certupdater/cert_updater.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+	cert, err := tls.X509KeyPair(certPEM, keyPEM)
+	if err != nil {
+		return nil, err
+	}
+
+	caPEM, err := fsys.ReadFile(TLSCAKey)
+	if err != nil {
+		caPEM = certPEM
```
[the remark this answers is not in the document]
bernot-dev: The intention is that if a separate CA cert is not provided, then the cert will be used as its own the CA. There is nothing special about generated certificates that makes them different in this regard.

None of the certs we see for the webhook server are going to be signed by a Root CA, because a Kubernetes service address (e.g. `gmp-operator.gmp-system.svc`) is not a public DNS name. So, they're all essentially "self-signed."

Sometimes (read: in GKE) the certs will be signed by the cluster CA (also different from a Root CA). In which case, that should be provided and can be used instead of the cert itself to validate the webhook server cert.

cmd/operator/main.go
```diff
@@ -38,6 +38,10 @@ import (
 	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator"
 )
 
+const (
+	defaultTLSDir = "/etc/tls/private"
```
TheSpiritXIII, line 42: question: how did you come up with this directory? Is there a standard TLS directory?

pkg/operator/certupdater/cert_updater.go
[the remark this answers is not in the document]
TheSpiritXIII: I was thinking to `New` and make it return `(*CertUpdater, error)`. This feels like an initialization error to me so to keep checking this in a loop is inefficient. Afterall, the source cannot change after the constructor. Thoughts?

pkg/operator/certupdater/cert_updater.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	ca, err := x509.ParseCertificate(pemToDER(caPEM))
+	if err != nil {
+		return nil, err
+	}
+
+	return func(context.Context) (tls.Certificate, *x509.Certificate, error) {
+		return cert, ca, nil
```
TheSpiritXIII, line 189: question: it seems like none of the certificates can be changed dynamically. Is that intended? If so, will that change anytime soon? Perhaps we overengineered this.

pkg/operator/operator.go
```diff
@@ -284,24 +292,26 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
 		logger:  logger,
 		opts:    opts,
 		client:  client,
+		getCA:   certUpdater.GetCA,
 		manager: manager,
 	}
+	logger.Info("Operater created")
```
TheSpiritXIII, line 298: nit: leftover from testing?

pkg/operator/operator.go
```diff
@@ -614,5 +576,45 @@ func (o *Operator) continuouslySetCABundle(ctx context.Context, caBundle []byte)
...
+		cu := certupdater.New(certSourceGenerated, certupdater.WithLogging(logger))
+		cert, err := cu.GetCertificate(nil)
+		if err != nil {
+			logger.Error(err, "failed to GetCertificate", "cert", cert.Certificate[0])
+		}
+		return cu
+	}
 
+	// All potential sources failed, webhook server will not function
+	panic("No supported source of TLS certificates was found")
```
TheSpiritXIII, line 619: nit: why not return error and exit gracefully?

TheSpiritXIII (changes requested):
Looking really good! Just had a few more comments.

pkg/operator/certupdater/cert_updater.go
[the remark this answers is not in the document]
bernot-dev: Oh! You're talking about `if cu.source == nil`.

I think this is necessary because it's possible (although inadvisable) to create a CertUpdater literal. `cu := CertUpdater{}` which would not have a source on it. But I should be able to make the type unexported.

cmd/operator/main.go
bernot-dev: It's not exactly standard, but this is what prometheus-operator uses.
https://github.com/prometheus-operator/prometheus-operator/blob/aedb6cacda6ffed64b43893b1f4fc12b4677db8a/pkg/server/server.go#L38

pkg/operator/certupdater/cert_updater.go
bernot-dev: Certificates from the file system can change.

pkg/operator/target_status_test.go
```diff
@@ -1223,7 +1223,7 @@ func TestPolling(t *testing.T) {
 				Port: intstr.FromString("metrics"),
 			}},
 		},
-	}).WithObjects(pod).Build()
+	}).WithObjects(pod).WithStatusSubresource(&monitoringv1.PodMonitoring{}).Build()
```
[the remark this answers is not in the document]
bernot-dev: Not sure how that slipped into this PR. Fixed.

pkg/operator/operator.go
```diff
@@ -98,6 +99,7 @@ type Operator struct {
 	logger  logr.Logger
 	opts    Options
 	client  client.Client
+	getCA   func() (*x509.Certificate, error)
```
TheSpiritXIII, line 102: nit: is this used anywhere? I couldn't find it.

pkg/operator/operator.go
```diff
@@ -131,6 +135,12 @@ type Options struct {
 	CollectorHTTPClient *http.Client
 }
 
+type getCertificateMethod interface {
```
TheSpiritXIII, line 138: nit: I'm half-tempted to say we should just use `certUpdater` since the API is exactly the same.

pkg/operator/certupdater/cert_updater.go
```diff
@@ -0,0 +1,234 @@
...
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+	return nil
+}
+
+func (cu *certUpdater) poll(ctx context.Context) error {
```
TheSpiritXIII, line 99: suggestion: I may have brought this up before but why not have `poll` live within the source? The way a source watches for changes may be different. For example, the filesystem could use the native OS watch, flags will never poll, and if we ever add something like fetching Kubernetes secrets directly, that would be a Kubernetes Watch API.

TheSpiritXIII (changes requested):
Thanks for the changes! I still have some questions before we can finally merge this. :)

pkg/operator/certupdater/cert_updater.go
bernot-dev: Basically, I found that polling is simpler and more reliable than other methods in Kubernetes.

Filesystem watches in Kubernetes behave oddly because of the way volumes work. When a change is detected using fsnotify, for instance, it gets reported as the file being removed. [Certwatcher gets around this by creating a new watch for each change](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.2/pkg/certwatcher/certwatcher.go#L182-L187). This is different from how fsnotify works outside Kubernetes. It's certainly possible to implement a similar solution, but it adds complexity.

Watches in Kubernetes are not free, and on my last team we started running into limitations on the Kubernetes API Server for how many watches could be set up at once, so I carried forward some bias from that. Each watch keeps a persistent connection open between the operator and the Kubernetes API Server. I also observed that Kubernetes secrets mounted as a volume (the intended use case) are cached and it takes about 2 minutes to propagate a change to a secret to the volume. So, whatever responsiveness you gain by using a watch is likely relatively inconsequential in that context.

Ultimately, using simple polling seemed like a better trade-off. It reduces complexity while having minimal impact on overall performance. That said, it would be possible to implement watches in the future, if we discovered a need to do so.

pkg/operator/operator.go
bernot-dev: I didn't want to have `certUpdater` be an exported type, because a `certupdater.CertUpdater{}` would not be useful. Also, it was compatible with certwatcher until I added the GetCA method.

pkg/operator/operator.go
bernot-dev: Now it is. 😄

TheSpiritXIII (changes requested):
Thanks for the update! Can you take a look at some of my small stylist suggestions?

pkg/operator/certupdater/cert_updater.go
```diff
@@ -0,0 +1,253 @@
...
+// WithPolling causes certUpdater to check for changes to certificates periodically.
+func WithPolling(d time.Duration) Option {
+	return func(cu *certUpdater) {
+		cu.pollingInterval = d
+	}
+}
+
+// New creates a new certUpdater.
+//
+//nolint:revive // Intentionally return unexported type, to use methods only.
```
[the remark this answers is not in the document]
bernot-dev: Same as: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/663#discussion_r1501214467

A `CertUpdater{}` would not be useable if it's not initialized with `New`.

PR #667, opened 2023-11-15
Fix prometheus-adapter install manifest and instructions

pintohutch:
Let's pull in the manifests from [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/blob/v0.13.0/manifests/prometheusAdapter-service.yaml) and peg them for convenient `kubectl` installing.


examples/hpa/prometheus-adapter.yaml
```diff
@@ -0,0 +1,368 @@
...
+          namespace:
+            resource: namespace
+          pod:
+            resource: pod
+      name:
+        matches: "^(.*)_total$"
+        as: "${1}_per_second"
+      metricsQuery: 'sum(rate(<<.Series>>{<<.LabelMatchers>>}[5m])) by (<<.GroupBy>>)'
+    externalRules:
+    - seriesQuery: 'logging_googleapis_com:byte_count{monitored_resource="k8s_cluster"}'
```
bwplotka, line 169: What's this, do we need it?

bwplotka (approved):
Nice! Indeed we don't how users configure or use prom-adapter so passing just configuration might have been not enough. Thanks!

examples/hpa/prometheus-adapter.yaml
pintohutch: Oh whoops - just saw this. I wanted to provide an example of an ["external" metric](https://kubernetes.io/docs/reference/external-api/external-metrics.v1beta1/) as well as a custom one.

PR #669, opened 2023-11-16
chore: switch to kube-codegen

bernot-dev:
Replace deprecated generate-groups.sh script with recommended kube_codegen.sh script.


bernot-dev:
@pintohutch I think the prometheus-adapter version change in `examples/hpa/prometheus-adapter.yaml` must be related to #667. Not sure how it got caught up in this PR.

TheSpiritXIII (approved):
Great catch! LGTM

PR #671, opened 2023-11-17
Remove generated clientsets

TheSpiritXIII:
These are obsolete now that we use controller-runtime which has a different API surface.


bernot-dev (approved):
LGTM

Issue #145, opened 2022-02-17
Succeeded Pods are not ignored


fkollmann:
Hi, when using *PodMonitoring* to monitor Pods of Jobs (esp. CronJobs), the pods are not being ignored once the Job finished and the Pod enter the 'Succeeded' phase.

This results in the following error message:

```
caller=log.go:124 component="scrape manager" scrape_pool=PodMonitoring/uplift/tkimporter-monitoring/9090 level=debug target=http://10.212.54.208:9090/metrics msg="Scrape failed" err="Get \"http://10.212.54.208:9090/metrics\": context deadline exceeded"
```

Best Regards, Felix

fabxc:
Thanks for reporting. CronJobs were not on our radar as I assumed people would mostly use PGW for it.

In general we can [filter by phase](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#pod).
Besides `Suceeded`, would you also expect pods in Failed state to be ignored?

fkollmann:
I looked into the documentation:  https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase  

It seems to me that only pods in the *Running* phase should be considered being able to provide metrics. All other phases should be ignored, from my POV.

tomasgareau:
Just thought I'd drop a quick breadcrumb here to link to how `prometheus-operator` handled this:

Issue: https://github.com/prometheus-operator/prometheus-operator/issues/4816
PR: https://github.com/prometheus-operator/prometheus-operator/pull/5049

in case the googleapis CRs wanted to mirror the coreos CRs

f0o:
Bump?

Would be great to have this, we just got a load of corrupted metrics because of this issue

pintohutch:
Hi @f0o,

We can take a look at supporting this. I am curious - how were metrics ingested with pods not `Running`? Presumably the containers wouldn't be available to scrape metrics from.

But I may be misunderstanding.

f0o:
@pintohutch The issue comes when a pod reuses the same IP as one that's not running anymore. The scraper then just uses the same Tags as the previously discovered one (since the target still exists) but the data it gets is from a whole different one.

This happens relatively often if you use Spot-Containers and they shift around a lot.

I see this about every two months on GKE Autopilot with Spot-Containers and only using 5 workloads total, so not even a lot of pods really.

pintohutch:
[...]

PR #672, opened 2023-11-17
feat: pod lifecycle filtering

pintohutch:
Only pods that are not "Failed" or "Succeeded" will be default candidates for scraping.

BREAKING CHANGE: this changes the default behavior of PodMonitoring and ClusterPodMonitoring

### DIFF
--- a/pkg/operator/apis/monitoring/v1/types.go
+++ b/pkg/operator/apis/monitoring/v1/types.go
@@ -1044,8 +1044,9 @@ type ClusterPodMonitoringSpec struct {
 	// FilterRunning will drop any pods that are in the "Failed" or "Succeeded"
 	// pod lifecycle.
 	// See: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
-	// This prevents noisy errors when attempting to scrape Suceeded pods from
-	// K8s Jobs. Additionally, it can mitigate issues with reusing stale target
+	// Specifically, this prevents scraping Suceeded pods from K8s jobs, which
+	// could contribute to noisy logs or irrelevent metrics.
+	// Additionally, it can mitigate issues with reusing stale target
 	// labels in cases where Pod IPs are reused (e.g. spot containers).
 	// See: https://github.com/GoogleCloudPlatform/prometheus-engine/issues/145
 	FilterRunning *bool `json:"filterRunning,omitempty"`
--- a/pkg/operator/apis/monitoring/v1/types.go
+++ b/pkg/operator/apis/monitoring/v1/types.go
@@ -1044,6 +1044,10 @@ type ClusterPodMonitoringSpec struct {
 	// FilterRunning will drop any pods that are in the "Failed" or "Succeeded"
 	// pod lifecycle.
 	// See: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
+	// This prevents noisy errors when attempting to scrape Suceeded pods from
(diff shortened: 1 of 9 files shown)


pintohutch:
Fixes #145

pintohutch:
The proposal here is similar to the prometheus-operator [approach](https://github.com/prometheus-operator/prometheus-operator/pull/5049), however it moves the boolean up to the pod-level, as opposed to the endpoint level. The thinking behind this is that since the [lifecycle phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) label is at the pod-level, then the configuration for scraping should be at the pod-level as well.

The drawback to this approach is: If one has a `PodMonitoring` with multiple scrape `endpoints`, and it's desired to have one endpoint be scraped regardless of pod status, while the other scraped only if the pod is "running", then this is not possible.

However, I can't imagine a case where this would be desired as, by definition, all containers in a `Succeeded` or `Failed` pod have exited. If there is some valid use case that we're not foreseeing here, it should still be possible by splitting up the `PodMonitoring` by scrape endpoint.

The considerations for making this a new default (as opposed to opt-in) are:
* Ignoring failed pods is a net beneficial change for managed collection and we want all users to have the feature with little friction.
* Scraping a "failed" pod should be the exception and not the rule. The majority of existing `PodMonitorings` shouldn't need to specify any new configuration to be covered.

Happy to discuss more though if there are other thoughts.

pintohutch:
cc @lyanco @bwplotka for your input ^

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1016,6 +1038,10 @@ type ClusterPodMonitoringSpec struct {
 	TargetLabels TargetLabels `json:"targetLabels,omitempty"`
 	// Limits to apply at scrape time.
 	Limits *ScrapeLimits `json:"limits,omitempty"`
+	// FilterRunning will drop any pods that are in the "Failed" or "Succeeded"
```
bwplotka, line 1041: Can we add to the comment what's the benefit for enabling this for the user?

bwplotka (approved):
I'm fine with that 👍🏽 LGTM for pod level too.

QQ: What's the summary benefits of this? Just to reduce overhead? Decrease up metrics samples a bit?

pkg/operator/apis/monitoring/v1/types.go
pintohutch: Done.

pintohutch:
@bwplotka - added context in the field description to answer your question. Basically the cases from #145 are the motivation. Helping folks reduce their monitoring bill is also a nice benefit 💸

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1019,6 +1041,14 @@ type ClusterPodMonitoringSpec struct {
...
+	// pod lifecycle.
+	// See: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
+	// This prevents noisy errors when attempting to scrape Suceeded pods from
```
bwplotka, line 1047: It's not errors really, the succedded (completed) jobs which are not cleaned up report normal metrics just perhaps not relevant in majority of cases (if not all). (:

bwplotka (approved):
LGTM! 💪🏽 Thanks! Just minor comment nit.

pkg/operator/apis/monitoring/v1/types.go
pintohutch: Done.

PR #679, opened 2023-11-22
Create delete_metric_descriptors.go

lyanco:
Add the metric deletion script to the repo so we can refer to it in troubleshooting documentation


examples/delete_metric_descriptors.go
```diff
@@ -0,0 +1,96 @@
...
+	for {
+		resp, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			log.Fatalf("Failed ListMetricDescriptors request: %v", err)
+		}
+		var metricType = resp.Type
+		match, err := regexp.MatchString("insert_regex_pattern_to_match", metricType)
```
lyanco, line 67: I know this should probably by a var up above but
1. I am not totally sure how to do it
2. I don't have the bandwidth to test it
3. I'm scared about putting anything in as a default value

examples/delete_metric_descriptors.go
```diff
@@ -0,0 +1,96 @@
...
+				if err != nil {
+					log.Fatalf("Failed DeleteMetricDescriptors: %v", err)
+				}
+				numTotal++
+				deleted++
+				fmt.Printf("%s deleted\n", metricType)
+				// Delete metrics in batches of 1000 metrics and sleep inbetween batches to avoid overwhelming
+				// configuration servers.
+				if deleted == 1000 {
+					time.Sleep(5 * time.Minute)
```
TheSpiritXIII, line 85: nit: Maybe put a log before the sleep? As a user, I'd be wondering why my app froze for 5 whole minutes in the middle.

examples/delete_metric_descriptors.go
bwplotka: Sure, we can take it from here

bwplotka: QQ: I guess acceptance criteria for this script is to delete metric descriptors from certain project that matches certain regex, right? I can write the script quickly so that it also asks for confirmation if we want

examples/delete_metric_descriptors.go
```diff
@@ -0,0 +1,96 @@
...
+var (
+	cloudMonitoringEndpoint = flag.String("address", "monitoring.googleapis.com:443", "address of monitoring API")
+	resourceContainer       = flag.String("resource_container", "", "target resource container, ex. projects/test-project")
+	dryRun                  = flag.Bool("dry_run", false, "whether to dry run or not")
+)
+
+/*
+* To acquire Application Default Credentials, run:
+
+gcloud auth application-default login
```
TheSpiritXIII, line 28: question: How do you pass this to the tool or does the tool automatically know?

examples/delete_metric_descriptors.go
```diff
@@ -0,0 +1,96 @@
...
+
+* One way to run this file is to initialize a go module.
```
TheSpiritXIII, line 30: question: Why? Is this because of the dependencies?

examples/delete_metric_descriptors.go
TheSpiritXIII: For 3, we can have an empty string as the default and have the tool fail if it's empty. What do you think?

TheSpiritXIII (changes requested):

examples/delete_metric_descriptors.go
lyanco: @bwplotka yes - that's exactly it - to help people delete bad (statsd-style) descriptors from the system so they can free up their quota and speed up their autocomplete.

lyanco: We might want to expand it to accept an array of projects that it iterates through, as often times you make the same bad metric names across multiple projects.

bwplotka:
On it, will push some tested script to this PR in ~1h

examples/delete_metric_descriptors.go
bwplotka: All done.

TheSpiritXIII (approved):
Great! Just a couple of nits but good to go!

examples/scripts/delete_metric_descriptors/delete_metric_descriptors.go
```diff
@@ -0,0 +1,188 @@
...
+	"unicode"
+
+	monitoring "cloud.google.com/go/monitoring/apiv3/v2"
+	"cloud.google.com/go/monitoring/apiv3/v2/monitoringpb"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+)
+
+/*
+This script deletes metric descriptors from the given projects (-projects flag),
```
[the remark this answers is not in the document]
bwplotka: Generally yes, but for one-file script it might be more accessible here? If we focus to always have one file for simplicity. It's robust code, but only an example after all.

examples/scripts/delete_metric_descriptors/delete_metric_descriptors.go
```diff
@@ -0,0 +1,188 @@
...
+		if saEnvVar != "" {
+			return []option.ClientOption{
+				option.WithEndpoint(endpoint),
+				option.WithCredentialsJSON([]byte(os.Getenv(saEnvVar))),
+			}
+		}
+		return []option.ClientOption{option.WithEndpoint(endpoint)}
+	}()...)
+	if err != nil {
+		log.Fatalf("failed to build client for %s", endpoint)
```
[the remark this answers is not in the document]
bwplotka: ups, missed this

bwplotka:
Addressed/responded to your comments @TheSpiritXIII, thanks for the review!

examples/scripts/delete_metric_descriptors/delete_metric_descriptors.go
```diff
@@ -0,0 +1,188 @@
...
+	cloudMonitoringEndpoint = flag.String("address", "monitoring.googleapis.com:443", "address of monitoring API")
+
+	projectNames    = flag.String("projects", "", "required: comma-separated project IDs of the projects on which to execute the requests. Name format is as defined in https://cloud.google.com/monitoring/api/ref_v3/rpc/google.monitoring.v3#listmetricdescriptorsrequesttarget, e.g. projects/test-project,projects/test-project2")
+	metricTypeRegex = flag.String("metric_type_regex", "", "required: RE2 regex expression matching metric.type (anchored), so metric descriptor names to delete. Guarded with the interactive 'y' confirmation. See --dry_run to only print those")
+	dryRun          = flag.Bool("dry_run", false, "whether to dry run or not")
+
+	serviceAccountEnvVar = flag.String("sa-envvar", "", "optional environment variable containing Google Service Account JSON, without it application-default flow will be used.")
+)
+
+func deleteDescriptors(endpoint string, projects []string, re2 *regexp.Regexp, saEnvVar string, dryRun bool) error {
```
TheSpiritXIII, line 57: nit: why is it called `re2`? Where is `re1`?

examples/scripts/scripts_test.go
```diff
@@ -0,0 +1,168 @@
...
+			},
+		},
+	}); err != nil {
+		t.Fatalf("test data write: %v", err)
+	}
+
+	assertExpectedTypesExists(t, client, reqName, expectedTypes)
+
+	b := bytes.Buffer{}
+	cmd := exec.Command(
```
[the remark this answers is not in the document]
TheSpiritXIII: SGTM!

examples/scripts/delete_metric_descriptors/delete_metric_descriptors.go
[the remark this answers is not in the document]
TheSpiritXIII: A `README.md` is more accessible because it's what users would naturally look for and we'd get nice rendering in GitHub, etc. There's some intimidation looking at the code but maybe for a sensitive script like this, that's what we want so I'm indifferent.

examples/scripts/delete_metric_descriptors/delete_metric_descriptors.go
bwplotka: It's for https://github.com/google/re2/wiki/Syntax but can change 🙈

bwplotka: Fixed!

PR #680, opened 2023-11-26
Add TLS insecure skip verify tests for ClusterPodMonitoring

TheSpiritXIII:
This PR adds ClusterPodMonitoring-equivalent tests for the TLS insecure-skip-verify flag.

The method was moved into a utility method, which allows us to reuse this method for other authorization types (e.g basic auth).


e2e/authorization_test.go
```diff
@@ -16,149 +16,258 @@ package e2e
...
-	"k8s.io/apimachinery/pkg/util/wait"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func TestTLS(t *testing.T) {
-	t.Parallel()
-	tctx := newOperatorContext(t)
-	ctx := context.Background()
+func isPodMonitoringTargetCertificateError(message string) error {
+	const expected = "tls: failed to verify certificate: x509: certificate signed by unknown authority"
```
bwplotka, line 32: Looks a bit brittle, but little harm in trying this! 😄 🤞🏽

bwplotka (approved):
Sweet, thanks!

e2e/authorization_test.go
TheSpiritXIII: Switched to the raw errors. Sadly it's coming from Prometheus' `/target` endpoint so we can't use `errors.Is`, but we can use string comparison.

PR #684, opened 2023-11-28
feat: [kubelet/cAdvisor metrics package] add NodeMonitoring ScrapeConfig logic

macxamin:
All changes are in types.go and types_test.go. The rest of the files are generated via `make regen`

types.go changes:
1. Add functions needed to support NodeMonitoring. Will refactor code in the future to reuse functions from PodMonitoring.

types_test.go changes:
1. Add selector to `TestPodMonitoring_ScrapeConfig` and `TestClusterPodMonitoring_ScrapeConfig` since `relabelingsForSelector` was modified. 
2. Added a tests to validate and create NodeMonitoring Scrape configs.


pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1405,6 +1417,201 @@ type NodeMonitoring struct {
 	Spec NodeMonitoringSpec `json:"spec"`
 }
 
+// NodeScrapeEndpoint specifies a node's Prometheus metrics endpoint to scrape.
+type NodeScrapeEndpoint struct {
```
bwplotka, line 1421: * Hm.. I am missing "port" 🤔 
* Somehow interval is required?
image
* Scheme (you can totally use TLS) and basic HTTPClientConfig without secrets would be nice for a start 🤔 (to at least skip verify?)
* Params we could bring back as well, there is little harm

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1405,6 +1417,201 @@ type NodeMonitoring struct {
...
+}
+
+func (nm *NodeMonitoring) endpointScrapeConfig(index int, projectID, location, cluster string) (*promconfig.ScrapeConfig, error) {
+	// Filter targets that belong to selected nodes.
+	relabelCfgs, err := relabelingsForSelector(nm.Spec.Selector, nm)
+	if err != nil {
+		return nil, err
+	}
+
+	relabelCfgs = append(relabelCfgs,
```
bwplotka, line 1484: idea: hmmm I wonder if there's no opportunity to reuse some code between Pod/Cluster monitorings. Not a blocker (:

bwplotka (commented):
Thanks! Some comments around left out fields in NodeMonitoring, otherwise fine 👍🏽

pkg/operator/apis/monitoring/v1/types.go
macxamin: - I removed port, since it's not part of the [prometheus metadata labels for nodes. ](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#node) Do you see users wanting to user port?
- If IIUC the Kubelet serves metrics over HTTPs, so I removed it to reduce the chance of a user error and simplify the code. I guess we can return error if it's not https. The kubelet has it's ca. So we need the HTTPClientConfig to look like this:
image
Should we expose HTTPClientConfig and just block those fields?
- Yeah, I can bring it back.

bwplotka: Got it, sounds like you want to tailor it to kubelet. However given "NodeMonitoring" semantics that might be surprising 🤔 I think I would be ok to start with tailored kubelet approach you propose, wonder what's @pintohutch opinion.

pintohutch: Discussed with @macxamin offline - IIUC we want to keep port number and http as options. Port name however isn't applicable for the node as it's not in the [spec](https://pkg.go.dev/k8s.io/api/core/v1#NodeDaemonEndpoints).

bwplotka:
Let us know once it's rdy for a second review iteration @macxamin ! 💪🏽

pkg/operator/apis/monitoring/v1/types.go
macxamin: Yes, I will add port logic in a future PR. I want to get nodeMonitoring to a working state and then I can make refinements to it.

macxamin:
> Let us know once it's rdy for a second review iteration @macxamin ! 💪🏽

@bwplotka Ready!

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1406,6 +1418,189 @@ type NodeMonitoring struct {
...
+		},
+		TLSConfig: config.TLSConfig{
+			CAFile: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+		},
+	}
+	return extractScrapeParams(nm.GetKey(), ep, httpCfg, relabelCfgs, discoveryCfgs, nm.Spec.Limits)
+}
+
+// extractScrapeParams extracts the scrape parameters of a NodeMonitoring ScrapeEndpoint.
+func extractScrapeParams(id string, ep ScrapeEndpoint, httpCfg config.HTTPClientConfig, relabelCfgs []*relabel.Config, discoveryCfgs discovery.Configs, limits *ScrapeLimits) (*promconfig.ScrapeConfig, error) {
```
bwplotka, line 1540: hm... something could be improved here. Is it specific to NodeMonitoring (as comment suggests?). I don't think it is at the moment.

Also the function name is pretty generic here.. maybe simply `newPromScrapeConfig` is better? Or `endpointScrapeConfigForNM`, or `func (nm *NodeMonitoring) newPromScrapeConf`? Essentially it feels you tried to mimic https://github.com/GoogleCloudPlatform/prometheus-engine/blob/69d276c973b8e4f00531a3366e4140998a10da4d/pkg/operator/apis/monitoring/v1/types.go#L616 function, but this function is only only used once. 

In this state you could reuse `extractScrapeParam` function in https://github.com/GoogleCloudPlatform/prometheus-engine/blob/69d276c973b8e4f00531a3366e4140998a10da4d/pkg/operator/apis/monitoring/v1/types.go#L616, but not sure if want reuse ScrapeEndpoint for NodeMonitoring, what's the plan there, given my comments above?

bwplotka (commented):
Thanks! Some questions/suggestions (:

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1406,6 +1418,189 @@ type NodeMonitoring struct {
...
+	var metricRelabelCfgs []*relabel.Config
+	for _, r := range ep.MetricRelabeling {
+		rcfg, err := convertRelabelingRule(r)
+		if err != nil {
+			return nil, err
+		}
+		metricRelabelCfgs = append(metricRelabelCfgs, rcfg)
+	}
+
+	// return interval, timeout, path, metricRelabelCfgs, nil
```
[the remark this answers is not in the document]
macxamin: good catch

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1406,6 +1418,189 @@ type NodeMonitoring struct {
...
+				{
+					Role:  discoverykube.RoleNode,
+					Field: fmt.Sprintf("metadata.name=$(%s)", EnvVarNodeName),
+				},
+			},
+		},
+	}
+
+	ep := nm.Spec.Endpoints[index]
+	// TODO(macxamin) Add support for int port. String ports are not supported by the Kubernetes spec: https://pkg.go.dev/k8s.io/api/core/v1#NodeDaemonEndpoints.
```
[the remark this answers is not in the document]
macxamin: Discussed offline, will add via feature branch

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1406,6 +1418,189 @@ type NodeMonitoring struct {
...
+	if ep.Port.StrVal != "" || ep.Port.IntVal != 0 {
+		return nil, errors.New("port not supported")
+	}
+	// TODO(macxamin) Add support for HTTPClientConfig. Add it for the user for now.
```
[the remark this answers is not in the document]
macxamin: Discussed offline, will add via feature branch

pkg/operator/apis/monitoring/v1/types.go
macxamin: done. PTAL.

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1374,13 +1337,52 @@ type RulesStatus struct {
 	// TODO: add status information.
 }
 
+// CommonScrapeEndpointSpec
```
bwplotka, line 1340: Some comment left from older impl?

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1374,13 +1337,52 @@ type RulesStatus struct {
...
+	timeout          string
+	params           map[string][]string
+	metricRelabeling []RelabelingRule
+	relabelCfgs      []*relabel.Config
+	httpCfg          config.HTTPClientConfig
+	discoveryCfgs    discovery.Configs
+	limits           *ScrapeLimits
+}
+
+// ScrapeNodeEndpoint specifies a Prometheus metrics endpoint on a node to scrape.
```
bwplotka, line 1355: Could you summarize here, the difference between ScrapeEndpoint and this (e.g TLS fields are not present)?  (for users & future devs)

pkg/operator/apis/monitoring/v1/types.go
```diff
@@ -1374,13 +1337,52 @@ type RulesStatus struct {
 	// TODO: add status information.
 }
 
+// CommonScrapeEndpointSpec
+type endpointParams struct {
```
bwplotka, line 1341: Hm, sounds like this is only for one function use (that we use twice). Perhaps we could keep ScrapeEndpoint in that function as param and kill this struct (and make sure we transition from ScrapeNodeEndpoint to ScrapeEndpoint for that one function)? Sounds like a similar choice, just less code (I understand roughly ScrapeEndpoint >= ScrapeNodeEndpoint in scope)

bwplotka (approved):
Thank you! Two suggestions, otherwise LGTM!

pkg/operator/apis/monitoring/v1/types.go
macxamin: Great idea!

PR #688, opened 2023-12-04
Upgrade to yaml v3

TheSpiritXIII:
One less dependency to manage.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,6 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20230711160842-782d3b101e98
 	google.golang.org/grpc v1.58.3
 	google.golang.org/protobuf v1.31.0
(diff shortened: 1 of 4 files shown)


bwplotka (approved):
👍🏽

TheSpiritXIII:
Upon further review, this would be a risky change to merge. The reason being is that the yaml library is no longer maintained. Kubernetes is looking into forking it, but until then, this would be a breaking change.

Here's why: v2 serializes an object such as:
```yaml
array:
- map
    key1: value1
```

While v3 would serialize the same object as:
```yaml
array:
  - map
      key1: value1
```

Both are valid YAML and there is no way to customize this on the library-level.

If the user does absolutely nothing, this may break a user's scraping completely. For example, if this change brings the user's config over the 1 MiB limit for ConfigMaps, the user will have to enable compression if they even realize their scraping broke.

PR #689, opened 2023-12-04
Use a single manager cache for the operator

TheSpiritXIII:
The recent controller-runtime allows use to use a single cache for the operator and remove our hacky secondary cache.


bwplotka (approved):
Makes sense, thanks for the cleanup!

PR #690, opened 2023-12-04
refactor: remove replica counts from manifests

bernot-dev:
In GKE, these fields will be reconciled if they are set. However, they default to 1 anyways, so this does not change behavior.


pintohutch (approved):

bwplotka (commented):
👍🏽

PR #691, opened 2023-12-04
feat: autoscale alertmanager and rule-evaluator

bernot-dev:
Scale the Alertmanager StatefulSet and Rule Evaluator Deployment to zero when there are no active rules in the cluster.

### DIFF
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -142,8 +142,8 @@ func (r *rulesReconciler) scaleRuleConsumers(ctx context.Context) error {
 	var desiredReplicas int32
 
 	var hasAnyRules bool
-	for _, check := range []RuleCheck{r.hasRules, r.hasClusterRules, r.hasGlobalRules} {
-		hasRules, err := check(ctx)
+	for _, check := range []ruleCheck{hasRules, hasClusterRules, hasGlobalRules} {
+		hasRules, err := check(ctx, r.client)
 		if err != nil {
 			return err
 		}
@@ -160,7 +160,7 @@ func (r *rulesReconciler) scaleRuleConsumers(ctx context.Context) error {
 	if err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: "alertmanager"}, &alertManagerStatefulSet); errors.IsNotFound(err) {
 		msg := fmt.Sprintf("Alertmanager StatefulSet not found, cannot scale to %d. In-cluster Alertmanager will not function.", desiredReplicas)
 		logger.Error(err, msg)
-	} else if client.IgnoreNotFound(err) != nil {
+	} else if err != nil {
 		return err
 
 	} else if *alertManagerStatefulSet.Spec.Replicas != desiredReplicas {
@@ -174,7 +174,7 @@ func (r *rulesReconciler) scaleRuleConsumers(ctx context.Context) error {
 	if err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: "rule-evaluator"}, &ruleEvaluatorDeployment); errors.IsNotFound(err) {
 		msg := fmt.Sprintf("Rule Evaluator Deployment not found, cannot scale to %d. In-cluster Rule Evaluator will not function.", desiredReplicas)
 		logger.Error(err, msg)
-	} else if client.IgnoreNotFound(err) != nil {
+	} else if err != nil {
 		return err
 	} else if *ruleEvaluatorDeployment.Spec.Replicas != desiredReplicas {
 		*ruleEvaluatorDeployment.Spec.Replicas = desiredReplicas
@@ -185,25 +185,25 @@ func (r *rulesReconciler) scaleRuleConsumers(ctx context.Context) error {
 	return nil
 }
 
-type RuleCheck func(context.Context) (bool, error)
+type ruleCheck func(context.Context, client.Client) (bool, error)
 
-func (r *rulesReconciler) hasRules(ctx context.Context) (bool, error) {
+func hasRules(ctx context.Context, c client.Client) (bool, error) {
 	var rules monitoringv1.RulesList
-	if err := r.client.List(ctx, &rules); err != nil {
+	if err := c.List(ctx, &rules); err != nil {
 		return false, err
 	}
 	return len(rules.Items) > 0, nil
 }
-func (r *rulesReconciler) hasClusterRules(ctx context.Context) (bool, error) {
+func hasClusterRules(ctx context.Context, c client.Client) (bool, error) {
 	var rules monitoringv1.ClusterRulesList
-	if err := r.client.List(ctx, &rules); err != nil {
+	if err := c.List(ctx, &rules); err != nil {
 		return false, err
 	}
 	return len(rules.Items) > 0, nil
 }
-func (r *rulesReconciler) hasGlobalRules(ctx context.Context) (bool, error) {
+func hasGlobalRules(ctx context.Context, c client.Client) (bool, error) {
 	var rules monitoringv1.GlobalRulesList
-	if err := r.client.List(ctx, &rules); err != nil {
+	if err := c.List(ctx, &rules); err != nil {
 		return false, err
 	}
 	return len(rules.Items) > 0, nil
--- a/pkg/operator/rules.go
+++ b/pkg/operator/rules.go
@@ -20,7 +20,9 @@ import (
 
 	"github.com/go-logr/logr"
 	yaml "gopkg.in/yaml.v3"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


pkg/operator/rules.go
```diff
@@ -126,9 +127,83 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+	var alertManagerStatefulSet appsv1.StatefulSet
+	if err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: "alertmanager"}, &alertManagerStatefulSet); err != nil {
+		return err
+	}
+	var ruleEvaluatorDeployment appsv1.Deployment
+	if err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: "rule-evaluator"}, &ruleEvaluatorDeployment); err != nil {
+		return err
+	}
+
+	if *alertManagerStatefulSet.Spec.Replicas != desiredReplicas {
```
macxamin, line 165: Should we check if any alerts are configured in the rule-evaluator, if no alerts then we can turn off the alertmanager?

https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/operator/apis/monitoring/v1/types.go#L1356-L1359

macxamin (commented):
Good job! I tested it, however I still see the rule-evaluator and alertmanager active but in the pending state. 

I used `DOCKER_PUSH=1 make operator` and made sure there was no rule, clusterrules or Globalrules. Maybe I'm doing something wrong, Let's check it over chat.

pkg/operator/rules.go
```diff
@@ -126,9 +127,83 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+	return len(rules.Items) > 0, nil
+}
+func (r *rulesReconciler) hasGlobalRules(ctx context.Context) (bool, error) {
+	var rules monitoringv1.GlobalRulesList
+	if err := r.client.List(ctx, &rules); err != nil {
+		return false, err
+	}
+	return len(rules.Items) > 0, nil
+}
+func ignoreErr(b bool, err error) bool {
```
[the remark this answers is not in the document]
bernot-dev: Ah yes, I forgot about that. Good catch. Thanks!

pkg/operator/rules.go
bernot-dev: We could, if you think it's worthwhile. I was making the (perhaps incorrect) assumption that customers are likely to set up recording rules and alerting rules in the same place, whether that's in-cluster or externally. And it somewhat simplifies the solution to not have to inspect the rules for their type.

Given that the goal is just having both off by default, I'm not sure there's meaningful value added by handling them separately. Thoughts?

pkg/operator/rules.go
```diff
@@ -126,9 +127,80 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+			break
+		}
+	}
+	if hasAnyRules {
+		desiredReplicas = 1
+	}
+
+	var alertManagerStatefulSet appsv1.StatefulSet
+	if err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: "alertmanager"}, &alertManagerStatefulSet); err != nil {
+		return err
```
macxamin, line 158: What if the user deleted alertmanager but not rule-evaluator. In that case this code will just return err and not scale the rule-evaluator. Let's just to continue in that case? This is a suggestion of what that code could look like:

```suggestion
	var alertManagerStatefulSet appsv1.StatefulSet
	err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: NameAlertmanager}, &alertManagerStatefulSet)
	if err != nil {
		// only return error if alertmanager exists
		if !apierrors.IsNotFound(err) { 
			return err
		}
	} else {
		if *alertManagerStatefulSet.Spec.Replicas != desiredReplicas {
			*alertManagerStatefulSet.Spec.Replicas = desiredReplicas
			if err := r.client.Update(ctx, &alertManagerStatefulSet); err != nil {
				return err
			}
		}
	}
```

pkg/operator/rules.go
```diff
@@ -126,9 +127,80 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+	}
+	var ruleEvaluatorDeployment appsv1.Deployment
```
macxamin, line 160: add a check for the edge case if the rule-evaluator doesn't exist. Similar to above comment.

pkg/operator/rules.go
macxamin: I think it's out of scope for this PR, and might be overkill but I think there is no reason to have an alertmanager if you have no alerts. 

You are right alerts and recording rules are set in the same place. 

We can revist this later. perhaps addd a comment saying that could be possible. Up to you!

pkg/operator/rules_test.go
```diff
@@ -246,3 +251,101 @@ func TestGenerateGlobalRules(t *testing.T) {
...
+			Replicas: &ruleEvaluatorReplicas,
+		},
+	}
+	emptyRules := newFakeClientBuilder().WithObjects(&alertManager, &ruleEvaluator).Build()
+	hasRule := newFakeClientBuilder().WithObjects(&alertManager, &ruleEvaluator, &monitoringv1.Rules{}).Build()
+	errGettingRules := newFakeClientBuilder().WithInterceptorFuncs(interceptor.Funcs{
+		List: func(_ context.Context, _ client.WithWatch, _ client.ObjectList, _ ...client.ListOption) error {
+			return errors.New("error")
+		}}).Build()
+
```
macxamin, line 316: Can you add two unit tests where the rule evaluator/alert manager are missing and see if it still works as expected.

macxamin (changes requested):
Great job! Code works but is missing 1 edge case! I ran:
1. DOCKER_BUILD=1 make operator
2.  kubectl apply -f manifests/operator.yaml 
3. Confirmed rule-evaluator and alertmanager had 0 replicas.
4. Created rules using `kubectl apply -n gmp-system -f https://raw.githubusercontent.com/GoogleCloudPlatform/prometheus-engine/v0.7.4/examples/rules.yaml` and watched replicas become 1
5. deleted the rules and saw replicas become 0

Some comments:
1. I did find an edge case where if I delete the alertmanager or rule-evaluator the code fails prematurely and doesn't work as expected. Added some comments on how we can resolve that.

2. Would it be possible to add an E2E test in (ruler_test.go)[https://github.com/GoogleCloudPlatform/prometheus-engine/blob/03d823efbbfe68eb91f2796e406491211befa905/e2e/ruler_test.go#L53]. Confirm before `testRuleEvaluatorConfig` that replicas are 0 and then after `testRuleEvaluatorConfig` is run we can confirm that the replicas is 1.

pintohutch:
Hey @bernot-dev @macxamin - what's the status here?

bernot-dev:
I think the last outstanding issue is about an E2E test. Personally, I feel good about the coverage from the existing test with fake client.

pkg/operator/rules_test.go
```diff
@@ -246,3 +251,117 @@ func TestGenerateGlobalRules(t *testing.T) {
...
+			}
+			if err != nil && !tc.wantErr {
+				t.Errorf("Unexpected error: %s", err)
+			}
+		})
+	}
+}
+
+func TestScaleRuleConsumers(t *testing.T) {
+	var alertmanagerReplicas int32
```
pintohutch, line 292: suggestion: Could you initialize this to 0 to make it more clear?

Same with rule-evaluator below.

bernot-dev: This is an [explicit non-decision in the Go Style Decisions](go/go-style/decisions#non-decisions). Here, I prefer this format because the type is important, and I'm using the zero value - which is well-known and accepted behavior in Go, in my experience.

`alertmanagerReplicas := 0` would be an `int`, not an `int32`.

The alternative is `alertmanagerReplicas := int32(0)` which feels odd to me because it has the format of a type cast while initializing a new variable. If I was initializing a field on a struct to its zero value, I would not assign that a specific value.

The long version (`var alertmanagerReplicas int32 = 0`) [is discouraged](go/go-style/best-practices#vardeclinitialization).

Do you feel strongly about this?

pkg/operator/rules.go
```diff
@@ -126,9 +128,87 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+	} else if *ruleEvaluatorDeployment.Spec.Replicas != desiredReplicas {
+		*ruleEvaluatorDeployment.Spec.Replicas = desiredReplicas
+		if err := r.client.Update(ctx, &ruleEvaluatorDeployment); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type RuleCheck func(context.Context) (bool, error)
```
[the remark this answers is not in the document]
bernot-dev: Good catch. I'll make that change.

pkg/operator/rules.go
```diff
@@ -126,9 +128,87 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+
+func (r *rulesReconciler) hasRules(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
bernot-dev: Yeah, that's fair. I'll make the change.

pkg/operator/rules.go
```diff
@@ -126,9 +128,87 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+	}
+	if hasAnyRules {
+		desiredReplicas = 1
+	}
+
+	var alertManagerStatefulSet appsv1.StatefulSet
+	if err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: "alertmanager"}, &alertManagerStatefulSet); errors.IsNotFound(err) {
+		msg := fmt.Sprintf("Alertmanager StatefulSet not found, cannot scale to %d. In-cluster Alertmanager will not function.", desiredReplicas)
+		logger.Error(err, msg)
+	} else if client.IgnoreNotFound(err) != nil {
```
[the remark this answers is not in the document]
bernot-dev: Sure is. I think there was a time while I was writing this when I had the `IgnoreNotFound` case before the `IsNotFound` case, but it's not needed here. I'll update accordingly.

pkg/operator/rules.go
```diff
@@ -126,9 +128,87 @@ func (r *rulesReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+			break
+		}
+	}
+	if hasAnyRules {
+		desiredReplicas = 1
+	}
+
+	var alertManagerStatefulSet appsv1.StatefulSet
+	if err := r.client.Get(ctx, client.ObjectKey{Namespace: r.opts.OperatorNamespace, Name: "alertmanager"}, &alertManagerStatefulSet); errors.IsNotFound(err) {
+		msg := fmt.Sprintf("Alertmanager StatefulSet not found, cannot scale to %d. In-cluster Alertmanager will not function.", desiredReplicas)
```
pintohutch, line 161: nitpick (non-blocking): what's the benefit of logging the error here over returning an error and letting it bubble up to be logged by the upstream process (e.g. the controller manager)?

pkg/operator/rules_test.go
pintohutch: I do not - thanks for pointing out those references to me!

pintohutch (approved):
% non-blocking comment

pkg/operator/rules.go
bernot-dev: In a practical sense, this allows it to continue and scale the ruleEvaluator without relying on return 

Whether it's an "error" if the alertmanager or rule-evaluator don't exist at all seems to be something of a philosophical question.

pintohutch: Ah fair point

pintohutch (approved):

bernot-dev:
@macxamin Any additional thoughts on the E2E test? Personally, I don't see how it would cover the code in this PR any better than the existing fakeClient tests.

macxamin:
Great job! 👏

PR #693, opened 2023-12-05
Replace deprecated wait methods

TheSpiritXIII:
The update to the new wait methods broke a couple of tests due to the new immediate execution of the poll callback. The fix is just updating the poll methods to retry after an error. This is something we would have had to fix anyway because failures could still happen without the immediate execution, if the operator had any delay updating the resources. We have seen some flaky tests in the recent past, and this might have been the reason.

### DIFF
--- a/e2e/alertmanager_test.go
+++ b/e2e/alertmanager_test.go
@@ -119,9 +119,10 @@ func testAlertmanager(ctx context.Context, t *OperatorContext, spec *monitoringv
 }
 
 func testAlertmanagerDeployed(ctx context.Context, t *OperatorContext, config *monitoringv1.ManagedAlertmanagerSpec) {
-	err := wait.Poll(time.Second, 1*time.Minute, func() (bool, error) {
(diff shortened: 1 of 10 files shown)


e2e/alertmanager_test.go
```diff
@@ -155,15 +157,18 @@ func testAlertmanagerDeployed(ctx context.Context, t *OperatorContext, config *m
 			}
 
 			if diff := cmp.Diff(strings.Join(wantArgs, " "), getEnvVar(c.Env, "EXTRA_ARGS")); diff != "" {
-				return false, fmt.Errorf("unexpected flags (-want, +got): %s", diff)
+				err = fmt.Errorf("unexpected flags (-want, +got): %s", diff)
 			}
-			return true, nil
+			return err == nil, nil
```
bwplotka, line 162: You don't want return error?

bwplotka (approved):
LGTM, thanks!

PR #694, opened 2023-12-05
Add user namespace to e2e tests

TheSpiritXIII:
This PR adds a 3rd namespace to e2e tests.

Here's when you use each now:
1. `namespace`: The operator namespace. Use only for operator resources or for `PodMonitoring` or `ClusterPodMonitorings` that are self-scraping.
2. `pubNamespace`: The operator public namespace. Use only for resources that are stored here, e.g. `OperatorConfig` and the rule evaluator `ConfigMaps` and `Secrets` .
3. `userNamespace`: For everything else -- applications, `PodMonitorings` or `ClusterPodMonitorings`, application `Secrets`. etc.

### DIFF
--- a/e2e/operator_context_test.go
+++ b/e2e/operator_context_test.go
@@ -198,7 +198,7 @@ func TestMain(m *testing.M) {
 type OperatorContext struct {
 	*testing.T
 
-	namespace, pubNamespace string
+	namespace, pubNamespace, userNamespace string
 
 	kClient kubeutil.DelegatingClient
 }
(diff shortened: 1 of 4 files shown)


e2e/operator_context_test.go
```diff
@@ -347,6 +349,13 @@ func createNamespaces(ctx context.Context, kubeClient client.Client, opNamespace
...
 		return err
 	}
+	if err := kubeClient.Create(ctx, &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: userNamespace,
+		},
+	}); err != nil {
+		return err
+	}
 	return nil
```
bwplotka, line 359:
```suggestion
	return kubeClient.Create(ctx, &corev1.Namespace{
		ObjectMeta: metav1.ObjectMeta{
			Name: userNamespace,
		},
	})
```

e2e/ruler_test.go
```diff
@@ -116,7 +116,7 @@ func createRuleEvaluatorOperatorConfig(ctx context.Context, t *OperatorContext,
 				Alertmanagers: []monitoringv1.AlertmanagerEndpoints{
 					{
 						Name:       "test-am",
-						Namespace:  t.namespace,
+						Namespace:  t.userNamespace,
```
bwplotka, line 119: hm, or namespace (depends if we test managed or unmanaged?)

bwplotka (approved):
Nice! One question and a small sugestion, both nits. Thanks!

e2e/ruler_test.go
TheSpiritXIII: Good catch but that is not how managed alertmanager is configured! It looks like we automagically insert managed-alertmanager into the YAML configuration if we find the service (in `operator_config.go`; see the beginning of `makeAlertmanagerConfigs`.

PR #698, opened 2023-12-06
Auth header e2e test and update auth header example logic

TheSpiritXIII:
This PR adds a placeholder for the authentication header e2e test.

Additionally, I also updated the authentication header parsing based on [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-introduction), which says that there could be any number of spaces between the scheme and credentials.


bwplotka (approved):
Thanks 👍🏽

PR #700, opened 2023-12-07
Update monitoring protobuf import

TheSpiritXIII:
`cloud.google.com/go/monitoring/apiv3/v2/monitoringpb` is deprecated


macxamin (approved):
Nice job! 

Nit: add in the description that `cloud.google.com/go/monitoring/apiv3/v2/monitoringpb` is now deprecated. So we have a record of why this change was needed.

PR #703, opened 2023-12-08
Prepare 0.8.2 release

TheSpiritXIII:
After this PR, we will update the export version and then create the tag.

### DIFF
--- a/cmd/operator/deploy/operator/12-alertmanager.yaml
+++ b/cmd/operator/deploy/operator/12-alertmanager.yaml
@@ -41,7 +41,7 @@ spec:
       labels:
         app: managed-prometheus-alertmanager
         app.kubernetes.io/name: alertmanager
-        app.kubernetes.io/version: 0.8.0
+        app.kubernetes.io/version: 0.8.2
       annotations:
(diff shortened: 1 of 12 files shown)


cmd/operator/deploy/operator/12-alertmanager.yaml
```diff
@@ -97,7 +97,7 @@ spec:
             - all
           privileged: false
       - name: config-reloader
-        image: gke.gcr.io/prometheus-engine/config-reloader:v0.8.0-gke.4
+        image: gke.gcr.io/prometheus-engine/config-reloader:v0.8.2
```
bwplotka, line 100: Hm, no such image 🤔 

https://console.cloud.google.com/gcr/images/gke-release/EU/prometheus-engine/config-reloader

bwplotka (approved):
Thanks! Just I think the image version is wrong in our YAMLs

cmd/operator/deploy/operator/12-alertmanager.yaml
TheSpiritXIII: Good catch. Thanks!

PR #707, opened 2023-12-12
fix: webhook permissons check

bernot-dev:
Webhook permissions check was underspecified. Operator only has permission to update specific webhookconfig, not all webhookconfigs.

Fix oversight on #661.


TheSpiritXIII (approved):
Thanks for figuring this out!

bwplotka (commented):
👍🏽  Thanks for the investigation!

PR #708, opened 2023-12-13
Fixed gcm based tests failures for external contributors.

bwplotka:
Intention is to skip not fail those. There was bug with IF makefile statement.


bwplotka:
Main is broken, fixing in another PR.

TheSpiritXIII (approved):
Thanks!

Issue #696, opened 2023-12-06
v0.9.0* self-deployed operator does not inject `caBundle`


bwplotka:
Applying PodMonitorings with self-deployed operator (using our manifests/...) from 0.9.0 releases yields following errors:

```
Error from server (InternalError): error when creating "examples/pod-monitoring.yaml": Internal error occurred: failed calling webhook "default.podmonitorings.gmp-operator.gmp-system.monitoring.googleapis.com": failed to call webhook: Post "https://gmp-operator.gmp-system.svc:443/default/monitoring.googleapis.com/v1/podmonitorings?timeout=10s": no endpoints available for service "gmp-operator"
``` 

For some reasons it does not inject `caBundles` to webhooks. 

To investigate. Not sure if managed collection is affected, but we didn't release it there yet.

Reverting to `v0.8.1-gke.6` fixes this.

bwplotka:
Still can reproduce, even with https://github.com/GoogleCloudPlatform/prometheus-engine/pull/707 fix, investigating...🤔

bwplotka:
I can confirm we [essentially hit `updateAllowed = false` here](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/570ffb1930bc0937d6464645146a28348e8fb04e/pkg/operator/operator.go#L303), so either we check permissions wrongly, or we actually don't have enough permissions.

PR #712, opened 2023-12-13
operator: Fixed operator caBundle injection.

bwplotka:
Fixes https://github.com/GoogleCloudPlatform/prometheus-engine/issues/696

I build `gcr.io/gpe-test-1/operator:bw-test5` from this branch and manually tested with self-deployed operator from manifests/.*

This essentially reverts https://github.com/GoogleCloudPlatform/prometheus-engine/pull/661, but I think it makes sense to revert, the case we tried to handle better was never seen (?), nor it's particularly harmful (periodic error log for a good reason is not too bad idea) 🤔

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -32,7 +32,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	arv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
-	authv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -293,15 +292,14 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
 // custom resources and registers handlers with the webhook server.
 func (o *Operator) setupAdmissionWebhooks(ctx context.Context) error {
 	// Write provided cert files.
-	caBundle, err := o.ensureCerts(ctx, o.manager.GetWebhookServer().(*webhook.DefaultServer).Options.CertDir)
+	caBundle, err := o.ensureCerts(o.manager.GetWebhookServer().(*webhook.DefaultServer).Options.CertDir)
 	if err != nil {
 		return err
 	}
 
-	if updateAllowed, err := o.canUpdateWebhooks(ctx); err != nil {
-		return err
-	} else if updateAllowed {
-		// Keep setting the caBundle in the expected webhook configurations.
+	if len(caBundle) > 0 {
+		// Keep setting the caBundle, if "ensureCerts" gives us those, in the expected webhook configurations.
+		// In case of not enough permissions we will keep trying with error message.
 		go o.continuouslySetCABundle(ctx, caBundle)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/operator/operator.go
```diff
@@ -293,15 +292,14 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
 // custom resources and registers handlers with the webhook server.
 func (o *Operator) setupAdmissionWebhooks(ctx context.Context) error {
 	// Write provided cert files.
-	caBundle, err := o.ensureCerts(ctx, o.manager.GetWebhookServer().(*webhook.DefaultServer).Options.CertDir)
+	caBundle, err := o.ensureCerts(o.manager.GetWebhookServer().(*webhook.DefaultServer).Options.CertDir)
 	if err != nil {
 		return err
 	}
 
-	if updateAllowed, err := o.canUpdateWebhooks(ctx); err != nil {
```
bwplotka, line 301: So I proposed removing this step. Rationales:

* It's clear this code is tricky to maintain, we still for some reason check wrongly here, so removing this complexity would be ideal.
* Do we need to support case of NOT injecting periodically caBundle? If yes, perhaps flag would help?

bwplotka: I don't know the history of this step. Especially I am not aware of any use case where somebody else would manually set caBundle 🤔

bwplotka: So before https://github.com/GoogleCloudPlatform/prometheus-engine/pull/661 we didn't have this check. Sounds like the intention was to have one check instead of failing indefinitely, but did we see ever that happening or harmful @bernot-dev @TheSpiritXIII ? 🤔 

I think it should be safe to have periodic errors in cases of perms error.

pkg/operator/operator.go
```diff
@@ -557,58 +555,17 @@ func (o *Operator) setMutatingWebhookCABundle(ctx context.Context, caBundle []by
...
-		return
-	}
 	// Initial sleep for the client to initialize before our first calls.
 	// Ideally we could explicitly wait for it.
 	time.Sleep(5 * time.Second)
 
 	for {
 		if err := o.setValidatingWebhookCABundle(ctx, caBundle); err != nil {
-			o.logger.Error(err, "Setting CA bundle for ValidatingWebhookConfiguration failed")
+			o.logger.Error(err, "Setting CA bundle for ValidatingWebhookConfiguration failed; retrying in 1m...")
```
TheSpiritXIII, line 565: 1m or 5s?

pkg/operator/operator.go
```diff
@@ -293,15 +292,14 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
...
 		return err
 	}
 
-	if updateAllowed, err := o.canUpdateWebhooks(ctx); err != nil {
-		return err
-	} else if updateAllowed {
-		// Keep setting the caBundle in the expected webhook configurations.
+	if len(caBundle) > 0 {
+		// Keep setting the caBundle, if "ensureCerts" gives us those, in the expected webhook configurations.
+		// In case of not enough permissions we will keep trying with error message.
```
TheSpiritXIII, line 302: Great find!

pkg/operator/operator.go
TheSpiritXIII: Your reasoning makes sense. Great investigation! I like how this code looks simpler.

pkg/operator/operator.go
bwplotka: 1m, see time.After below (:

macxamin: 1m because of line 573

pkg/operator/operator.go
bwplotka: More context from @bernot-dev :

> #661 is part of the work necessary to handle the case where certs will be provided by GKE and the bundle will not need to be set.

Let's iterate over that in the next PRs then (with more tests, bit more commentary on the intended behavior). One idea is to maybe decide about this behavior based on some extra annotation/label/operator config instead of permission? The disadvantage of perm based deciding is that we might miss perms misconfiguration cases easily. Anyway, let's iterate!

PR #717, opened 2023-12-15
Improve e2e test reliability when running on GitHub Actions

TheSpiritXIII:
- Increases timeouts for a few tests.
- Ensures we don't poll too regularly and accidentally overload the Kubernetes API server.
- Ensures that we can recover when we get an error due to the Kubernetes API server being overloaded (e.g. try to fetch an object and get a timeout).
- Waits until collectors are up before validating the PodMonitoring status.
- Increases e2e timeout on GitHub Actions.

Why: I am adding more tests and without this change, the tests are very flaky.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -62,3 +62,5 @@ jobs:
     - uses: actions/checkout@v3
     - name: Run e2e
       run: make e2e
+      env:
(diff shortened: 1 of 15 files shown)


bwplotka (approved):
Thanks! We could try to shard tests too, but probably hard without running multiple kind clusters.

PR #727, opened 2023-12-20
feat: [kubelet/cAdvisor metrics package] add frontend for NodeMonitoring

macxamin:
Add the frontend part of NodeMonitoring and make port optional.

Tested via running the following commands:
1. `DOCKER_PUSH=1 make operator` 
2. `kubectl apply -f manifests/setup.yaml`
3. `kubectl apply -f manifests/operator.yaml`
4. `kubectl apply -f examples/cadvisor-metrics.yaml` and `kubectl apply -f examples/kubelet-metrics.yaml` 

You can see the node metrics being scraped after that. Additionally I toggled KubeletScraping on/off to watch the scraping turn off/on.

### DIFF
--- a/pkg/operator/collection.go
+++ b/pkg/operator/collection.go
@@ -96,6 +96,12 @@ func setupCollectionControllers(op *Operator) error {
 			enqueueConst(objRequest),
 			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
 		).
+		// Any update to a NodeMonitoring requires regenerating the config.
+		Watches(
+			&monitoringv1.NodeMonitoring{},
+			enqueueConst(objRequest),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
+		).
 		// The configuration we generate for the collectors.
 		Watches(
 			&corev1.ConfigMap{},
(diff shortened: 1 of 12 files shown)


pkg/operator/collection.go
```diff
@@ -462,6 +469,41 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
...
+	}
+	// The following job names are reserved by GMP for NodeMonitoring. They will not be generated if kubeletScraping is enabled.
+	var (
+		reservedCAdvisorJobName = "gmp-cadvisor-metrics"
+		reservedKubeletJobName  = "gmp-kubelet-metrics"
+	)
+	// Mark status updates in batch with single timestamp.
+	for _, nm := range NodeMons.Items {
+		if spec.KubeletScraping != nil && (nm.Name == reservedKubeletJobName || nm.Name == reservedCAdvisorJobName) {
+			logger.Info("NodeMonitoring job %s not applied because KubeletScraping is enabled.", nm.Name)
```
bwplotka, line 483:
```suggestion
			logger.Info("NodeMonitoring job %s not applied because OperatorConfig.collector.kubeletScraping is enabled.", nm.Name)
```

Perhaps to be precise?

pkg/operator/collection.go
```diff
@@ -462,6 +469,41 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 		}
 	}
 
+	if err := r.client.List(ctx, &NodeMons); err != nil {
+		return nil, fmt.Errorf("failed to list NodeMonitorings: %w", err)
+	}
+	// The following job names are reserved by GMP for NodeMonitoring. They will not be generated if kubeletScraping is enabled.
```
bwplotka, line 475: Hm, can we add a safeguard around namespaces? One could be surprised that certain names are blocked, although rare

bwplotka (approved):
Nice, LGTM modulo tiny non-blocking suggestions.

pkg/operator/collection.go
macxamin: thanks! good idea!

PR #732, opened 2023-12-21
Fix example-app OAuth2 should support POST requests

TheSpiritXIII:
Prometheus uses POST requests so this is a prerequisite for writing e2e tests that utilize the example application.


bwplotka (approved):
Thanks!

PR #733, opened 2023-12-21
Remove NodeMonitoring port

macxamin:
Remove NodeMonitoring port.
It seems Kubelet just has [one port](https://pkg.go.dev/k8s.io/api/core/v1#NodeDaemonEndpoints) that is used by [Prometheus as the default](https://github.com/GoogleCloudPlatform/prometheus/blob/release-2.41.0-gmp/discovery/kubernetes/node.go#L196).

I suggest we remove port support to simplify our code until we have a good reason to add it.

@bwplotka @pintohutch do you see any reason to keep NodeMonitoring port?


bwplotka:
To be explicit -- Prometheus SD configuration port ["defaults" to Kubelet port (which is typically only one)](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#nod). This does not mean somebody might have different workloads on node with different port. For this reason it would make sense to keep port.

However, you are right we don't have any specific ask around it. 

At the end... up to you. Personally I would keep it because there is little harm, but if you want we can kill it for now and add later if needed (: Good call.

bwplotka (approved):

PR #742, opened 2024-01-02
Split types.go into multiple files

TheSpiritXIII:
Since the file was becoming unwieldy...


pintohutch:
My vote would be to keep a single `types.go` with all the canonical structs and client generation tags, but move the individual functions to separate files.

TheSpiritXIII:
Thanks for your insight, @pintohutch! Can you elaborate on some reasons why you think we should do it this way?

My reasons for not doing it this way are:
1. controller-gen does not care what the file is called. It just scans for the annotations. Currently only `po-docgen` needs filenames, but this tool is deprecated and we should move off of it at some point.
2. This is the same way other projects split up their types. For instance, a sample GitHub search shows that nearly everyone does a "_types.go" file suffix: https://github.com/search?q=%2Bkubebuilder%3Asubresource%3Astatus&type=code
3. It feel unconventional to have functions in separate files.

TheSpiritXIII:
Had an offline discussion with @pintohutch who okayed this change, as long as controller-generators still work (as part of #738).

doc/api.md
```diff
@@ -17,339 +17,375 @@ This Document documents the types introduced by the GMP CRDs to be consumed by u
 > Note this document is generated from code comments. When contributing a change to this document please do so by changing the code comments.
 
 ## Table of Contents
-* [AlertingSpec](#alertingspec)
-* [AlertmanagerEndpoints](#alertmanagerendpoints)
-* [Auth](#auth)
-* [Authorization](#authorization)
-* [BasicAuth](#basicauth)
 * [ClusterPodMonitoring](#clusterpodmonitoring)
```
bwplotka, line 20: Hm, this is bit brutal consequence. It is definitely easier for users to read this when structs are sorted lex. I guess we could do some post processing too, but maybe there is a way to contribute/configure generator? 🤔

bwplotka (commented):
I don't mind both ways. If @pintohutch agrees, I am fine with split too. I like some separation.

However one small blocker is documentation change (ordering), is there a way we could maintain old order of structs? 🤔

doc/api.md
TheSpiritXIII: The tool we use here was deleted by the maintainers. We currently use `po-docgen`, which was originally developed by prometheus-operator. They have since switched to `gen-crd-api-reference-docs`. However, if you go to the `gen-crd-api-reference-docs` README, it states that it will not be maintained anymore either, with 3 possible alternatives. We should switch to a different tool but this is more work that I'm not getting myself into yet.

TheSpiritXIII: Address in #756, which this branch was rebased off of. Please review that PR first!

TheSpiritXIII:
fyi, renamed collection_types to pod_types, and created monitoring_types for common status objects that would be used by podmonitoring, nodemonitoring and probemonitoring!

macxamin (approved):
LGTM

PR #743, opened 2024-01-03
[E2E refactor - 1] slim down Dockerfiles

pintohutch:

This is the first one, where we slim down our Docker images to only copy code that's necessary to build and run the binary or script.


bwplotka (approved):
Love it, thanks!

PR #746, opened 2024-01-03
[E2E refactor - 4] refactor alertmanager e2e tests

pintohutch:

This is the fourth one, where we refactor and simplify alertmanager e2e tests.


e2e/alertmanager_test.go
```diff
@@ -20,188 +20,172 @@ import (
...
+func testAlertmanagerDeployed(ctx context.Context, t *testing.T, kubeClient kubernetes.Interface) func(*testing.T) {
+	return func(t *testing.T) {
+		t.Log("checking alertmanager is running")
+
+		err := wait.PollUntilContextCancel(ctx, pollDuration, false, func(ctx context.Context) (bool, error) {
+			ss, err := kubeClient.AppsV1().StatefulSets(operator.DefaultOperatorNamespace).Get(ctx, operator.NameAlertmanager, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			} else if err != nil {
+				return false, fmt.Errorf("getting alertmanager StatefulSet failed: %w", err)
```
bernot-dev, line 55: Avoid using "stop words" like "failed" when wrapping errors.
go/gotip/episodes/100

Maybe just, `"get alertmanager: %w"` or something.

e2e/alertmanager_test.go
```diff
@@ -20,188 +20,172 @@ import (
...
 route:
   receiver: "foobar"
 `
-	testAlertmanager(context.Background(), tctx, nil,
-		&corev1.Secret{
+		secret := &corev1.Secret{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      operator.AlertmanagerPublicSecretName,
-				Namespace: tctx.pubNamespace,
+				Name: "my-secret-name",
```
bernot-dev, line 110: question: What is the advantage of the literal here, over the constant `operator.AlertmanagerPublicSecretName`?

e2e/alertmanager_test.go
```diff
@@ -20,188 +20,172 @@ import (
...
-			return true, nil
-		}
-
-		const containerName = "alertmanager"
-		// TODO(pintohutch): clean-up wantArgs init logic.
-		var wantArgs []string
-		for _, c := range ss.Spec.Template.Spec.Containers {
-			if c.Name != containerName {
-				continue
+		err = wait.PollUntilContextCancel(ctx, pollDuration, false, func(ctx context.Context) (bool, error) {
```
bernot-dev, line 143: suggestion: consider a refactor to pull out the func that gets polled here. This gets to be a lot of nesting, especially once you get to the conditionals inside the `for` loop, inside this closure, inside another closure, inside the test helper function.

e2e/operator_context_test.go
```diff
@@ -116,7 +112,7 @@ func setRESTConfigDefaults(restConfig *rest.Config) error {
 
 // TestMain injects custom flags and adds extra signal handling to ensure testing
 // namespaces are cleaned after tests were executed.
-func TestMain(m *testing.M) {
+func OldTestMain(m *testing.M) {
```
bernot-dev, line 115: Question: Is this needed? Is there a better name for it?

pkg/operator/operator_config.go
```diff
@@ -66,17 +66,19 @@ const (
 	AlertmanagerPublicSecretKey  = "alertmanager.yaml"
 	rulesDir                     = "/etc/rules"
 	secretsDir                   = "/etc/secrets"
-	alertmanagerConfigKey        = "config.yaml"
+	AlertmanagerConfigKey        = "config.yaml"
```
bernot-dev, line 69: Suggestion: I like to use the [test-only export trick](https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd) when something is needed for tests, but shouldn't actually be part of the Public API.

bernot-dev (approved):
Nothing blocking here, just a few suggestions. LGTM.

e2e/alertmanager_test.go
pintohutch: Done.

e2e/alertmanager_test.go
pintohutch: The `wait.PollUntilContextCancel` function signature basically requires a closure, as the `wait.ConditionWithContextFunc` is a closure that only takes in a `context.Context`. Given we almost always want to fetch something from the `kubeClient`, the variable would need to be declared outside of the function call, leading to this pattern.

Given this function is provided by the k8s apimachinery library, I'm going to leave this pattern as-is. If we want to leverage a different utility in the future for our k8s tests, we can discuss then.

e2e/operator_context_test.go
pintohutch: This is temporary to avoid duplicate `TestMain` functions in this package. I didn't want to completely break everything, so just renamed. This file is deleted entirely in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/752.

pkg/operator/operator_config.go
pintohutch: Neat! I'm inclined to have a dedicated commit for that so we can establish that pattern holistically instead of in this chain of PRs. But I like the idea :)

PR #747, opened 2024-01-03
[E2E refactor - 5] refactor ruler e2e tests

pintohutch:

This is the fifth one, where we refactor and simplify rule-evaluator e2e tests.


e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
+			t.Fatalf("get operatorconfig: %s", err)
+		}
+
+		// Configure
+		config.Rules.ExternalLabels = map[string]string{
+			"external_key": "external_val"}
+		config.Rules.GeneratorURL = "http://example.com/"
+
+		_, err = opClient.MonitoringV1().OperatorConfigs(operator.DefaultPublicNamespace).Update(ctx, config, metav1.UpdateOptions{})
+		if err != nil {
```
bwplotka, line 110:
```suggestion
		if _, err := opClient.MonitoringV1().OperatorConfigs(operator.DefaultPublicNamespace).Update(ctx, config, metav1.UpdateOptions{}); err != nil {
```

e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
-							},
-							KeySecret: keySecret,
-						},
-					},
-				},
-			},
-		},
-	})
-}
+		_, err = kubeClient.CoreV1().Secrets(operator.DefaultPublicNamespace).Create(ctx, tlsPair, metav1.CreateOptions{})
```
bwplotka, line 179: all of it has opportunity for simplification (if _, err := ... ; err != nil ), will stop commenting on this one (:

e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
+		want := map[string][]byte{
+			fmt.Sprintf("secret_%s_alertmanager-tls_cert", operator.DefaultPublicNamespace):            cert,
+			fmt.Sprintf("secret_%s_alertmanager-tls_key", operator.DefaultPublicNamespace):             key,
+			fmt.Sprintf("secret_%s_alertmanager-authorization_token", operator.DefaultPublicNamespace): []byte("auth-bearer-password"),
+		}
+		err = wait.PollUntilContextCancel(ctx, pollDuration, false, func(ctx context.Context) (bool, error) {
+			secret, err := kubeClient.CoreV1().Secrets(operator.DefaultOperatorNamespace).Get(ctx, operator.RulesSecretName, metav1.GetOptions{})
 			if apierrors.IsNotFound(err) {
 				return false, nil
+			} else if err != nil {
```
bwplotka, line 201: ditto (else not needed)

e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
+			}
+			return true, nil
+		})
+		if err != nil {
+			t.Fatalf("failed waiting for generated rule-evaluator deployment: %s", err)
+		}
+	}
+}
+
+func testRuleEvaluatorConfigured(ctx context.Context, t *testing.T, kubeClient kubernetes.Interface, opClient versioned.Interface) func(*testing.T) {
```
bwplotka, line 95: Hm, the name `testRuleEvaluatorConfigured` and `testRuleEvaluatorConfiguration` are a bit too close, it's hard to understand from function name what's for what. This also indicated the problem of the bit duplicated purpose of those. Can we have one where we ensure OperatorConfig has a custom change and we check rule-eval Prometheus like config?

bwplotka (approved):
Nice, LGTM mod some suggestions

e2e/ruler_test.go
bwplotka: The same applies to all other tests in other PRs. A bit confusing 🤔

e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
-		t.Run("check rule metrics", tctx.subtest(testValidateRuleEvaluationMetrics))
+		t.Run("rules-gcm", testValidateRuleEvaluationMetrics(ctx, t))
+	}
+}
+
+func testRuleEvaluatorDeployed(ctx context.Context, t *testing.T, kubeClient kubernetes.Interface) func(*testing.T) {
+	return func(t *testing.T) {
+		t.Log("checking rule-evaluator is running")
+
+		err := wait.PollUntilContextCancel(ctx, pollDuration, false, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
pintohutch: Yea - I was actually a big proponent of that pattern, but it wasn't convention in the repo so I kept with what was here. I would rather have that change in a dedicated commit, where we agree on the convention as a team before doing it here.

e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
+			deploy, err := kubeClient.AppsV1().Deployments(operator.DefaultOperatorNamespace).Get(ctx, operator.NameRuleEvaluator, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			} else if err != nil {
```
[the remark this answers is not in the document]
pintohutch: Well we need to check if `err` is not nil after we've validated it's not a "not found" error. If so, then something else is wrong.

e2e/ruler_test.go
pintohutch: Yea - similar to my comment above.

e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
+		config.Rules.ExternalLabels = map[string]string{
+			"external_key": "external_val"}
+		config.Rules.GeneratorURL = "http://example.com/"
+
+		_, err = opClient.MonitoringV1().OperatorConfigs(operator.DefaultPublicNamespace).Update(ctx, config, metav1.UpdateOptions{})
+		if err != nil {
+			t.Fatalf("update operatorconfig: %s", err)
+		}
+		// Keep checking the state of the collectors until they're running.
+		err = wait.PollUntilContextCancel(ctx, pollDuration, false, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
pintohutch: same with my comment above

e2e/ruler_test.go
```diff
@@ -25,174 +25,212 @@ import (
...
+			deploy, err := kubeClient.AppsV1().Deployments(operator.DefaultOperatorNamespace).Get(ctx, operator.NameRuleEvaluator, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			} else if err != nil {
```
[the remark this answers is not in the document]
pintohutch: same with my comment above

e2e/ruler_test.go
pintohutch: sgtm :)

e2e/ruler_test.go
pintohutch: Agreed. Changed the name along with the new pattern from collector_test.go to make this more clear.

PR #749, opened 2024-01-03
feat: [kubelet/cAdvisor metrics package] add NodeMonitoring Status support

macxamin:


pkg/operator/collection.go
```diff
@@ -503,6 +503,42 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
...
+		change, err := nm.Status.SetMonitoringCondition(nm.GetGeneration(), metav1.Now(), cond)
+		if err != nil {
+			// Log an error but let operator continue to avoid getting stuck
+			// on a potential bad resource.
+			logger.Error(err, "setting nodemonitoring status state")
+		}
+
+		if change {
+			// convert ScrapeNodeEndpoint to ScrapeEndpoint.
+			var scrapeEndpoints []monitoringv1.ScrapeEndpoint
```
macxamin, line 516: Do you have any better ideas on how to use NodeMonitoringCRD and PodMonitoringCRD in statusUpdates?

pkg/operator/apis/monitoring/v1/types_test.go
```diff
@@ -1265,19 +1265,19 @@ func TestSetPodMonitoringCondition(t *testing.T) {
 		cond       *MonitoringCondition
 		generation int64
 		now        metav1.Time
-		curr, want *PodMonitoringStatus
+		curr, want *MonitoringStatus
```
TheSpiritXIII, line 1268: nit: maybe also rename function (remove the `Pod` prefix)?

pkg/operator/collection.go
```diff
@@ -503,6 +503,42 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
...
+				Spec: monitoringv1.PodMonitoringSpec{
+					Selector:  nm.Spec.Selector,
+					Endpoints: scrapeEndpoints,
+					Limits:    nm.Spec.Limits,
+				},
+				Status: monitoringv1.PodMonitoringStatus{
+					MonitoringStatus: nm.Status,
+				},
+			}
+			r.statusUpdates = append(r.statusUpdates, &nodeMonConv)
```
TheSpiritXIII, line 540: issue: I'm not sure if this is right. `statusUpdates` is for updating statuses of _existing_ [Cluster]PodMonitorings but one doesn't exist here. I believe your intention was to pass the NodeMonitoring object here.  An e2e test would have caught this.

I would rename `PodMonitoringCRD.GetStatus` to something like `PodMonitoringCRD.GetPodMonitoringStatus` and create a new interface with a method `MonitoringCRD.GetStatus` that returns a `MonitoringStatus`. Then, `statusUpdates` can store a `MonitoringCRD` and you'd be able to pass in your `NodeMonitoring` here.

TheSpiritXIII (changes requested):
Just a few nits and an issue to look at. Great job otherwise!

macxamin:
Closing, this is no longer relevant after the types.go split and https://github.com/GoogleCloudPlatform/prometheus-engine/pull/771.

PR #751, opened 2024-01-03
[E2E refactor - 7] refactor authorization e2e test

pintohutch:

This is the seventh one, where we refactor and simplify authorization e2e tests.


pintohutch:
Yea - here you just run `TEST_RUN=TestBasicAuthClusterPodMonitoring make e2e` to get the desired affect.

I did it this way, against some [convention](https://go.dev/blog/subtests) for two reasons:
1. Readability and reusability - each `Test` has clear steps with comments, and each flavor of test is separated out for clear distinction. Obviously, the main drawback here is DRY - but I lean towards [repeating oneself](https://groups.google.com/g/golang-nuts/c/xZCIoPNs07k/m/Ca5O5vplEK4J) if it makes the code easier to follow - particularly in Go, where this is more conventional. Additionally, the subtests run sequentially and are not marked with `t.Parallel()`. 
2. Parallelism in GH Actions. The conventional way would essentially recommend bundling everything in one `TestXXX` function, in-line your set and teardown, and sandwich in your subtests do the various flavors. This works in some cases, and doesn't work in others. For example, in our case, where we need to patch the `example-app` with specific flags for each subtest. Doing this in one `Test` function increases build times and can cause timeout issues against the K8s API server.

If we really wanted to force convention for our use case, we could look into changing the way we [build](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/753/files#diff-345bb67bae17c6d0352a24685e91206cfe1b40e6420004d3bfdd3e193d798c09R66) our `TEST_RUN` matrix to list all subtests as well. However, I'm not sure how to do this.

TheSpiritXIII (approved):
Approving this for now. Let's keep some of the drawbacks in mind and maybe we'll find a better solution for a subtest-supported test matrix in the future!

pintohutch:
Thanks for the feedback - that SGTM!

PR #753, opened 2024-01-04
[E2E refactor - 9] use matrix strategy to parallelize github action e2e test runs

pintohutch:

This is the ninth one, where we use parallel Github Actions to run each kind cluster test in parallel for faster and more reliable tests (< 10m total).


TheSpiritXIII (approved):
LGTM!

PR #754, opened 2024-01-04
[E2E refactor - 10] move e2e utils to dedicated pkg for future use

pintohutch:

This is the tenth (and final!) one, where we move util packages out of `e2e/` to a dedicated library in managed collection for future use.


bwplotka (changes requested):
Thanks! So many utils!

image

What exactly future use you have in mind? Will we ever use those outside of e2e test? If not I would keep them in e2e honestly, otherwise it suggests it should be used in prod code.

The util name also means nothing, it would epic to at least have only one "util" e.g. `e2e/kubeutil/...` and `e2e/operatorutil/<files without util suffix>` unless we literally import those from outside of e2e package. (maybe I am missing something).

BTW I am surprised you didn't have to change any caller references? Does it mean those are not used? The let's remove for now (we can always bring those back via git).

bwplotka (approved):
LGTM, I would say let's work with @TheSpiritXIII to create a dedicated repo (library) for those utils, they are super useful, just not at the moment. 

Let's figure the name of it and maintain there, seriously (:

TheSpiritXIII:
Let's nuke it. I know where to look if we ever need these utilities back.

TheSpiritXIII (approved):

pintohutch:
Thanks all!

PR #755, opened 2024-01-04
[E2E refactor - 11] fix go-synthetic e2e deploy and reduce local parallelism

pintohutch:

This is the eleventh (and really final!) one, where we fix a bug that allows a build and deploy of the go-synthetic container from source. Also reduce the parallelism for local e2e test runs to improve stability.


bwplotka (approved):
👍🏽

PR #756, opened 2024-01-04
Switch to gen-crd-api-reference-docs

TheSpiritXIII:
`po-docgen` has been deprecated and deleted so we moved to the newer [gen-crd-api-reference-docs](https://github.com/ahmetb/gen-crd-api-reference-docs), which is also used in prometheus-operator.

This tool is slightly more powerful, as it supports templating. I made a few changes from the default templates to ensure old hyperlinks would still work:
- Show all types in the package.
- Allow links without the package prefix in URLs.
- Don't inline tables within tables.


bwplotka (approved):
Thanks! 💪🏽

PR #760, opened 2024-01-08
Cherry-picked datasource-syncer for 0.9.1 release

bwplotka:


macxamin (approved):
Thanks!

pintohutch (changes requested):
Actually - seems there's a presubmit failing. Can you fix? `make regen` should do it I think?

bwplotka:
I was discussing this with @macxamin, the reason was convoluted, `go mod tidy` now bumps our go mod file to 1.21 because grafana deps requires this, but this branch Ci tooling does not support 1.21 yet. Checking today what's wrong

PR #768, opened 2024-01-12
Add TLS to datasource-syncer

macxamin:
1. Added TLS support, including four new flags `tls-cert`, `tls-key`, `tls-ca-cert` and `insecure-skip-verify`.
2. Added a condition to return error and exist if any data source syncs fail.
3. Added a label to datasource-syncer.yaml so users can use `kubectl get pods -l app=datasource-syncer` to view pods associated with their job.

### DIFF
--- a/cmd/datasource-syncer/main.go
+++ b/cmd/datasource-syncer/main.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"flag"
 	"fmt"
 	"net/http"
@@ -32,6 +34,11 @@ var (
 
 	gcmEndpointOverride = flag.String("gcm-endpoint-override", "",
 		"gcm-endpoint-override is the URL where queries should be sent to from Grafana. This should be left blank in almost all circumstances.")
+
+	certFile           = flag.String("tls-cert", "", "Path to the server TLS certificate.")
+	keyFile            = flag.String("tls-key", "", "Path to the server TLS key.")
+	caFile             = flag.String("ca-cert", "", "Path to the server certificate authority")
+	InsecureSkipVerify = flag.Bool("insecure-skip-verify", false, "Skip TLS certificate verification")
 )
 
 func main() {
@@ -60,7 +67,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	grafanaClient, err := grafana.New(*grafanaEndpoint, grafana.Config{APIKey: *grafanaAPIToken})
+	client, err := getTLSClient(logger, *certFile, *keyFile, *caFile, *InsecureSkipVerify)
+	if err != nil {
+		level.Error(logger).Log("msg", "couldn't create client", "err", err)
+		os.Exit(1)
+	}
+
+	grafanaClient, err := grafana.New(*grafanaEndpoint, grafana.Config{
+		APIKey: *grafanaAPIToken,
+		Client: client,
+	})
 	if err != nil {
 		level.Error(logger).Log("msg", "couldn't create grafana client", "err", err)
 		os.Exit(1)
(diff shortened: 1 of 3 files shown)


cmd/datasource-syncer/main.go
```diff
@@ -32,6 +34,11 @@ var (
 
 	gcmEndpointOverride = flag.String("gcm-endpoint-override", "",
 		"gcm-endpoint-override is the URL where queries should be sent to from Grafana. This should be left blank in almost all circumstances.")
+
+	certFile           = flag.String("tls-cert", "", "Path to the server TLS certificate.")
+	keyFile            = flag.String("tls-key", "", "Path to the server TLS key.")
+	caFile             = flag.String("ca-cert", "", "Path to the server certificate authority")
```
TheSpiritXIII, line 40: nit: name it `tls-ca-cert`?

cmd/datasource-syncer/main.go
```diff
@@ -32,6 +34,11 @@ var (
...
+	InsecureSkipVerify = flag.Bool("insecure-skip-verify", false, "Skip TLS certificate verification")
```
TheSpiritXIII, line 41: nit: lowercase `insecureSkipVerify`?

cmd/datasource-syncer/main.go
```diff
@@ -222,3 +238,43 @@ func buildUpdateDataSourceRequest(dataSource grafana.DataSource, token string) (
...
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+		tlsConfig.RootCAs = caCertPool
+	}
+
+	transport := &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
+
+	client = &http.Client{Transport: transport}
```
TheSpiritXIII, line 278: suggestion: you should use the default client and update the transport.

cmd/datasource-syncer/main.go
```diff
@@ -222,3 +238,43 @@ func buildUpdateDataSourceRequest(dataSource grafana.DataSource, token string) (
...
 	return &dataSource, nil
 }
+
+func getTLSClient(logger log.Logger, certFile, keyFile, caFile string, insecureSkipVerify bool) (*http.Client, error) {
+	if (certFile != "" || keyFile != "") && (certFile == "" || keyFile == "") {
+		return nil, fmt.Errorf("--tls-cert and tls-key must both be set or unset")
+	}
+
+	if certFile == "" && keyFile == "" && caFile == "" && !insecureSkipVerify {
+		return http.DefaultClient, nil
```
TheSpiritXIII, line 248: suggestion: the grafana library actually uses https://github.com/hashicorp/go-cleanhttp. Alternatively, just return nil and let the library decide the defaults.

cmd/datasource-syncer/main.go
```diff
@@ -222,3 +238,43 @@ func buildUpdateDataSourceRequest(dataSource grafana.DataSource, token string) (
...
+		caCert, err := os.ReadFile(caFile)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read ca cert: %w", err)
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+		tlsConfig.RootCAs = caCertPool
+	}
+
+	transport := &http.Transport{
```
TheSpiritXIII, line 274: suggestion: use default transport and update its values.

TheSpiritXIII (changes requested):
Looks good! Just left some minor nits and suggestions.

cmd/datasource-syncer/main.go
macxamin: breaks code.

cmd/datasource-syncer/main.go
macxamin: breaks code.

cmd/datasource-syncer/main.go
TheSpiritXIII: Can you put this in a comment so we understand why we didn't use it?

TheSpiritXIII (approved):
Thank you!

PR #776, opened 2024-01-22
feat: support secrets in scrape authorization

### LINKED ISSUE #450: Protected scrape endpoints

TheSpiritXIII:
Fixes #450 and #241.

This PR adds the following types to `PodMonitoring`/`ClusterPodMonitoring`'s `ScrapeEndpoint`:
- `scrapeEndpoint.tls.ca`
- `scrapeEndpoint.tls.cert`
- `scrapeEndpoint.tls.key`
- `scrapeEndpoint.oauth2.clientSecret`
- `scrapeEndpoint.auth.credentials`
- `scrapeEndpoint.basicAuth.password`

These are all the same object pointing to a Kubernetes Secret with these fields:
```
secret.name
secret.key
secret.namespace
```


pkg/operator/apis/monitoring/v1/pod_types.go
```diff
@@ -250,16 +262,26 @@ func (pm *PodMonitoring) endpointScrapeConfig(index int, projectID, location, cl
 	}
 
 	return endpointScrapeConfig(
-		pm.GetKey(),
+		pm,
```
pintohutch, line 265: thought (non-blocking): this is a bit of a code-smell

We're running in an instance method, calling a pure helper function, and passing the instance to it. This feel like it should just be flattened in-line to avoid the confusion.

Maybe part of this stems from the fact that we're passing in a mutable argument at the top-level with the `pool PrometheusSecretConfigs` object.

This works - so I'm not here to block it being merged, but I do have mild concerns about long-term maintainability of this for others without context.

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -79,33 +197,45 @@ func TLSVersionFromString(s string) (config.TLSVersion, error) {
...
-	}, nil
+	tls.CertRef, err = c.Cert.toPrometheusSecretRef(m, pool)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	tls.KeyRef, err = c.Key.toPrometheusSecretRef(m, pool)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	return tls, errors.Join(errs...)
```
pintohutch, line 227: suggestion: Why would we not short-circuit error handling earlier, as done previously?

Doing so has nice advantages, including:
1. Prompter code execution
2. Returned error context is more relevant
3. Callers can check the specific error type
4. Reduce risk of propagating error or mishandling "fall-through" code
5. More conventional in golang (and in this repo)

Same throughout other places where this pattern is coming in...

pintohutch (commented):
This review is coupled in a lot of ways with https://github.com/GoogleCloudPlatform/prometheus/pull/130, so forgive any overlapping comments

pkg/operator/apis/monitoring/v1/pod_types.go
pintohutch: Yeah - I can see how we got here.

Addressing this would ultimately be a larger refactoring, which isn't necessary or appropriate in this PR. Thanks for explaining!

pintohutch (commented):
Could you give some more detail in the PR description about what fields are being added?

There's a lot here and it would be nice to have a summary to refer back to so we understand the user-facing changes.

pkg/operator/apis/monitoring/v1/http_types.go
TheSpiritXIII: I'm inclined to say we should measure the impact on a case-by-case basis. In this case, the odds of a panic from fall-through or bad performance are low because we're only validating and building a configuration object.

Despite this pattern being introduced in 2023, it's widely used internally: http://cs/%22%5B%5Derror%22%20%22errors.Join%22

And while there are limited usages in the official Golang and Kubernetes, they are used.

Also, we've been using this pattern in this repository in quite a few places already!

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,142 @@ import (
...
+
+	return s.KubernetesSecret.toPrometheusSecretRef(m, pool)
+}
+
+// KubernetesSecretKeySelector represents selector for Kubernetes secret.
+// It's similar to k8s.io/api/core/v1.SecretKeySelector, but
+// allows cross namespace selections.
+type KubernetesSecretKeySelector struct {
+	// Name of the secret to select from.
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
TheSpiritXIII: I learn towards our own struct because:
1. `LocalObjectReference` only has 1 field. Not much value.
2. `LocalObjectReference` documentation is confusing. What type of "referent"?
3. We always inline it. Extra layer of indirection for no reason.
4. `LocalObjectReference` may add additional fields that we'd have to support or decide to replace this object. Hypothetically this type should be a "local" version of [`ObjectReference`](https://pkg.go.dev/k8s.io/api/core/v1?utm_source=godoc#ObjectReference), so I can envision them adding these fields. See [official discussions]( https://github.com/kubernetes-retired/service-catalog/issues/1411).
5. `corev1.SecretKeySelector` has an `optional` field which we'll never use. We're not using it in other places -- does this mean our implementation is incorrect/non-conforming?

TheSpiritXIII:
@pintohutch hello, I've responded to all of your comments. Can you please take another look and let me know your thoughts?

pkg/operator/apis/monitoring/v1/http_types.go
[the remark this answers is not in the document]
pintohutch: Sounds like you have strong opinions so SGTM

pkg/operator/apis/monitoring/v1/http_types.go
pintohutch: Non-blocking so SGTM

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,142 @@ import (
...
+	"github.com/prometheus/prometheus/secrets"
+	"github.com/prometheus/prometheus/secrets/kubernetes"
 )
 
-// Auth sets the `Authorization` header on every scrape request.
-//
-// Currently the credentials are not configurable and always empty.
+// SecretSelector represents a reference to a secret from the given provider
+// e.g. Kubernetes Secret. Only one provider can be used at the same time.
+type SecretSelector struct {
```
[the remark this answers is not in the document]
pintohutch: Great thanks for the explanation - resolving.

pintohutch (approved):
:shipit: !

bwplotka:
Nice! Will check it out, will try to deploy it too & test.

e2e/authorization_test.go
```diff
@@ -23,135 +23,207 @@ import (
...
+	ctx := context.Background()
+	kubeClient, opClient, err := setupCluster(ctx, t)
+	if err != nil {
+		t.Fatalf("error instantiating clients. err: %s", err)
+	}
+
+	t.Run("patch-example-app-args", testPatchExampleAppArgs(ctx, kubeClient, []string{
+		"--tls-create-self-signed=true",
+	}))
+	authorizationClusterPodMonitoringTest(ctx, t, kubeClient, opClient, "tls",
```
bwplotka, line 79: nit(non-blocking): There might be opportunity to make those table tests at some point: `TestTLSPodMonitoring` and `TestTLSClusterPodMonitoring` is only different by one thing - first is invoking `authorizationClusterPodMonitoringTest` second `authorizationPodMonitoringTest`

bwplotka: Similar to other tests for diff security types.

e2e/authorization_test.go
```diff
@@ -23,135 +23,207 @@ import (
...
+		"--basic-auth-password=pass",
+	}))
+	authorizationPodMonitoringTest(ctx, t, kubeClient, opClient, "basic-auth-no-username",
+		&monitoringv1.ScrapeEndpoint{
+			Port:     intstr.FromString("web"),
+			Interval: "5s",
+			HTTPClientConfig: monitoringv1.HTTPClientConfig{
+				BasicAuth: &monitoringv1.BasicAuth{
+					Password: &monitoringv1.SecretSelector{
+						KubernetesSecret: &monitoringv1.KubernetesSecretKeySelector{
```
bwplotka, line 175: hmmmmm, I thought we renamed to secret, should we rename the type quickly, while we are renaming CR field?

e2e/authorization_test.go
bwplotka: Another argument for table test is authorizationPodMonitoringTest (and cluster one) has a few arguments where you really have understand order and meaning of them. Table test would allow describing scenarios in explicit struct perhaps.

pkg/operator/apis/monitoring/v1/http_types.go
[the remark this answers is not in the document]
bwplotka: Had same question month ago and agree with @TheSpiritXIII , one line more allows us to support more secure password sources one day and the resulting struct is not too bad for users:

image

But there is 95% of chances we will never add any other field (that's fine)

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
+	// KubernetesSecret represents reference to a given key from certain Kubernetes Secret
+	// in a given namespace.
+	// +optional
+	KubernetesSecret *KubernetesSecretKeySelector `json:"secret,omitempty"`
```
bwplotka, line 34:
```suggestion
	Secret *KubernetesSecretKeySelector `json:"secret,omitempty"`
```

Let's keep the type explicit, but here we could keep it same as field name to avoid confusion. We are configuring Kubernetes resource so "secret" meaning Kubernetes secret makes sense.

Notably, the API description is misleading (secret vs KubernetesSecret) 

image

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
+	// Name of the secret to select from.
+	Name string `json:"name"`
+
+	// The key of the secret to select from. Must be a valid secret key.
+	Key string `json:"key"`
+
+	// Namespace of the secret to select from.
+	// If empty the parent resource namespace will be chosen.
+	// +optional
+	// +kubebuilder:field:scope=Cluster
```
bwplotka, line 62: sweeeet, nice job!

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
-// configured username.
-//
-// Currently the password is not configurable and always empty.
+// BasicAuth sets the `Authorization` header on every HTTP request with the
+// configured username and optional password.
 type BasicAuth struct {
 	// The username for authentication.
 	Username string `json:"username,omitempty"`
-	// TODO: Add password: https://github.com/GoogleCloudPlatform/prometheus-engine/issues/450
+	// Password references the Kubernetes secret's key with the password to use.
```
bwplotka, line 153:
```suggestion
	// Password references the secret with the password to use.
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
+
+	// Namespace of the secret to select from.
+	// If empty the parent resource namespace will be chosen.
+	// +optional
+	// +kubebuilder:field:scope=Cluster
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// PrometheusSecretConfigs allows quick gathering of SecretConfigs for Prometheus configuration.
+// NOTE(bwplotka): This could be removed depending on how upstream Prometheus would like us to reference the secrets.
```
bwplotka, line 67:
```suggestion
```
Perhaps not helpful anymore

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
+		Name:      s.Name,
+		Key:       s.Key,
+	})
+	return ref, nil
+}
+
+// Auth sets the `Authorization` header on every HTTP request.
 type Auth struct {
-	// The authentication type. Defaults to Bearer, Basic will cause an error.
+	// The authentication type. Defaults to Bearer.
```
bwplotka, line 127:
```suggestion
	// Type represents the authentication type. Defaults to Bearer.
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
+	// Basic will cause an error, as the BasicAuth object should be used instead.
 	Type string `json:"type,omitempty"`
-	// TODO: Add credentials: https://github.com/GoogleCloudPlatform/prometheus-engine/issues/450
+	// Credentials references the Kubernetes secret's key with the credentials
+	// (token) for the auth header to send along the request.
+	// Optional, as in previous resource versions we allowed no credentials.
```
bwplotka, line 132:
```suggestion
	// Credentials references the secret with the credentials
	// (token) for the auth header to send along the request.
	// Optional, as in previous resource versions we allowed no credentials.
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -67,6 +181,22 @@ type TLS struct {
 	// If unset, Prometheus will use Go default minimum version, which is TLS 1.2.
 	// See MinVersion in https://pkg.go.dev/crypto/tls#Config.
 	MaxVersion string `json:"maxVersion,omitempty"`
+
+	// CA references the Kubernetes secret's key with the CA certificate to
```
bwplotka, line 185:
```suggestion
	// CA references the secret with the CA certificate to
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -67,6 +181,22 @@ type TLS struct {
...
+	// validate API server certificate with.
+	// Optional, as in previous resource versions we allowed no credentials.
+	// +optional
+	CA *SecretSelector `json:"ca,omitempty"`
+	// Cert references the Kubernetes secret's key with the certificate (public
```
bwplotka, line 190:
```suggestion
	// Cert references the secret with the certificate (public
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -67,6 +181,22 @@ type TLS struct {
...
+	// key) for client cert authentication to the server.
+	// Optional, as in previous resource versions we allowed no credentials.
+	// +optional
+	Cert *SecretSelector `json:"cert,omitempty"`
+	// Key references the Kubernetes secret's key with the private key
```
bwplotka, line 195:
```suggestion
	// Key references the secret with the private key
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -79,33 +209,46 @@ func TLSVersionFromString(s string) (config.TLSVersion, error) {
...
+	return tls, errors.Join(errs...)
 }
 
 // OAuth2 is the OAuth2 client configuration.
-//
-// Currently the client secret is not configurable and always empty.
 type OAuth2 struct {
 	// Public identifier for the client.
 	ClientID string `json:"clientID"`
+	// ClientSecret references the Kubernetes secret's key with the client secret
```
bwplotka, line 247:
```suggestion
	// ClientSecret references the secret with the client secret
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
 }
 
-// BasicAuth sets the `Authorization` header on every scrape request with the
-// configured username.
-//
-// Currently the password is not configurable and always empty.
+// BasicAuth sets the `Authorization` header on every HTTP request with the
+// configured username and optional password.
 type BasicAuth struct {
 	// The username for authentication.
```
bwplotka, line 151:
```suggestion
	// Username specifies the username for authentication.
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
+	return &config.BasicAuth{
+		Username:    c.Username,
+		PasswordRef: ref,
+	}, nil
 }
 
-// TLS specifies TLS configuration parameters from Kubernetes resources.
+// TLS specifies TLS configuration used for HTTP requests.
 type TLS struct {
 	// Used to verify the hostname for the targets.
```
bwplotka, line 172:
```suggestion
	// ServerName is used to verify the hostname for the targets.
```

bwplotka: BTW shouldn't all of those be optional? 🤔  I don't remember now but why sometimes we put explicit +optional sometimes not (for optional fields). It feels the behaviour is that if message has any +optional, other fields are marked as required? 🤔 

image

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -20,40 +20,154 @@ import (
...
 	ServerName string `json:"serverName,omitempty"`
```
bwplotka, line 173: For 

`InsecureSkipVerify` we could do comment like `// InsecureSkipVerify disables target certificate validation.

bwplotka: and so on to start comment by field name (non-blocking, I know we have this like that in Prometheus config, but perhaps worth to improve?)

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -79,33 +209,46 @@ func TLSVersionFromString(s string) (config.TLSVersion, error) {
...
+		errs = append(errs, err)
+	}
+	return tls, errors.Join(errs...)
 }
 
 // OAuth2 is the OAuth2 client configuration.
-//
-// Currently the client secret is not configurable and always empty.
 type OAuth2 struct {
 	// Public identifier for the client.
```
bwplotka, line 245:
```suggestion
	// ClientID represents a public identifier for the client.
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -79,33 +209,46 @@ func TLSVersionFromString(s string) (config.TLSVersion, error) {
...
 	ClientID string `json:"clientID"`
+	// ClientSecret references the Kubernetes secret's key with the client secret
+	// token for Oauth2 flow.
+	// Optional, as in previous resource versions we allowed no credentials.
+	// +optional
+	ClientSecret *SecretSelector `json:"clientSecret,omitempty"`
 	// Scopes for the token request.
 	Scopes []string `json:"scopes,omitempty"`
 	// The URL to fetch the token from.
```
bwplotka, line 254:
```suggestion
	// TokenURL specifies the URL to fetch the token from.
```

bwplotka: ditto below

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -115,37 +258,46 @@ type OAuth2 struct {
...
+			oauth2.ProxyURL = proxyConfig
 		}
-		oauth2.ProxyURL = proxyConfig
 	}
-	return oauth2, nil
+	return oauth2, errors.Join(errs...)
 }
 
 type ProxyConfig struct {
 	// HTTP proxy server to use to connect to the targets. Encoded passwords are not supported.
```
bwplotka, line 297:
```suggestion
	// ProxyURL specifies the HTTP proxy server to use to connect to the targets.
	// Encoded passwords are not (yet) supported.
```

bwplotka (approved):
Epic work! 

LGTM, **modulo small comments** (some of them non-blocking when marked so). Thanks!

During my testing I noticed we could improve error handling for permissions on our Prom fork, but I shared that with you already, we can do that even after releasing this feature and definitely in further PRs. 💪🏽

bwplotka:
Proposed adding example https://github.com/GoogleCloudPlatform/prometheus-engine/pull/916

pkg/operator/apis/monitoring/v1/http_types.go
TheSpiritXIII: According to https://book.kubebuilder.io/reference/markers/crd-validation this only applies when using `+kubebuilder:validation:Required`. Since these are not the top-level CRD, I will include this annotation on all fields.

e2e/authorization_test.go
```diff
@@ -23,400 +23,337 @@ import (
...
+					TokenURL: "http://go-synthetic.default.svc.cluster.local:8080/token",
+				},
+			},
+		},
+		errInvalidClientCredentials,
+	)
+}
+
+func authorizationTest(ctx context.Context, t *testing.T, kubeClient kubernetes.Interface, opClient versioned.Interface, name string, successConfig, failureConfig *monitoringv1.ScrapeEndpoint, errMsg string) {
+	t.Run("podmonitoring", func(t *testing.T) {
```
bwplotka, line 342: Thanks!

bwplotka (approved):
Thanks! 💪🏽 Let's go!

mpskovvang:
Thrilled about the new feature! Could you possibly share when it's expected to be available on GKE? Much appreciated!

TheSpiritXIII:
> Thrilled about the new feature! Could you possibly share when it's expected to be available on GKE? Much appreciated!

The feature is already available on the GKE rapid release channel 1.29.3-gke.1113000 and above. We expect to port this change to the regular release channel sometime before the end of Q2.

For an example configuration, see: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/examples/instrumentation/go-synthetic/go-synthetic-basic-auth.yaml

For full usage, see: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/doc/api.md

Feel free to report any issues you see as a new bug on GitHub. Thanks for the interest!

mpskovvang:
> Feel free to report any issues you see as a new bug on GitHub. Thanks for the interest!

Thanks for the quick reply. I've successfully upgraded to 1.29.3-gke.1282000.

Unfortunately I'm having a issue with namespaces:

> strict decoding error: unknown field "spec.endpoints[0].authorization.credentials.secret.namespace"

The `PodMonitoring` resource is also failing without a `secret.namespace` provided. If I omit the `authorization` the request succeeds.

```yaml
kind: PodMonitoring
metadata:
  name: meilisearch
  namespace: ecommerce
  labels:
    app.kubernetes.io/name: meilisearch
    app.kubernetes.io/instance: meilisearch
    app.kubernetes.io/version: "v1.7.0"
    app.kubernetes.io/component: search-engine
    app.kubernetes.io/part-of: meilisearch
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: meilisearch
      app.kubernetes.io/instance: meilisearch
  endpoints:
    - port: http
      path: /metrics
      interval: 1m
      timeout: 10s
      authorization:
        type: Bearer
        credentials:
          secret:
            name: meilisearch-master-key
            key: MEILI_MASTER_KEY
            namespace: ecommerce
```

TheSpiritXIII:
> Unfortunately, I'm having an issue with namespaces

The fix is available in version:

GKE 1.30: 1.30.2-gke.1054000 or later.

And will be available in versions:

GKE 1.29: 1.29.6-gke.1150000 or later.
GKE 1.28: 1.28.11-gke.1107000 or later.

Please allow 2-3 weeks for these versions to get rolled out. Thanks!

cagataygurturk:
> > Unfortunately, I'm having an issue with namespaces
> 
> The fix is available in version:
> 
> GKE 1.30: 1.30.2-gke.1054000 or later.
> 
> And will be available in versions:
> 
> GKE 1.29: 1.29.6-gke.1150000 or later. GKE 1.28: 1.28.11-gke.1107000 or later.
> 
> Please allow 2-3 weeks for these versions to get rolled out. Thanks!

Hi, on 1.29.6-gke.1254000 I am still experiencing the same issue. Can you confirm the release schedule?

TheSpiritXIII:
> Can you confirm the release schedule?

Hi, we ended up doing additional security patch releases. 0.12 as shown in the releases page is available on the following GKE minor versions:

- 1.28.11-gke.1176000
- 1.29.6-gke.1256000
- 1.30.2-gke.1394000

These are the corresponding available versions on the GKE rapid release channel:

- 1.28.11-gke.1260000
- 1.29.6-gke.1326000
- 1.30.2-gke.1394003

Please create a new issue if you're having still problems since messages here could easily get lost! Thanks!

PR #783, opened 2024-01-25
Cleaned code for Option3 (consistent selector); added validation.

bwplotka:
Some decision rationales:

* Only ClusterSecretKeySelector for consistency and explicitness. We could rename it to SecretKeySelector I guess.
* Credentials and others etc are optional for compatibiltiy AND name is similar to e.g. secretKeyRef in https://pkg.go.dev/k8s.io/api/core/v1#EnvVarSource It feels cleaner this way and allows extensions if needed (e.g. credentials from some different secret provider). It could be *SecretKeyRef or *Ref or *SecretRef suffix but chosen something that reflects it's not a string, but also not too verbose.
* Flattened the structure, make it a pointer, for simplicity.
* Detached low level types from scrape in comments (it's generic HTTP)
* One method for accessing things (and adding used secrets) for smaller boilerplate

### Verification

* Unit && e2e tests
* Manually running e2e against cluster and trying to access metric from authorized endpoint 

1. Added time.Sleep  in podmon success test case
2. Run:`GMP_CLUSTER=bwplotka-test NO_DOCKER=1 TEST_RUN=TestBasicAuth/full make e2e`
3. Checked PodMonitoring target status is OK 

Config:

image

4. Prometheus UI:

image

5. Metric from secure endpoint visible

image

6. Verified the endpoint is secure

image


pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -24,145 +24,119 @@ import (
...
-	Type        string                `json:"type,omitempty"`
-	Credentials ClusterSecretSelector `json:"credentials,omitempty"`
+	// The authentication type. Defaults to Bearer.
+	// Basic will cause an error, as the BasicAuth object should be used instead.
+	Type string `json:"type,omitempty"`
+	// CredentialsSecret references the Kubernetes secret's key with the credentials
+	// (token) for the auth header to send along the request.
+	// Optional, as in previous resource versions we allowed no credentials.
+	// +optional
+	CredentialsSecret *ClusterSecretKeySelector `json:"credentialsSecret,omitempty"`
```
TheSpiritXIII, line 106: suggestion: Does flattening them make it any easier for users? An extra indentation should not be much difficult.

The biggest benefit for splitting it into a separate type is the flexibility of adding additional secret types, e.g. ConfigMaps or Google Cloud Secret Manager. Then, there's also compatibility with the existing `Authorization` object which we should eventually merge.

Thoughts?

pkg/operator/collection.go
```diff
@@ -475,7 +468,10 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 			r.statusUpdates = append(r.statusUpdates, &cmon)
 		}
 	}
-	cfg.SecretConfigs = set.UnsortedList()
+
+	// TODO(bwplotka): Warn about missing RBAC?
```
TheSpiritXIII, line 472: nit: Can you create a bug for this as a potential improvement?

pkg/operator/apis/monitoring/v1/http_types_test.go
```diff
@@ -0,0 +1,137 @@
...
+				Name:      "secret1",
+				Key:       "key1",
+				Namespace: ns,
+			}
+
+			ref, err := c.toPrometheusSecretRef(p, pool)
+			if err != nil {
+				t.Fatalf("unexpected failure: %s", err)
+			}
+			if exp, got := "foo/secret1/key1", ref; exp != got {
```
TheSpiritXIII, line 48: nit: maybe we can extract this string format into a method? We currently have one but the format is slightly different (see `pathForSelector` in `operator_config.go`.

pkg/operator/apis/monitoring/v1/http_types_test.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+			if exp, got := []secrets.SecretConfig[kubernetes.SecretConfig]{
+				{
+					Name: "foo/secret1/key1", Config: kubernetes.SecretConfig{
+						Name:      c.Name,
+						Key:       c.Key,
+						Namespace: p.Namespace,
+					},
+				},
+			}, pool.SecretConfigs(); cmp.Diff(exp, got) != "" {
```
TheSpiritXIII, line 59: nit: can you store the result of `cmp.Diff` in a variable and reuse it or does Go not let you in the middle of the conditional?

pkg/operator/apis/monitoring/v1/http_types_test.go
```diff
@@ -0,0 +1,137 @@
...
+	t.Run("wrong namespace", func(t *testing.T) {
+		pool := PrometheusSecretConfigs{}
+		c := &ClusterSecretKeySelector{
+			Name:      "secret1",
+			Key:       "key1",
+			Namespace: "different",
+		}
+
+		_, err := c.toPrometheusSecretRef(p, pool)
+		if err == nil {
```
TheSpiritXIII, line 73: nit: should we check what the expected err should be? What if we fail by chance?

TheSpiritXIII (changes requested):
Great first pass and nice ideas, especially with regards to the secret pool! Also good job with the unit tests.

Let's discuss some of the tradeoffs of some of these decisions.

pkg/operator/apis/monitoring/v1/http_types_test.go
bwplotka: I think we could do that, but I prefer not as for "expected" strings it's more readable and safer to not use dynamic code. Only 4 very simple references

pkg/operator/apis/monitoring/v1/http_types_test.go
bwplotka: Yup, chaining is bit limited, so can't do this.

image

pkg/operator/apis/monitoring/v1/http_types_test.go
bwplotka: Currently there is only a single error, so there is no chance, so let's skip it (:

TheSpiritXIII (approved):
Thank you! Great work here.

PR #785, opened 2024-01-26
style: resolve linter errors and enable linter

bernot-dev:
Consolidated fixes after running linters on operator package.

Summary:

- Unused arguments changed to `_`
- `k8s.io/utils/pointer` is deprecated. Replace with `k8s.io/utils/ptr`.
- error messages start with lowercase letter.
- Exported functions have comments
- Packages are only imported once (including aliases).
- "if block ends with a return statement, so drop else and outdent its block"

Resolved or ignored (by line) linter errors and enabled default linters for golangci-lint, plus gofmt linter.

Enabled `make lint` in CI.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -26,6 +26,12 @@ jobs:
     - uses: actions/checkout@v3
     - name: Build go images
       run: make bin
+  lint:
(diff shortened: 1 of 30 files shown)


pintohutch (approved):
:shipit:

PR #786, opened 2024-01-26
Fix NodeMonitoring breaks target status

TheSpiritXIII:
After NodeMonitoring was added, when it is configured, target status stops working completely. Long story short, the problem is early exit on an unknown scrape pool format.

This change includes some refactoring originally part of the ProbeMonitoring PR (#766) but also adds in the possibility of supporting a NodeMonitoring target status if we choose to add it in the future, by at the very least detecting and validating the scrape pool format.

Changes include:
- Ensures we still patch everything on error (with new unit test case).
- Show all errors, not just the last one we found.
- Ensures we parse NodeMonitoring scrape pool formats (with new unit test case).
- Validate expected errors in the test cases.

Thanks @bwplotka for finding this issue!

### DIFF
--- a/pkg/operator/endpoint_status_builder.go
+++ b/pkg/operator/endpoint_status_builder.go
@@ -15,7 +15,7 @@
 package operator
 
 import (
-	"errors"
+	"fmt"
 	"sort"
 	"strconv"
 	"strings"
@@ -32,7 +32,7 @@ const (
 
 func buildEndpointStatuses(targets []*prometheusv1.TargetsResult) (map[string][]monitoringv1.ScrapeEndpointStatus, error) {
 	endpointBuilder := &scrapeEndpointBuilder{
-		mapByJobByEndpoint: make(map[string]map[string]*scrapeEndpointStatusBuilder),
+		mapByKeyByEndpoint: make(map[string]map[string]*scrapeEndpointStatusBuilder),
 		total:              0,
 		failed:             0,
 		time:               metav1.Now(),
@@ -48,7 +48,7 @@ func buildEndpointStatuses(targets []*prometheusv1.TargetsResult) (map[string][]
 }
 
 type scrapeEndpointBuilder struct {
-	mapByJobByEndpoint map[string]map[string]*scrapeEndpointStatusBuilder
+	mapByKeyByEndpoint map[string]map[string]*scrapeEndpointStatusBuilder
 	total              uint32
 	failed             uint32
 	time               metav1.Time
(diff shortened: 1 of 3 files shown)


pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,155 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+	if err != nil {
+		return nil, err
+	}
+	switch kind {
+	case ScrapeJobKubelet:
+		if len(split) != 1 {
+			return nil, fmt.Errorf("invalid Kubelet scrape job format %q", pool)
+		}
+		return nil, nil
+	case ScrapeJobPodMonitoring:
```
macxamin, line 123: suggestion: Instead of hardcoding the int number, we can assign the job names to variables and calculate the value. 

In t[his file](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/pkg/operator/apis/monitoring/v1/pod_types.go#L71-L73) we can add 

`podKey := "PodMonitoring/%s/%s" ` in pod_tyes.go

Then in this line we can have:
`if len(split) != len(strings.Split(podKey,"/"))`
That way the code is more future proof.

pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,155 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+		}
+		return scrapePool{
+			job:   split[0],
+			group: split[1],
+		}, nil
+	case ScrapeJobPodMonitoring:
+		if len(split) != 4 {
+			return scrapePool{}, fmt.Errorf("invalid PodMonitoring scrape pool format %q", pool)
+		}
+		index := len(split[0]) + len(split[1]) + len(split[2]) + 2
```
macxamin, line 176: why is it +2?

pkg/operator/target_status.go
```diff
@@ -362,29 +313,29 @@ func updateTargetStatus(ctx context.Context, logger logr.Logger, kubeClient clie
 		return err
 	}
 
-	var patchErr error
+	var errs []error
 	for job, endpointStatuses := range endpointMap {
-		// Kubelet scraping is configured through hard-coding and not through
-		// a PodMonitoring. As there's no status to update, we skip.
-		if strings.HasPrefix(job, "kubelet") {
```
pintohutch, line 369: IIUC the bug is we try and build the PodMonitoring to patch if we encounter a NodeMonitoring job, correct?

If so, couldn't the fix be a lot smaller with a simple
```go
if !(strings.HasPrefix("PodMonitoring/") || strings.HasPrefix("ClusterPodMonitoring")) {
	continue
}
```

with some new test cases to validate?

pintohutch (changes requested):
Thanks!

Can you also, either:
1. Create an issue describing the bug, or
2. Add the issue description in the PR

I think I understand the problem, but had to do a bit of reverse-engineering here. I may still not have the full picture.

pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,155 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+		return nil, fmt.Errorf("unknown scrape kind %q", split[0])
+	}
+}
+
+type scrapePool struct {
+	job   string
+	group string
+}
+
+func parseScrapePool(pool string) (scrapePool, error) {
```
macxamin, line 157: what does `scrapePool` look like vs `parseScrapeJob`?

macxamin (changes requested):
Thanks for quickly coding this! I have minor comments and I think it should be good to go after.

pkg/operator/target_status.go
TheSpiritXIII: I updated the original PR description.

tl;dr yes it could be a lot smaller. I pulled in some changes from https://github.com/GoogleCloudPlatform/prometheus-engine/pull/766 because I felt like it made the code cleaner overall.

My personal opinion is that simple is best if we want the fix out quickly. I confirmed with @macxamin that NodeMonitoring won't be used for a while so we we can just include this change as part of our next (hopefully biweekly) release. @macxamin feel free to let me know if I'm wrong or if something changes.

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: It's confusing but let me explain the terminology.

If you look at Prometheus' configuration, a scrape configuration has a job name. That job name is the fully qualified name, e.g. for PodMonitoring it's `PodMonitoring/namespace/name/endpoint`. When Prometheus actually does the scraping, this is called the "scrape pool". That's because the job actually could scrape multiple targets, so it actually scrapes a pool of resources. All of the targets have the same "job" label however, which is the PodMonitoring name. I use the same terminology of scrape pool and job here. The scrape pool contains the job name.

We have to split up this information twice. Remember for PodMonitoring, we want to patch target status for all endpoints into the same PodMonitoring status. Thus, for `PodMonitoring/namespace/name/endpoint`, `scrapePool.job=PodMonitoring/namespace/name` and `scrapePool.group=endpoint`.

After we grouped all endpoints together, we have to know which PodMonitoring to save the target status too. That's where we split up the job `PodMonitoring/namespace/name` and turn that into a PodMonitoring object.

Let me know if that's unclear.

pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,155 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+	case ScrapeJobNodeMonitoring:
+		return "NodeMonitoring"
+	default:
+		return ""
+	}
+}
+
+func parseScrapeKind(kind string) (ScrapeJobKind, error) {
+	switch kind {
+	case probeKindToString(ScrapeJobKubelet):
```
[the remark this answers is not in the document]
TheSpiritXIII: Great catch!

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: Added comments. Let me know if it's unclear.

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: Updated the names to make it clearer!

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: I don't expect this code to change in the future. I added a comment to explain the numbers. I felt that `strings.Split` inside a loop felt a little dirty although that may be considered a premature optimization. I'm not sure how well the Golang compiler inlines bytecode. If you're still adamant about it, I possibly add:

```golang
PodKeyFormat := "PodMonitoring/%s/%s"
PodKeySectionsAmount := strings.Split(PodKeyFormat, "/")
```

Or even something like:
```golang
// GetScrapePoolKindScope returns whether the kind is cluster scoped, namespaced or neither.
func GetScrapePoolKindScope() Scope
```

and use that to figure out how to generate and parse the keys. Thoughts?

pkg/operator/target_status.go
bwplotka: Yea, I did what @pintohutch suggested in other PR to get unblocked, but I'm happy for @TheSpiritXIII to fix this properly here 👍🏽  (and add support). We can always port or quickly add fix to the release if needed... 

Let's decide if we want 0.10 to have node monitoring properly, we can announce it on 0.11 or do 0.10.1 up to us

bwplotka (approved):
Thanks 👍🏽 LGTM

pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,145 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+
+const (
+	ScrapePoolInvalid ScrapePoolKind = iota
+	ScrapePoolKubelet
+	ScrapePoolPodMonitoring
+	ScrapePoolClusterPodMonitoring
+	ScrapePoolNodeMonitoring
+)
+
+func parseScrapeKind(kind string) (ScrapePoolKind, error) {
```
pintohutch, line 81: question: Why use this function at all (with the extra `iota` declarations)? What's wrong with rewriting the case statement to use the prefix directly?
i.e.
```go
kind := split[1]
switch kind {
case "kubelet":
	// lala
case "PodMonitoring":
	// lala
case "ClusterPodMonitoring":
	// lala
case "NodeMonitoring":
	// lala
default:
```

pkg/operator/endpoint_status_builder.go
pintohutch: Great explanation - thanks!

Could you add a comment in-code to clarify? Especially with the example - makes it really clear.

pkg/operator/target_status.go
pintohutch: Gotcha.

Thanks for the context. The new PR subscription colors this in nicely 😄 .

pintohutch (commented):
Round 2

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: Good point. Thanks!

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: Added the example in-code. Thanks!

pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,116 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+	}
+}
+
+// scrapePool is the parsed Prometheus scrape pool, which is assigned to the job name from our
+// configurations. For example, for PodMonitoring this is `PodMonitoring/namespace/name/port`. The
+// key is what idenitifies the resource (`PodMonitoring/namespace/name`) and the group indiciates a
+// small subset of that resource (`port`).
+type scrapePool struct {
+	key   string
+	group string
```
pintohutch, line 119: nit (non-blocking): it may be more clear to just call the fields what you're describing in the comment, i.e. renaming `group` to `port`. Unless you think it's generalizable to another concept in our CRs...

pintohutch (approved):
:shipit:

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: It's different in each CRD:
- PodMonitoring: port
- NodeMonitoring: path
- Probe: URL/index

pintohutch: Ah touche - LGTM!

pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,132 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+// configurations. For example, for PodMonitoring this is `PodMonitoring/namespace/name/port`. The
+// key is what identifies the resource (`PodMonitoring/namespace/name`) and the group indicates a
+// small subset of that resource (`port`).
+type scrapePool struct {
+	key   string
+	group string
+}
+
+// slicesJoinLength is strings.Join but only returns the length of the resulting string.
+func stringsJoinLength(elems []string, sep string) int {
```
macxamin, line 127: sep is not used.

pkg/operator/endpoint_status_builder.go
```diff
@@ -68,24 +68,132 @@ func (b *scrapeEndpointBuilder) add(target *prometheusv1.TargetsResult) error {
...
+		}
+		return scrapePool{
+			key:   split[0],
+			group: split[1],
+		}, nil
+	case "PodMonitoring":
+		if len(split) != 4 {
+			return scrapePool{}, fmt.Errorf("invalid PodMonitoring scrape pool format %q", pool)
+		}
+		index := stringsJoinLength(split[:3], "/")
```
macxamin, line 153: what do you think about using `index := strings.LastIndex(scrapePool, "/")` ?

pkg/operator/endpoint_status_builder.go
macxamin: Thanks for adding the comment!

macxamin (changes requested):
Thanks!

pkg/operator/endpoint_status_builder.go
macxamin: Thanks, it's clearer now!

pkg/operator/endpoint_status_builder.go
TheSpiritXIII: Great catch! Mostly because this won't work in the future.

For Probe, this will not be the last index. In Probe, the way it currently works is you specify multiple URLs per protocol. For example, you might probe `google.com` and `example.com` using HTTP and `monitoring.googleapis.com:443` for gRPC in the same Probe object. The way we create this group is `protocol/index`, e.g. for `example.com`, `http_2xx/1` (with 1 because it is an array and that is its position `["google.com", "example.com"]`). A workaround might be to avoid using a slash here but I feel that using the same delimiter all throughout is the cleanest.

macxamin (approved):
Amazing 💪

PR #745, opened 2024-01-03
[E2E refactor - 3] re-introduce generated clientsets to prepare for simpler e2e tests

pintohutch:
I decided to break out the kind E2E test refactor PR https://github.com/GoogleCloudPlatform/prometheus-engine/pull/738 into smaller, digestible PRs for reviewing.

Note: because of the nature of the change, presubmits (i.e. Github Actions) may fail until the final PR is merged.

This is the third one, where we reintroduced generating k8s clientsets of our CRD types. This sets us up to ensure e2e tests don't require controller-runtime, which can be beneficial when importing the library elsewhere. This is an effective revert of c5eeee3209b63920ba888f2025f127ac4202564c.

PR #787, opened 2024-01-26
[E2E refactor - 12] Remove unnecessary lister and informer sets

pintohutch:

This is the twelfth (and really really final!) one, where we reduce the generated clientsets as informer and lister aren't needed for E2E tests. This also speeds up build times and makes builds idempotent.


pintohutch:
Folding this into #745

PR #788, opened 2024-01-26
style: add additional linters

bernot-dev:
Add linter to check fields have matching name and value for loggers.


bwplotka (approved):
Nice! Thanks 💪🏽

Issue #725, opened 2023-12-19
High cardinality of Prometheus metrics in self-scrape scenarios (only self-deploy).


bwplotka:
The way we configure scrape jobs causes bit high cardinality of `net_conntrack_dialer_conn_failed_total`. It would be nice to work with the upstream to provide an option to disable it, or improve our example PodMonitoring for self-scrape to explictly ignore this metric.

![image](https://github.com/GoogleCloudPlatform/prometheus-engine/assets/6950331/ca5341fb-1837-46f4-adbf-fb7594839788)

bwplotka:
All metrics with scrape_job has the same issue e.g. `prometheus_target_sync_length_seconds` but they might have some value for some advanced debugging cases.

https://github.com/prometheus/prometheus/blob/bfbb13cf369da4cd1b29ee52c396c902723febfb/scrape/metrics.go#L138

PR #794, opened 2024-01-29
fix: update self-pod-monitoring.yaml example to reduce ingestion of self-scraped metrics.

macxamin:
This PR aims to reduce aim to reduct the amount of self-scraped metrics that are ingested.  

1. Increase the scrape interval to 30s.
2. Drop metrics with high cardinality.

Also resolves b/322791992 and part of https://github.com/GoogleCloudPlatform/prometheus-engine/issues/725

### DIFF
--- a/examples/self-pod-monitoring.yaml
+++ b/examples/self-pod-monitoring.yaml
@@ -26,9 +26,14 @@ spec:
       app.kubernetes.io/name: collector
   endpoints:
   - port: prom-metrics
-    interval: 10s
+    interval: 30s
+    metricRelabeling:
+    - sourceLabels: [__name__]
+      regex: > 
+            prometheus_target.*|prometheus_sd.*|net_conntrack_.*
+      action: drop
   - port: cfg-rel-metrics
-    interval: 10s
+    interval: 30s
 ---
 apiVersion: monitoring.googleapis.com/v1
 kind: PodMonitoring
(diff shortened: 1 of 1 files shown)


examples/self-pod-monitoring.yaml
```diff
@@ -26,9 +26,14 @@ spec:
       app.kubernetes.io/name: collector
   endpoints:
   - port: prom-metrics
-    interval: 10s
```
lyanco, line 29: Is this scrape interval reduction worth it? People are complaining about the cardinality of less-useful metrics, not the scrape interval.
I'm not sure why these were set to 10s originally - perhaps it's more important to have this fast of a feedback cycle when it comes to self-evaluation - but we should figure out the Chesterton's Fence here.

macxamin: Good point. I see the [Prometheus example uses 15s.](https://github.com/prometheus/prometheus/blob/501bc6419ee54ee5617d75c3bc16ecf36d94c552/documentation/examples/prometheus.yml#L3-L4)

@bwplotka what do you think?

bwplotka: Yea. My understanding was that this is a "useful example for debugging GMP cases, by GMP (e.g. prometheus-engine) advanced users, devs/contributors." - that's why shorter interval, which is not needed for long term scraping.

However, we see users using this thinking it's mandatory or "generally" useful. No one blocks them, but they would be fine with 30s or even 60s.

lyanco: OK, then let's do 30s

bwplotka (approved):
Thanks!

PR #796, opened 2024-01-29
style: add conform tool to enforce commit messages

bernot-dev:
Add Conform tool to enforce conventional commits style along with other basic standards, like enforcing a body for the commit message.


.conform.yaml
```diff
@@ -0,0 +1,29 @@
...
+        invalidLastCharacters: .
+      body:
+        required: true
+      spellcheck:
+        locale: US
+      maximumOfOneCommit: false
+      conventional:
+        # Consistent with @commitlint/config-conventional, as referenced in Conventional Commits Specification.
+        # https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum
+        types:
```
TheSpiritXIII, line 17: question: why is `feat` and `fix` missing here? Is that intended or does it append these to some defaults?

.github/workflows/presubmit.yml
```diff
@@ -26,6 +26,14 @@ jobs:
     - uses: actions/checkout@v3
     - name: Build go images
       run: make bin
+  conform:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
```
TheSpiritXIII, line 34: nit: is this necessary?

.github/workflows/presubmit.yml
```diff
@@ -36,6 +44,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
```
TheSpiritXIII, line 48: nit: is this necessary?

TheSpiritXIII (changes requested):
I'm really looking forward to this change! Can you take a look at my questions?

.conform.yaml
bernot-dev: Yes, Feat and Fix are included by default. Not sure why those two and not any others.

.github/workflows/presubmit.yml
bernot-dev: Yes, this is necessary if we want to allow an arbitrary number of commits per PR (I think we do). Otherwise, it errors out because it can't see the history.

.github/workflows/presubmit.yml
bernot-dev: No, this was a remnant of an intermediate step where I added conform to the lint step. I'll remove this. Good catch!

.conform.yaml
bernot-dev: Ref: https://github.com/siderolabs/conform/blob/v0.1.0-alpha.27/internal/policy/commit/check_conventional_commit.go#L78

TheSpiritXIII: Thanks for checking!

TheSpiritXIII (approved):
Thanks for answering everything! LTGM.

PR #797, opened 2024-01-29
fix: use operator namespace to check if a NodeMonitoring should be applied

macxamin:
NodeMonitorings will be generated by GMP using the operator namespace. Only apply them if KubeletScraping is disabled.


TheSpiritXIII (approved):
Thank you!

PR #800, opened 2024-01-30
docs: improve datasource-syncer README

macxamin:
1. Update cronjob command with namespace
2. Correct local Grafana address

### DIFF
--- a/cmd/datasource-syncer/README.md
+++ b/cmd/datasource-syncer/README.md
@@ -13,7 +13,7 @@ By regularly refreshing the oAuth2 access token, you can configure Grafana to di
 
 ### Run
 
-1. Figure out the URL of your Grafana instance. e.g. `https://your.grafana.net` for a Grafana Cloud instance or `http://localhost:3000` for a local instance.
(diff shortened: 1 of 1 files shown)


cmd/datasource-syncer/README.md
```diff
@@ -29,6 +29,7 @@ DATASOURCE_UIDS=YOUR_DATASOURCE_UIDs
 GRAFANA_API_TOKEN=YOUR_GRAFANA_SERVICE_ACCOUNT_TOKEN
 GRAFANA_API_ENDPOINT=YOUR_GRAFANA_INSTANCE_URL
 PROJECT_ID=PROJECT_ID_TO_QUERY_GCM
+NAMESPACE_NAME=NAMESPACE_NAME that is binded to your Google Cloud service account.
```
lyanco, line 32: would get rid of the help text as it probably will make an error if copy-pasted

Could put it below, "choose the namespace in which you deploy the data source syncer. This should be the same namespace you bind your workload identity service account to, if enabled"

macxamin:
```suggestion
NAMESPACE_NAME=NAMESPACE_NAME #that you deploy the datasource-syncer to, and that you bound your service account to if using workload identity
```

PR #801, opened 2024-01-30
ci: change make lint command

bernot-dev:
Using `go run` instead of running the binary inside of Docker reduces the linting time considerably, from on the order of a minute with Docker to about 3-5 seconds using `go run`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ conform:
 lint:        ## Lint code.
              ##
 	@echo ">> linting code"
-	DOCKER_BUILDKIT=1 docker run --rm -v $(shell pwd):/app -w /app golangci/golangci-lint:v1.54.2 golangci-lint run -v --timeout=5m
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -55,7 +55,7 @@ conform:
 lint:        ## Lint code.
              ##
 	@echo ">> linting code"
-	DOCKER_BUILDKIT=1 docker run --rm -v $(shell pwd):/app -w /app golangci/golangci-lint:v1.54.2 golangci-lint run -v --timeout=5m
+	go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2 run --verbose --timeout 5m
```
TheSpiritXIII, line 58: suggestion: we've been using `NO_DOCKER` in the Makefile for fast-paths. To keep current conventions, consider using that type of branch here.

TheSpiritXIII (approved):

PR #805, opened 2024-02-02
chore(configs): support dependabot conventional commits

TheSpiritXIII:


.conform.yaml
```diff
@@ -7,23 +7,26 @@ policies:
         case: lower
         invalidLastCharacters: .
       body:
-        required: true
+        required: false
```
bernot-dev, line 10: Question: Dependabot doesn't create commit bodies?

TheSpiritXIII: Dependabot includes commit bodies. This is more of a me change. I don't create commit bodies so this exact commit would have failed without this change. Thoughts?

bernot-dev: It looks to me like Dependabot creates well-formed commit bodies with a DCO and everything. Is this unrelated?

TheSpiritXIII: Correct. I can split this out if requested.

bernot-dev: Yeah, there's probably not a good reason for us to enforce commit bodies at the moment.

bernot-dev (approved):
LGTM.

PR #806, opened 2024-02-02
fix: make NodeMonitoring cluster-scoped

macxamin:
NodeMonitoring is cluster-scoped.


pintohutch (approved):
LGTM but I'd like @TheSpiritXIII or @bwplotka to sanity-check me here

TheSpiritXIII (approved):
LGTM! As long as all the tests pass, this looks generated right and should still be working.

bwplotka (approved):

PR #810, opened 2024-02-05
fix(deps): update k8s.io dependencies

TheSpiritXIII:
Doing it manually because telling Dependabot to recreate the PR does not do it in the new format we specified in the Dependabot config files.


macxamin (approved):
LGTM

PR #811, opened 2024-02-06
Limit size of kind cluster names

pintohutch:
Due to https://github.com/kubernetes-sigs/kind/issues/623, we need to
ensure we do not violate the length of a kind cluster's name.

We have been ensuring unique cluster names by using the name of the go
test function as the cluster name. However, for longer functions, we run
into this issue.

This change fixes the length of the name to 32 characters and appends a
random hash to the end of the name to prevent collisions for
similarly-named tests.

### DIFF
--- a/hack/kind-test.sh
+++ b/hack/kind-test.sh
@@ -25,7 +25,12 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 TAG_NAME=$(date "+gmp-%Y%d%m_%H%M")
 TEST_ARGS=""
 # Convert kind cluster name to required regex if necessary.
-KIND_CLUSTER=`echo ${GO_TEST} | sed -r 's/([A-Z])/-\L\1/g' | sed 's/^-//'`
+# We need to ensure this name is not too long due to: https://github.com/kubernetes-sigs/kind/issues/623
+# while still unique enough to avoid dups between similar test names when trimming.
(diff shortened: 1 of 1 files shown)


hack/kind-test.sh
```diff
@@ -25,7 +25,7 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 TAG_NAME=$(date "+gmp-%Y%d%m_%H%M")
 TEST_ARGS=""
 # Convert kind cluster name to required regex if necessary.
-KIND_CLUSTER=`echo ${GO_TEST} | sed -r 's/([A-Z])/-\L\1/g' | sed 's/^-//'`
+KIND_CLUSTER=`echo ${GO_TEST} | sed -r 's/([A-Z])/-\L\1/g' | sed 's/^-//' | head -c32`
```
TheSpiritXIII, line 28: suggestion: tests would probably use similar prefixes for tests so I don't think trimming down to 32 is a foolproof. Maybe we should trim it a little more and add a random number at the end?

pintohutch: Sure - lemme change this to point to `main` and just consider addressing the cluster name bug.

pintohutch: Ok done - PTAL!

TheSpiritXIII (approved):

PR #813, opened 2024-02-07
refactor(e2e): deploy resources within Go

TheSpiritXIII:
Notable changes:
- We don't pass GOOGLE_APPLICATION_CREDENTIALS into Docker if it's empty.
- We don't create mock credentials anymore since we now simply pass the no auth flag to the collector like we used to.
- We now create the example-app on-demand.


e2e/deploy/synthetic.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/GoogleCloudPlatform/prometheus-engine/e2e/kube"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	SyntheticAppPortName      = "web"
+	SyntheticAppContainerName = "go-synthetic"
+
+	appManifest = "../examples/instrumentation/go-synthetic/go-synthetic.yaml"
```
pintohutch, line 29: We won't have this in path in google3 - it only exists in this relative location in Github, so we won't be able to use these setup methods there unless we wanted to start maintaining a mirror internally and move everything over periodically...

hack/kind-test.sh
```diff
@@ -25,57 +25,37 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
...
 # Ensure a unique label on any test data sent to GCM.
 GMP_CLUSTER=$TAG_NAME
-if [ -f "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
-  PROJECT_ID=$(jq -r '.project_id' ${GOOGLE_APPLICATION_CREDENTIALS})
+if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
+  PROJECT_ID=$(jq -r '.project_id' "${GOOGLE_APPLICATION_CREDENTIALS}")
 else
   echo ">>> no credentials specified. running without GCM validation"
   TEST_ARGS="${TEST_ARGS} -skip-gcm"
-  # Create mock credentials.
```
pintohutch, line 37: Interesting - this bug may have been solved at some point then.

Before, even if you passed a disable-auth flag to the collector, it would still crashloop with this error message without some form of credentials.

pintohutch (commented):
I remember we discussed using go more to control e2e testing.

In this case, can you give the functional benefits of what this gives us to justify the +450 lines (that we didn't have before)?

bernot-dev:
Just a drive-by comment, but I would expect the commit type of a PR changes to tests to be "test" or perhaps "refactor" if there is no expected change to test behavior.

For me, `chore` implies that the change is mostly related to tooling, updating dependencies, and the like.

TheSpiritXIII:
The amount of actual logic in Go is difficult to measure in lines. In reality, there are only a few dozen more lines than Bash due to Go boilerplate and the addition of a few more features.

Some benefits over what's currently in the main branch:
1. The synthetic app utility could be used in internal testing for ensuring that different metrics are written correctly.
2. Go is better than Bash: You get compile-time warnings, better error handling (with the k8s libraries), developers are more familiar with it.
3. The current tests don't check if the operator is ready. Some tests fail if they start too fast. The reason it usually works today is because we deploy the synthetic app and wait for it before starting tests so we get lucky. The same readiness checks could replace what we use internally.

If we move all setup to Go tests (we would need to migrate Kind tests over), then we don't need to use xargs for parallelism. This means:
- Developers could use their existing IDEs to debug tests without needing to configure setup.
- Developers could run without Docker easily since `go test ./e2e` would just work.
- Increased performance when running all tests locally because the Go runtime can start more tests at once.

pintohutch:
Thanks for the details!

In general, I'm totally fine with using go to setup our tests over bash - I don't have a strong opinion really, whatever is easier to maintain for folks.

pintohutch:
I'm going to pull myself off the review to let others weigh in, as it would be good to spread the knowledge of how these work to the rest of the team.

Let me know if there's anything you'd like my take on besides that.

e2e/deploy/deploy.go
```diff
@@ -0,0 +1,189 @@
...
+				container.Args = append(container.Args, fmt.Sprintf("--operator-namespace=%s", opts.operatorNamespace))
+				container.Args = append(container.Args, fmt.Sprintf("--public-namespace=%s", opts.publicNamespace))
+			}
+		}
+		return obj, nil
+	})
+}
+
+type DeployOption interface {
+	applyToDeployOptions(*deployOptions)
```
bernot-dev, line 77: Question: Is there a reason to make these funcs bound to a type instead of `type DeployOption func(*deployOptions)`?

bernot-dev (approved):
LGTM

PR #815, opened 2024-02-08
fix: remove replica count for rule-evaluator

bernot-dev:
This was missed in #690, and needs to be fixed for autoscaling the rule-evaluator to work correctly on GKE.


TheSpiritXIII (approved):
Thanks!

PR #817, opened 2024-02-08
refactor: use helm to generate manifests

bernot-dev:
Helm offers a more elegant solution to some of the tasks around assembling the manifests for the operator.

There are a few key changes in this PR:
- Operator manifests move to a Helm chart in `charts/operator` instead of `cmd/operator/deploy`.
- Rule-evaluator manifests are split into a separate chart at `charts/rule-evaluator`.
- Helm is added to Bingo and pinned at v3.14.0.
- YQ is added to Bingo and pinned at v4.40.7.
- `update_manifests` only updates the yaml files in `manifests/` and `examples`. It does not attempt to update the Helm Charts. It previously updated the source manifests in `cmd/operator/deploy`.
- CRDs were regenerated with updated line wrapping. No actual changes.
- diff test for generated files produces output instead of being silenced.
- Documentation for the operator `cmd/operator/README.md` was updated to use Helm.
- Update Makefile target `updateversions` to modify `values.yaml` and re-run `helm template`.
- Added common labels to all operator resources.

Image versions for manifests now only need to be updated in the `values.yaml` files for their charts to be propogated properly. This has been reflected in the changes to the `make updateversions` target.

Manifests are sorted by the [standard Helm ordering](https://helm.sh/docs/intro/using_helm/#:~:text=Helm%20installs%20resources%20in%20the%20following%20order%3A). No changes were ultimately made to the substance of any resource in `manifests/operator.yaml`. This maintains compatibility with GKE tooling, which relies on that manifest.

This change creates an opportunity to use Helm to generate manifests on the GKE side using Helm as well, which could significantly reduce the complexity related to creating GKE-specific manifests.

**EXPERIMENTAL:** Installing and managing GMP with Helm should be considered experimental and is not supported at this time.


manifests/rule-evaluator.yaml
```diff
@@ -1,18 +1,11 @@
-# Copyright 2022 Google LLC
-#
```
pintohutch, line 2: You're losing our license header. Can we add that in through helm?

Makefile
```diff
@@ -231,29 +231,42 @@ presubmit:   ## Regenerate all resources, build all images and run all tests.
...
+export CURRENT_AM_TAG = v0.25.1-gmp.2-gke.0
+export CURRENT_RE_TAG = v0.9.0-gke.1
+export CURRENT_CONFIG_RELOADER_TAG = v0.9.0-gke.1
 #TODO(macxamin) Sync CURRENT_DATASOURCE_SYNCER_TAG with CURRENT_TAG
-CURRENT_DATASOURCE_SYNCER_TAG = v0.9.1-gke.0
-CURRENT_PROM_TAG = v2.41.0-gmp.9-gke.0
-CURRENT_AM_TAG = v0.25.1-gmp.2-gke.0
-LABEL_API_VERSION = 0.9.0
-FILES_TO_UPDATE = $(shell find manifests cmd/operator/deploy -type f -name "*.yaml")
+export CURRENT_DATASOURCE_SYNCER_TAG = v0.9.1-gke.0
```
macxamin, line 240:
```suggestion
export CURRENT_DATASOURCE_SYNCER_TAG = v0.10.0-gke.3
```

manifests/rule-evaluator.yaml
bernot-dev: Added `addlicense` tool.

Makefile
bernot-dev: Shouldn't this be in a separate change? I didn't change the version here.

bernot-dev: Nevermind, rebased on main again and that picked it up.

.bingo/Variables.mk
```diff
@@ -7,19 +7,37 @@ GO     ?= $(shell which go)
...
 #include .bingo/Variables.mk # Assuming -dir was set to .bingo .
 #
-#command: $(GOLANGCI_LINT)
-#	@echo "Running golangci-lint"
-#	@$(GOLANGCI_LINT) <flags/args..>
+#command: $(ADDLICENSE)
+#	@echo "Running addlicense"
+#	@$(ADDLICENSE) <flags/args..>
 #
+ADDLICENSE := $(GOBIN)/addlicense-v1.1.1
```
pintohutch, line 20: Doesn't addlicense need the path for the license header to use? https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/hack/boilerplate.txt

charts/operator/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: prometheus-engine
+description: Google Managed Service for Prometheus (GMP) Operator
+type: application
+version: "0.1.0"
```
pintohutch, line 5: question: what's the difference between `version` and `appVersion`?

cmd/operator/README.md
```diff
@@ -31,11 +31,10 @@ environment variables, such as `PROJECT_ID` but these are configured
 automatically if they are not set. The command will give you the uploaded image
 URL and update all necessary configurations to use it.
 
-Next, apply the Kubernetes configuration files, starting with the CRDs:
+Next, install the Kubernetes configuration files using Helm:
```
pintohutch, line 34: I think we want to still advertise the `kubectl apply -f` option for our manifests to:
1. Avoid being prescriptive
2. Be consistent with the GCP docs

.bingo/Variables.mk
bernot-dev: addlicense generates exactly our license by default.

manifests/operator.yaml
```diff
@@ -1,51 +1,95 @@
...
 
-# NOTE: This file is autogenerated.
+---
+# Source: prometheus-engine/templates/priority-class.yaml
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: gmp-critical
+  labels:
+    helm.sh/chart: prometheus-engine-0.1.0
```
pintohutch, line 22: Can we avoid adding the helm labels?

charts/operator/Chart.yaml
bernot-dev: `version` is the version of the chart. `appVersion` is the version of GMP.

.bingo/Variables.mk
pintohutch: oh sweet

manifests/operator.yaml
bernot-dev: Done.

charts/operator/Chart.yaml
pintohutch: ah yea ok

cmd/operator/README.md
bernot-dev: Reverted this file to original.

pintohutch: Maybe you forgot the commit?

manifests/operator.yaml
pintohutch: Maybe you forgot the commit?

cmd/operator/README.md
bernot-dev: LOL, I pushed it into #819. I'm just going to merge that into this branch. I'm having a bad Git day.

pintohutch (approved):

PR #819, opened 2024-02-09
refactor: split webhook rbac into separate role

bernot-dev:
Move RBAC permissions to manage MutatingWebhookConfigurations and ValidatingWebhookConfigurations into a separate Role. This allow these permissions to be removed more easily when they are not needed.


charts/operator/templates/role.yaml
```diff
@@ -149,3 +132,26 @@ rules:
   - statefulsets
   apiGroups: ["apps"]
   verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: gmp-system:webhook-admin
```
TheSpiritXIII, line 139: suggestion: the old name made it clear the permission was used solely for the operator binary. This one does not. Should we prefix it to something like `gmp-system:operator:webhook-admin`? Do you know if there are Kubernetes conventions we could follow?

TheSpiritXIII (approved):
Left a minor comment regarding naming! Otherwise looks good and welcomed!

charts/operator/templates/role.yaml
bernot-dev: I'm not familiar with any well-accepted naming conventions in this area. I updated the code to match your suggestion.

PR #820, opened 2024-02-10
fix: scale alertmanager/rule-evaluator using patch

bernot-dev:
When using a get followed by an update, controller-runtime attempts to apply the metadata again, resulting in an error. If we generate and apply a patch instead, it will only modify the field that has changed.


bernot-dev:
Incorrect assumption about the problem.

PR #821, opened 2024-02-11
fix: match exact test on e2e test matrix

TheSpiritXIII:
This fixes two issues:
1. `TEST_RUN` takes in a regular expression. After the recent refactoring, this no longer works. This is because we create a kind cluster with a name based off of the TEST_RUN. If there any special characters (e.g. from regular expressions), this fails.
2. If we add a new text that's a prefix or suffix of the same test, that test will be ran twice on GitHub. For example, if we have `TestA` and `TestAVariant`, then when GitHub runs `TestA`, it will run both tests within that runner and then it will run `TestAVariant` again in a separate runner. This is because Go takes in a regular expression when filtering tests and `TestA` matches both tests. We fix this by adding symbols for start and end of string, e.g. we run `TestA` with `^TestA$` to ensure it only runs `TestA`.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -85,7 +85,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - id: set-testruns
-      run: echo "testrun=$(go test ./e2e/... -list=. | grep -E 'Test*' | jq -R -s -c 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


pintohutch (approved):
nit (non-blocking): in terms of conventional commits, I'm unsure how to categorize this - either "fix: accidental multi-test matching", or "feat:  support regular expressions in TEST_RUN". It's kind of both.

Maybe a "fix" is a stronger case.

TheSpiritXIII:
Thanks! To be fair, it used to be supported before the e2e test refactor. The fact that it is not anymore can be considered a regression.

PR #831, opened 2024-02-13
chore: add license check/generation

bernot-dev:
Add `addlicense` checks to all source files, consistent with Google standards.

The `make regen` target is run in check mode in CI, so this will be enforced on new PRs.


Makefile
```diff
@@ -137,15 +137,17 @@ bin: $(GOCMDS) $(GOINSTR)
 regen:       ## Refresh autogenerated files and reformat code.
              ## Use CHECK=1 to only validate clean repo after run.
              ##
-regen:
+regen: $(ADDLICENSE)
```
TheSpiritXIII, line 140: question: maybe I'm missing something but where did this come from? I don't see this defined anywhere.

TheSpiritXIII (approved):
Just added a question. Thanks!

Makefile
bernot-dev: It's in Bingo. The dependency was already added with the Helm change. This is just expanding it to check the rest of the repo.

https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/.bingo/Variables.mk#L20-L24

pintohutch (approved):

Makefile
TheSpiritXIII: Thanks! Looks good!

PR #832, opened 2024-02-13
chore: fix dependabot commit scopes

bernot-dev:
Basically, dependabot tries to be smart. It detects the previous commit message, and then extrapolates based on that to determine how to to create the next one.

With `include: scope` on, we should not include the scope in the prefix, because [it will be included again](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator/pr_name_prefixer.rb#L110).

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,48 +5,48 @@ updates:
   schedule:
     interval: weekly
   commit-message:
-    prefix: fix(deps)
-    prefix-development: chore(deps)
+    prefix: fix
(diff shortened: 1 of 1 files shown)


bernot-dev:
Apparently, this has been confusing other people, too. Seems like the documentation of the feature is incomplete.

https://github.com/dependabot/dependabot-core/issues/8443

TheSpiritXIII (approved):
Let's give it a try!

PR #833, opened 2024-02-13
ci: add release-please

bernot-dev:
Testing release-please.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,33 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+on:
+  push:
+    branches:
+      - main
```
TheSpiritXIII, line 18: suggestion: Does this support regex? Could we add `release-*` too?

TheSpiritXIII (approved):

TheSpiritXIII:
Wondering if we should backfill the CHANGELOG. We could potentially pull it from older release. Would it be worth it? I'm leaning towards no, but I'm curious what other folks think. 🤔 @pintohutch @macxamin any opinions?

.github/workflows/release-please.yml
bernot-dev: I think it's likely going to take some tuning before we actually accept PRs. This will give us more information about what exactly it will do for us, though. I would defer additional branches to a later change.

bernot-dev:
I'm just going to see what it generates. This is a low-risk merge, as it will just create a PR, if it works. We can always close the PR, make adjustments, or remove the workflow if we don't like what it's doing for some reason.

pintohutch:
> Wondering if we should backfill the CHANGELOG. We could potentially pull it from older release. Would it be worth it? I'm leaning towards no, but I'm curious what other folks think. 🤔 @pintohutch @macxamin any opinions?

No strong opinions from me

PR #836, opened 2024-02-13
chore: update linters

bernot-dev:
Upgrade golangci-lint to v1.56.1.

Enable all linters that are supported and currently passing.

Keep a list of disabled linters instead of a list of enabled linters. This will make it easier to correct issues and enable linters without searching documentation.


macxamin (approved):
LGTM

PR #839, opened 2024-02-13
chore: enable nilerr linter

bernot-dev:
nilerr: `Finds the code that returns nil even if it checks that the error is not nil.`


macxamin (approved):
Nice!

pkg/operator/target_status_test.go
```diff
@@ -1380,7 +1380,7 @@ func TestPolling(t *testing.T) {
 		if pollErr := wait.PollUntilContextTimeout(ctx, 100*time.Millisecond, 2*time.Second, true, func(ctx context.Context) (bool, error) {
 			var podMonitorings monitoringv1.PodMonitoringList
 			if err := kubeClient.List(ctx, &podMonitorings); err != nil {
-				return false, nil
+				return false, err
```
TheSpiritXIII, line 1383: For `wait.PollUntilContextTimeout`, returning an error stops the polling and this is usually not the behavior we want. However, in this case, I can't think of any reason that fetching a list of all `PodMonitoring`s would fail so this looks okay to me!

TheSpiritXIII (approved):
Awesome! Thanks!

PR #841, opened 2024-02-13
style: enable promlinter

bernot-dev:
promlinter: `Check Prometheus metrics naming via promlint.`

### DIFF
--- a/pkg/export/bench/app/example_app.go
+++ b/pkg/export/bench/app/example_app.go
@@ -77,12 +77,14 @@ var (
 	metricIncomingRequestsPending = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "example_incoming_requests_pending",
+			Help: "Number of incoming requests currently pending.",
 		},
 		[]string{"status", "method", "path"},
 	)
 	metricOutgoingRequestsPending = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "example_outgoing_requests_pending",
+			Help: "Number of outgoing requests currently pending.",
 		},
 		[]string{"status", "method", "path"},
 	)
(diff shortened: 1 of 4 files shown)


pkg/export/bench/app/example_app.go
```diff
@@ -77,65 +77,79 @@ var (
...
 	metricOutgoingRequestErrors = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "example_outgoing_request_errors_total",
+			Help: "Total number of errors on outgoing requests.",
 		},
 		[]string{"status", "method", "path"},
 	)
 
 	metricIncomingRequestDurationHistogram = prometheus.NewHistogramVec(
+		//nolint:promlinter
```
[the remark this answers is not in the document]
bernot-dev: The lint is:
`pkg/export/bench/app/example_app.go|122 col 3| Metric: example_histogram_incoming_request_duration Error: metric name should not include type 'histogram' (promlinter)`

I wasn't sure if renaming would break anything else and didn't want to get too far sidetracked. I will just note that they are test metrics.

pkg/secrets/manager.go
```diff
@@ -33,6 +33,7 @@ var (
 		},
 	)
 	secretsTotal = prometheus.NewGauge(
+		//nolint:promlinter
 		prometheus.GaugeOpts{
 			Name: "prometheus_kubernetes_secrets_total",
```
bernot-dev, line 38: @TheSpiritXIII This got flagged by promlinter.
`non-counter metrics should not have "_total" suffix`

Suggestions?

TheSpiritXIII: I'm not sure we can change the metric type 🤔 . Can we rename it to `prometheus_google_kubernetes_secrets_total` since it's for our fork, and turn it into a Counter?

bwplotka: Good question, already benefit of having this promlinter enabled!

Shouldn't it be a gauge of `prometheus_kubernetes_secrets_configured`?  and `prometheus_kubernetes_failed_secret_configs` I would do `prometheus_kubernetes_secrets_misconfigured`?

It's safe to change here quickly

pkg/export/bench/app/example_app.go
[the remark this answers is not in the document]
bwplotka: SGTM!

PR #842, opened 2024-02-14
style: enable revive linter

bernot-dev:
Revive is a meta-linter with many available rules, fixes were made in accordance with these default rules:

- empty-block
- error-strings
- increment-decrement
- indent-error-flow
- receiver-naming
- unexported-return
- unused-parameter
- var-naming

See [Revive Available Rules](https://github.com/mgechev/revive?tab=readme-ov-file#available-rules) for details.

Then, enabled all revive rules. Disabled all currently failing rules, to be evaluated on a case-by-case basis for future inclusion.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,7 +60,6 @@ linters:
     - prealloc
     - promlinter
     - protogetter
-    - revive
     - scopelint # Deprecated
     - structcheck # Deprecated
     - stylecheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -83,7 +83,150 @@ linters-settings:
       alias: metav1
     - pkg: "github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator/apis/monitoring/v1"
       alias: monitoringv1
+  revive:
+    enable-all-rules: true
+    rules:
+      # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#add-constant
+      - name: add-constant
+        severity: warning
+        disabled: true
+        arguments:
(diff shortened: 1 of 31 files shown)


cmd/frontend/main.go
```diff
@@ -153,7 +153,7 @@ func main() {
 			//nolint:errcheck
```
macxamin, line 153: Is there a plan to remove these?

macxamin (approved):
LGTM

.golangci.yml
```diff
@@ -84,7 +83,150 @@ linters-settings:
...
+        severity: warning
+        disabled: true
+      # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#useless-break
+      - name: useless-break
+        severity: warning
+        disabled: true
 issues:
+  exclude-use-default: false
+  exclude:
+    # errcheck: Almost all programs ignore errors on these functions and in most cases it's ok
```
macxamin, line 228:
```suggestion
    # errcheck: Almost all programs ignore errors on these functions and in most cases it's ok.
```

cmd/frontend/main.go
bernot-dev: This is a weird one. The logger can technically return an error, which is not being checked. Whether that's a useful error check is debatable. So, we have a few options:

1. Check for errors on logging, when using this logger.
2. Leave the `//nolint` annotations in place.
3. Add an exception to recognize this pattern of error and ignore it.
4. Use a different logging library that doesn't return errors.

Not sure which I would prefer yet, personally.

PR #843, opened 2024-02-14
style: enable usestdlibvars linter

bernot-dev:
usestdlibvars: A linter that detect the possibility to use variables/constants from the Go standard library.


macxamin (approved):
💪

PR #844, opened 2024-02-14
refactor: add command that runs e2e tests without rebuilding images

TheSpiritXIII:
Primarily to get faster local runs!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ endif
 ifeq ($(KIND_PERSIST), 1)
 E2E_DOCKER_ARGS += --env KIND_PERSIST=1
 endif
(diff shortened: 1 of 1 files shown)


pintohutch (approved):
LGTM

FYI Docker caches built images, so it shouldn't be re-building images every time. But it does save a couple of cache hits

Without `e2e-only`:
`TEST_RUN=TestRuleEvaluator make e2e  21.35s user 12.70s system 23% cpu 2:26.86 total`

With `e2e-only`:
`TEST_RUN=TestRuleEvaluator make e2e-only  2.62s user 0.65s system 2% cpu 2:01.84 total`

~25s speed up from a quick benchmark

TheSpiritXIII:
Thanks for the benchmarking!

PR #845, opened 2024-02-14
cleanup: use exact tags in e2e tests

pintohutch:
Before we were implicitly using the `latest` tags, which aren't guaranteed to be the same tags of the "currently built" images for the test (i.e. it's possible another process is building a `latest` tag). This is pretty rare in practice, but this tightens up the logic to avoid the situation entirely.


pintohutch:
Meh - can't figure out how to do this right now, but it's something we may want to do at some point.

PR #852, opened 2024-02-16
refactor(e2e): use consistent tag name when building images at the same time

TheSpiritXIII:
tl;dr: this ensures that when you run `make e2e`, all Docker images are built with the same tag.

Another try for https://github.com/GoogleCloudPlatform/prometheus-engine/pull/845 (thanks @pintohutch)

## How this works
1. The problem is we call `$(MAKE)` which calls and forks `make` without carrying variables over. This is fixed by using `export`.
2. The second problem is that `make` has rules for variable expansion. `=` and `?=` are considered "recursively expansions" which you can think of as lazily evaluated. `:=` is considered "simply expanded" and is evaluated at declaration time. This is what we want to ensure it's only evaluated once.
3. Now that we export the value correctly to targets, we need to make sure we don't override it, so we add the conditional.
4. We still use the latest tag for e2e tests because we need this for the `e2e-only` target which does not rebuild all images. This needs to be conditional so I'll fix this in a separate PR.
5. I added nanoseconds to the tag which helps with testing but also better so we reduce the odds of mutating tags for older images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,10 @@ DOCKER_HOST?=unix:///var/run/docker.sock
 DOCKER_VOLUME:=$(DOCKER_HOST:unix://%=%)
 
 IMAGE_REGISTRY?=us-east4-docker.pkg.dev/$(PROJECT_ID)/prometheus-engine
-TAG_NAME?=$(shell date "+gmp-%Y%d%m_%H%M")
+ifeq ($(origin TAG_NAME), undefined)
+  TAG_NAME:=$(shell date "+gmp-%Y%d%m_%H%M%S_%N")
+endif
+export TAG_NAME
 
 # If an individual test is not specified, run them all.
 TEST_RUN?=$(shell go test ./e2e/... -list=. | grep -E 'Test*')
@@ -46,16 +49,11 @@ SED?=$(shell which gsed 2>/dev/null || which sed)
 # TODO(pintohutch): this is a bit hacky, but can be useful when testing.
 # Ultimately this should be replaced with go templating.
 define update_manifests
-	find manifests examples -type f -name "*.yaml" -exec sed -i "s#image: .*/$(1):.*#image: ${IMAGE_REGISTRY}/$(1):${TAG_NAME}#g" {} \;
+	find manifests examples -type f -name "*.yaml" -exec sed -i "s#image: .*/$(1):.*#image: ${IMAGE_REGISTRY}/$(1):$(TAG_NAME)#g" {} \;
 endef
 
 define docker_build
 	DOCKER_BUILDKIT=1 docker build --label "part-of=gmp" $(1)
-endef
-
-define docker_tag_push
-	docker tag $(1) $(2)
-	docker push $(2)
 endef
 
 define ensure_registry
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -31,7 +31,10 @@ DOCKER_HOST?=unix:///var/run/docker.sock
 DOCKER_VOLUME:=$(DOCKER_HOST:unix://%=%)
 
 IMAGE_REGISTRY?=gcr.io/$(PROJECT_ID)/prometheus-engine
-TAG_NAME?=$(shell date "+gmp-%Y%d%m_%H%M")
+ifeq ($(origin TAG_NAME), undefined)
+  TAG_NAME:=$(shell date "+gmp-%Y%d%m_%H%M%S_%N")
+endif
+export TAG_NAME
```
bernot-dev, line 37: Can we just do this?
```
BUILD_DATE:=$(shell date "+%Y%d%m_%H%M%S_%N")
export TAG_NAME?=gmp-${BUILD_DATE}
```

Makefile
```diff
@@ -46,16 +49,19 @@ SED?=$(shell which gsed 2>/dev/null || which sed)
...
 endef
 
 define docker_build
 	DOCKER_BUILDKIT=1 docker build --label "part-of=gmp" $(1)
 endef
 
-define docker_tag_push
-	docker tag $(1) $(2)
-	docker push $(2)
+define docker_tag
```
bernot-dev, line 59: Suggestion: Does using `define` this add any value when the command is one line? Maybe just use the command directly.

Makefile
```diff
@@ -113,9 +119,11 @@ ifeq ($(NO_DOCKER), 1)
 	CGO_ENABLED=0 go build -tags builtinassets -mod=vendor -o ./build/bin/$(BIN_GO_NAME) ./$(BIN_GO_DIR)/$(BIN_GO_NAME)/*.go
 # If pushing, build and tag native arch image to GCR.
 else ifeq ($(DOCKER_PUSH), 1)
-	$(call docker_build, --tag gmp/$(BIN_GO_NAME) -f ./$(BIN_GO_DIR)/$(BIN_GO_NAME)/Dockerfile .)
+	$(call docker_build, --tag "gmp/$(BIN_GO_NAME):$(TAG_NAME)" -f "./$(BIN_GO_DIR)/$(BIN_GO_NAME)/Dockerfile" .)
 	@echo ">> tagging and pushing images"
-	$(call docker_tag_push,gmp/$(BIN_GO_NAME),${IMAGE_REGISTRY}/$(BIN_GO_NAME):${TAG_NAME})
+	$(call docker_tag, "gmp/$(BIN_GO_NAME):$(TAG_NAME)" "gmp/$(BIN_GO_NAME):latest")
+	$(call docker_tag, "gmp/$(BIN_GO_NAME):$(TAG_NAME)" "${IMAGE_REGISTRY}/$(BIN_GO_NAME):$(TAG_NAME)")
```
bernot-dev, line 125: Suggestion: It's possible to do [multiple tags with one build command](https://docs.docker.com/reference/cli/docker/image/build/#tag). No need to call tag separately for this case.

TheSpiritXIII: That's `docker build`. `docker tag` does not support multiple tags.

Makefile
TheSpiritXIII: Good point!

Makefile
TheSpiritXIII: I believe so. They point out it is equivalent in: https://www.gnu.org/software/make/manual/html_node/Setting.html

TheSpiritXIII: I investigated this and they're not 1:1.

With yours, we will always evaluate `BUILD_TIME`. If someone uses it and calls `$(MAKE)`, the value will be different in each execution in each target.

The reason why mine works is because my if statement says if the variable is not set, then set it. In your example, `TAG_NAME` is lazily evaluated which also means you also can't simply as `export TAG_NAME?=gmp-$(shell date "+%Y%d%m_%H%M%S_%N")`, since it's re-evaluated each time you call `$(MAKE)`. While your version is simpler I feel it's more error-prone in these 2 ways.

As such, I kept my original version. Any strong thoughts here?

PR #862, opened 2024-02-20
chore: release from release branches

bernot-dev:
In addition to releasing from `main`, release on release branches.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -16,6 +16,7 @@ on:
   push:
     branches:
       - main
+      - release/**
 
 permissions:
   contents: write


TheSpiritXIII:
I released from the `main` branch before. It is frail because subsequent patches will need a release branch, but it makes life easier at the moment. Do we want maybe both or are we sure we want to lock it down to release branches?

bernot-dev:
Yeah, maybe we should do both so we have the first tag of a release is an ancestor of `main`.

With releases from `main`:

1. Merge release PR. First tag is on `main` history.
2. Create release branch, with release commit at the root of the branch.
3. Add additional commits to release branch, as necessary. Release again on branch.

With releases only from `release/` branches:

1. Create release branch from main.
2. Merge release PR, release is not on `main`.
3. Add additional commits to release branch, as necessary. Release again on branch.

I'll make that change.

TheSpiritXIII (approved):
Thanks! I agree with your reasoning for branches.

PR #863, opened 2024-02-20
chore: automate export.go version bump

bernot-dev:
Automatically bump the version in `export.go` with release-please.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,0 +1,5 @@
+{
+    "extra-files": [
+        "pkg/export/export.go"
+    ]
+}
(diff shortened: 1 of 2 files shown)


bernot-dev:
Hm...the version update would be in the PR, but the tag doesn't get created until the PR is merged. I wonder if the test needs to be reconsidered.

TheSpiritXIII:
As discussed offline, could you remove the test now that we have this change? I believe you can just return empty string (and the comment) in the `if Testing()` block.

TheSpiritXIII (approved):
LGTM!

PR #864, opened 2024-02-21
feat(prometheus): add secret management to export library

TheSpiritXIII:
Original code review here: https://github.com/GoogleCloudPlatform/prometheus/pull/130

Prometheus change here: https://github.com/GoogleCloudPlatform/prometheus/pull/135


pkg/export/secrets/watch.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/go-kit/log"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+)
+
+type WatchSPConfig struct {
```
pintohutch, line 32: nit (non-blocking): can we add a TODO here about a polling implementation, given watches from a DaemonSet aren't scalable?

pkg/export/secrets/manager.go
```diff
@@ -0,0 +1,223 @@
...
+			errs = append(errs, fmt.Errorf("duplicate secret key %q", secret.Name))
+			secretNamesEnabled[secret.Name] = false
+		} else {
+			secretNamesEnabled[secret.Name] = true
+		}
+	}
+
+	newSecrets := map[string]*secretEntry{}
+	for i := range configs {
+		secret := &configs[i]
```
macxamin, line 158: What about using `incomingSecret`? I was having a hard time tracking all the secrets variables.

pkg/export/secrets/manager.go
```diff
@@ -0,0 +1,223 @@
...
+				continue
+			}
+			s, err := m.provider.Update(ctx, &secretConfig.config, &secret.Config)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			secretConfig.secret = s
+			newSecrets[secret.Name] = secretConfig
+			continue
```
macxamin, line 180: nit: use else instead of continue. Reads better. 
if it exists{ 
// modify
} else {
// add it
}

macxamin (changes requested):
💪 Nice,  I added some nits.

pkg/export/secrets/manager.go
```diff
@@ -0,0 +1,223 @@
...
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	failedSecretConfigs = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "prometheus_sp_failed_secret_configs",
```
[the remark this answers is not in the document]
TheSpiritXIII: I'm following the Prometheus convention here:
`sd` = service discovery
`sp` = secret provider
I'll find a clearer name unique to our fork. Thanks!

TheSpiritXIII:
Thanks for the review, @macxamin! Let me know if there's anything else that isn't clear that I could add comments to.

macxamin (approved):
Great job! thanks!

pkg/export/secrets/manager.go
```diff
@@ -0,0 +1,233 @@
...
+		ctx:      ctx,
+		cancelFn: func() {},
+		secrets:  make(map[string]*secretEntry),
+		opts:     opts,
+	}
+}
+
+// ApplyConfig applies the new secrets, diffing each one with the last configuration to apply the
+// relevant update.
+func (m *Manager) ApplyConfig(ctx context.Context, providerConfig *WatchSPConfig, configs []SecretConfig) error {
```
pintohutch, line 106: question: Do we need a mutex lock here? I see that [discovery](https://github.com/GoogleCloudPlatform/prometheus/blob/TheSpiritXIII/secret-management/discovery/manager.go#L195-L199) uses it, presumably because the config can be reloaded, concurrently, through [various methods](https://github.com/GoogleCloudPlatform/prometheus/blob/TheSpiritXIII/secret-management/cmd/prometheus/main.go#L1016-L1032).

pkg/export/secrets/watch.go
```diff
@@ -0,0 +1,241 @@
...
+func (c *WatchSPConfig) newProvider(ctx context.Context, opts ProviderOptions) (*watchProvider, error) {
+	client, err := c.ClientConfig.client()
+	if err != nil {
+		return nil, err
+	}
+	return newWatchProvider(ctx, opts.Logger, client)
+}
+
+type watcher struct {
+	mu       sync.Mutex
```
pintohutch, line 48: suggestion: Can you add a comment on why this mutex is needed? How would we see a race here (where's the concurrency)?

pkg/export/secrets/watch.go
```diff
@@ -0,0 +1,241 @@
...
+	w        watch.Interface
+	refCount uint
+	s        *corev1.Secret
+}
+
+func newWatcher(ctx context.Context, logger log.Logger, client kubernetes.Interface, config *KubernetesSecretConfig) (*watcher, error) {
+	val := &watcher{
```
pintohutch, line 55: nitpick (non-blocking): variable name `val` is a little generic, and hard to track. Maybe `watcher` and name the struct like `secretsWatcher`, i.e. `watcher := secretsWatcher{refCount: 1}`?

pkg/export/secrets/watch.go
```diff
@@ -0,0 +1,241 @@
...
+
+			// Run within separate function so we can use defer.
+			finished := func() bool {
+				// Check in case the channel cancelled intentionally.
+				if val.refCount == 0 {
+					return true
+				}
+
+				// Pseudo-arbitrarily jitter the length of the most common scrape interval.
+				time.Sleep(1*time.Second + time.Duration(rand.Intn(30)))
```
pintohutch, line 102: question: What's the utility of this sleep? And you're sleeping 1s + [0,30) [nanoseconds](https://pkg.go.dev/time#Duration)?

pkg/export/secrets/watch.go
```diff
@@ -0,0 +1,241 @@
...
+						val.mu.Lock()
+						defer val.mu.Unlock()
+						val.w.Stop()
+					}()
+					return
+				}
+			}
+
+			// Run within separate function so we can use defer.
+			finished := func() bool {
```
pintohutch, line 95: question: What's this routine doing, qualitatively? Some cleanup? Restarting the watch? Why?

suggestion: And can you add comments to explain?

pkg/export/secrets/manager.go
```diff
@@ -0,0 +1,233 @@
...
+
+	eq, err := yamlEqual(m.config, providerConfig)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		m.config = providerConfig
+	}()
+
```
pintohutch, line 125: suggestion: add a comment here explaining this logic (why are we reinitializing the manager)?

pkg/export/secrets/manager.go
```diff
@@ -0,0 +1,233 @@
...
+	m.secrets = secretsFinal
+
+	total := len(secretNamesEnabled)
+	success := len(m.secrets)
+	failedSecretConfigs.Set(float64(total - success))
+	secretsTotal.Set(float64(total))
+	return errors.Join(errs...)
+}
+
+// Fetch implements Secret.Fetch.
```
pintohutch, line 217: question: Is this true? `Secret.Fetch` just takes in a `context.Context` argument.

pkg/export/secrets/watch.go
```diff
@@ -0,0 +1,241 @@
...
+					if !ok {
+						break channelLoop
+					}
+					val.update(logger, e)
+				case <-ctx.Done():
+					// Run within separate function so we can use defer.
+					func() {
+						val.mu.Lock()
+						defer val.mu.Unlock()
+						val.w.Stop()
```
pintohutch, line 88: suggetion: IIUC the `Stop()` call here should signal a close of the `ResultChan()` above, so the `!ok` clause above should catch this case. i.e. you shouldn't need the extra `return` call here unless you want to skip the `finished` checking below.

pkg/export/secrets/kubernetes_test.go
```diff
@@ -0,0 +1,548 @@
...
+}
+
+type testCategory struct {
+	name    string
+	secret  *corev1.Secret
+	entries []entry
+}
+
+// Simpler re-implementation of the same method in the k8s library.
+func pollUntilContextTimeout(ctx context.Context, interval, timeout time.Duration, condition func(context.Context) (done bool, err error)) error {
```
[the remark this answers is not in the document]
TheSpiritXIII: Prometheus upstream (2.43) did not have that method. It used k8s libraries which were too old. This is no longer an issue. Our 2.45 fork and upstream are both on v0.29 now so that method is available. I'll replace it with the built-in one. Thanks!

pkg/export/secrets/watch.go
TheSpiritXIII: Whoops, this was meant to be [1, 31] seconds. This is some jitter because if the Kubernetes API server goes down, we don't want to send potentially dozens of watches at once across all the pods in the collector DaemonSet.

pkg/export/secrets/watch.go
```diff
@@ -0,0 +1,241 @@
...
+	return val.secret(config), nil
+}
+
+// Update updates the secret, restarting the watch if the key changes.
+func (p *watchProvider) Update(ctx context.Context, configBefore, configAfter *KubernetesSecretConfig) (Secret, error) {
+	secretBefore := configBefore.objectKey()
+	secretAfter := configAfter.objectKey()
+	if secretBefore == secretAfter {
+		// If we're using the same secret with a different key, just remap your current watch.
+		val := p.secretKeyToWatcher[secretAfter.String()]
```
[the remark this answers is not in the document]
TheSpiritXIII: Yes. This is just the `client.ObjectKey` (namespace/name) of the secret. If the new key does not exist, then there is an error at fetch-time.

pkg/export/secrets/manager.go
TheSpiritXIII: You are right, it actually implements `github.com/prometheus/common/config.SecretManager.Fetch`.

pkg/secrets/watch.go
```diff
@@ -0,0 +1,266 @@
...
+		return val.secret(configAfter), nil
+	}
+	if err := p.Remove(configBefore); err != nil {
+		return nil, err
+	}
+	return p.Add(configAfter)
+}
+
+// Remove removes the secret, stopping the watch if no other keys for the same secret are watched.
+func (p *watchProvider) Remove(config *KubernetesSecretConfig) error {
```
pintohutch, line 244: nit: It doesn't look like this function ever returns an error, just `nil`.

pkg/secrets/watch.go
```diff
@@ -0,0 +1,266 @@
...
+			return "", fmt.Errorf("secret %s/%s not found", config.Namespace, config.Name)
+		}
+		return getKey(w.s, config.Key)
+	})
+	return &fn
+}
+
+func (w *secretWatcher) start(ctx context.Context, client kubernetes.Interface, config *KubernetesSecretConfig) error {
+	var err error
+	w.w, err = client.CoreV1().Secrets(config.Namespace).Watch(ctx, metav1.ListOptions{
```
pintohutch, line 137: question (non-blocking): I'm curious, did you see if it was feasible or beneficial to share a watch among the secrets, instead of initializing a watch per secret?

pkg/secrets/watch.go
```diff
@@ -0,0 +1,266 @@
...
+
+	// We could wait for the first watch event, but it doesn't notify us if the resource doesn't exist.
+	w.s, err = client.CoreV1().Secrets(config.Namespace).Get(ctx, config.Name, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("unable to fetch secret %s/%s: %w", config.Namespace, config.Name, err)
+	}
+	return nil
+}
+
+// restart returns true and restarts the watch if there are any open secrets.
```
pintohutch, line 153: nit: can you clarify the docstring a bit? What does the routine do, and when does it return false vs true?

pkg/secrets/watch.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+type watchProvider struct {
+	ctx                context.Context
+	client             kubernetes.Interface
+	secretKeyToWatcher map[string]*secretWatcher
+	logger             log.Logger
+}
+
+func newWatchProvider(ctx context.Context, logger log.Logger, client kubernetes.Interface) (*watchProvider, error) {
```
pintohutch, line 197: suggestion: this doesn't ever return an error. remove that return arg. Or why not just call directly in the `newProvider` call above? i.e.
```go
func (c *WatchSPConfig) newProvider(ctx context.Context, opts ProviderOptions) (*watchProvider, error) {
	client, err := c.ClientConfig.client()
	if err != nil {
		return nil, err
	}
	return &watchProvider{
		ctx:                ctx,
		client:             client,
		secretKeyToWatcher: map[string]*secretWatcher{},
		logger:             logger,
	}, nil
}
```

pintohutch (approved):
LGTM! % nit suggestions/questions.

None are blocking as they're mostly style suggestions.

pkg/secrets/watch.go
TheSpiritXIII: No, because we would need RBAC permissions to watch all secrets in the entire cluster.

pkg/secrets/watch.go
TheSpiritXIII: `newWatchProvider` is a separate method because it is used by the unit tests, where it is initialized using a fake Kubernetes API client.

PR #868, opened 2024-02-23
chore: specify when frontend returns an error due to user cancelation

macxamin:
Whenever a user cancels a request to GCM via front end they see the error messages `requesting GCM failed` and `context canceled`. These error messages are unclear and it makes it seem like GCM or frontend failed. Users usually see this error because the program they are using to call frontend has a timeout. 
https://github.com/GoogleCloudPlatform/prometheus-engine/assets/89922129/3485cee1-ec0d-4c54-baf8-2f70b6b48702

We have received four support tickets about this error message. This change returns a more direct and actionable error message.

### DIFF
--- a/cmd/frontend/main.go
+++ b/cmd/frontend/main.go
@@ -21,6 +21,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


cmd/frontend/main.go
```diff
@@ -223,8 +224,13 @@ func forward(logger log.Logger, target *url.URL, transport http.RoundTripper) ht
 		resp, err := client.Do(newReq)
 		if err != nil {
 			//nolint:errcheck
-			level.Warn(logger).Log("msg", "requesting GCM failed", "err", err)
-			w.WriteHeader(http.StatusInternalServerError)
+			if errors.Is(err, context.Canceled) {
+				level.Warn(logger).Log("msg", "request to GCM was canceled by the caller of frontend. If a program made the request, consider increasing the timeout", "err", err)
+				w.WriteHeader(http.StatusBadRequest)
+			} else {
+				level.Warn(logger).Log("msg", "requesting GCM failed", "err", err)
```
realschwa, line 231: Can we change this default error message -- and the one above -- to explicitly state that the request was not sent to GoogleCloudMonitoring at all?

macxamin: According to my testing, it gets sent. I created a fake server and was sending/canceling requests to and I can see the fake server get the request.

PR #879, opened 2024-03-04
refactor: cleanup rule-evaluator Helm chart

TheSpiritXIII:
See comparison: https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.10.0-rc.2/manifests/rule-evaluator.yaml

- Added indentation to templates [as per best practices](https://helm.sh/docs/chart_best_practices/templates/#formatting-templates).
- Added standard labels to all resources [as per best practices](https://helm.sh/docs/chart_best_practices/labels/#standard-labels).
- Added a flag `noCommonLabels` which disables standard labels. Setting this flags to `true` gives you our original labels.
- Fixed our deployment pod now correctly uses the `AppVersion` variable.
- We're now using the previously-defined `serviceAccount` variables.
- Added a `rbac` variable [as per best practices](https://helm.sh/docs/chart_best_practices/rbac/#yaml-configuration).
- Fixed consistency around quoting our version. I decided for no quotes because:
  1. We're using SemVer, so YAML should never interpret the value as a floating point, and;
  2. We never used to quote it.

### DIFF
--- a/charts/rule-evaluator/Chart.yaml
+++ b/charts/rule-evaluator/Chart.yaml
@@ -15,9 +15,6 @@
 apiVersion: v2
 name: rule-evaluator
 description: Prometheus Rule Evaluator
-
 type: application
-
 version: 0.1.0
-
-appVersion: "0.9.0"
+appVersion: 0.9.0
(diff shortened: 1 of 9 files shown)


macxamin (approved):
LGTM

PR #882, opened 2024-03-06
fix: remove NodeMonitoring from release-0.10

macxamin:
NodeMonitoring was mistakenly added as namespace scoped. It can't be changed to cluster scoped, so we are removing it from release-0.10 and we will reintroduce it in 0.11 as NodeClusterMonitoring.


TheSpiritXIII (approved):
Thanks! Looks good.

PR #884, opened 2024-03-06
feat: export write from collectors to central prometheus example

macxamin:
This is an example of how GMP collectors can be configured to send metrics to a central Prometheus receiver.

1. Added a URL flag to OperatorConfig, this flag is then used to set `remote_write` in the Prometheus config.
2. examples/export-write contains the manifest to launch a central Prometheus that accepts incoming write requests.

### DIFF
--- a/pkg/operator/apis/monitoring/v1/operator_types.go
+++ b/pkg/operator/apis/monitoring/v1/operator_types.go
@@ -29,8 +29,12 @@ type OperatorConfig struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	// Rules specifies how the operator configures and deploys rule-evaluator.
 	Rules RuleEvaluatorSpec `json:"rules,omitempty"`
-	// Collection specifies how the operator configures collection.
+	// Collection specifies how the operator configures collection including
+	// integrated exports to Google Cloud Monitoring.
 	Collection CollectionSpec `json:"collection,omitempty"`
+	// Exports specifies additional exporting mechanism on top of Google Cloud Monitoring collection.
+	// Note: To disable integrated export to Google Cloud Monitoring specify a non-matching filter in Collection.
+	Exports []ExportSpec `json:"exports,omitempty"`
 	// ManagedAlertmanager holds information for configuring the managed instance of Alertmanager.
 	// +kubebuilder:default={configSecret: {name: alertmanager, key: alertmanager.yaml}}
 	ManagedAlertmanager *ManagedAlertmanagerSpec `json:"managedAlertmanager,omitempty"`
@@ -73,10 +77,10 @@ type RuleEvaluatorSpec struct {
 // CollectionSpec specifies how the operator configures collection of metric data.
 type CollectionSpec struct {
 	// ExternalLabels specifies external labels that are attached to all scraped
-	// data before being written to Cloud Monitoring. The precedence behavior matches that
-	// of Prometheus.
+	// data before being written to Google Cloud Monitoring or any other additional exports
+	// specified in the OperatorConfig. The precedence behavior matches that of Prometheus.
 	ExternalLabels map[string]string `json:"externalLabels,omitempty"`
-	// Filter limits which metric data is sent to Cloud Monitoring.
+	// Filter limits which metric data is sent to Cloud Monitoring(Doesn't apply to additional exports).
 	Filter ExportFilters `json:"filter,omitempty"`
 	// A reference to GCP service account credentials with which Prometheus collectors
 	// are run. It needs to have metric write permissions for all project IDs to which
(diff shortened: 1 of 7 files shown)


pkg/operator/apis/monitoring/v1/operator_types.go
```diff
@@ -88,6 +88,13 @@ type CollectionSpec struct {
 	KubeletScraping *KubeletScraping `json:"kubeletScraping,omitempty"`
 	// Compression enables compression of metrics collection data
 	Compression CompressionType `json:"compression,omitempty"`
+	// Export enables exporting metrics to an endpoint.
+	Export ExportSpec `json:"export,omitempty"`
+}
+
+type ExportSpec struct {
```
bwplotka, line 95: Nice, this unblocks simple remote write cases, but I wonder if we could clear up some consistency/functionality questions before moving forward: 

1. Should add type of export to be explicit if this is PRW, Otel or GCM API? If not (we can add type field later on, once use case appears) we could claim it PRW for now, but let's then specify that in commentary. Also let's mention this is ON TOP of automatic Cloud Monitoring export.
2. To be future proof, I wonder if Exports []ExportSpec wouldn't be more appropriate.
3. What's the relation of ExportSpec.URL to `Filter` `ExternalLabel` `Credentials` `Compression` etc? Filter would be nice to have (some translation to relabelling would be necessary), but wonder if `exports` should use those collection fields, or have their own? For filtering, I would say per export make more sense from user perspective, but open to anything. For external label e.g. we can't really do that per export as it's done per collector. Ideally we somehow separate what's per export, what's "global/cluster wide".

macxamin: PTAL

pkg/operator/apis/monitoring/v1/operator_types.go
```diff
@@ -73,10 +77,10 @@ type RuleEvaluatorSpec struct {
 // CollectionSpec specifies how the operator configures collection of metric data.
 type CollectionSpec struct {
 	// ExternalLabels specifies external labels that are attached to all scraped
-	// data before being written to Cloud Monitoring. The precedence behavior matches that
-	// of Prometheus.
+	// data before being written to Google Cloud Monitoring or any other additional exports
+	// specified in the OperatorConfig. The precedence behavior matches that of Prometheus.
 	ExternalLabels map[string]string `json:"externalLabels,omitempty"`
-	// Filter limits which metric data is sent to Cloud Monitoring.
+	// Filter limits which metric data is sent to Cloud Monitoring(Doesn't apply to additional exports).
```
bwplotka, line 83:
```suggestion
	// Filter limits which metric data is sent to Cloud Monitoring (it doesn't apply to additional exports).
```

pkg/operator/apis/monitoring/v1/operator_types.go
```diff
@@ -90,6 +94,11 @@ type CollectionSpec struct {
 	Compression CompressionType `json:"compression,omitempty"`
 }
 
+type ExportSpec struct {
+	// The URL of the endpoint to export samples to.
```
bwplotka, line 98:
```suggestion
	// The URL of the endpoint that supports Prometheus Remote Write to export samples to.
```

pkg/operator/apis/monitoring/v1/operator_types.go
```diff
@@ -29,8 +29,12 @@ type OperatorConfig struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	// Rules specifies how the operator configures and deploys rule-evaluator.
 	Rules RuleEvaluatorSpec `json:"rules,omitempty"`
-	// Collection specifies how the operator configures collection.
+	// Collection specifies how the operator configures collection including
+	// integrated exports to Google Cloud Monitoring.
```
bwplotka, line 33:
```suggestion
	// Collection specifies how the operator configures collection, including
	// scraping and an integrated export to Google Cloud Monitoring.
```

pkg/operator/apis/monitoring/v1/operator_types.go
```diff
@@ -29,8 +29,12 @@ type OperatorConfig struct {
...
 	Collection CollectionSpec `json:"collection,omitempty"`
+	// Exports specifies additional exporting mechanism on top of Google Cloud Monitoring collection.
+	// Note: To disable integrated export to Google Cloud Monitoring specify a non-matching filter in Collection.
```
bwplotka, line 36:
```suggestion
	// Exports specifies additional, optional endpoints to export to, on top of Google Cloud Monitoring collection.
	// Note: To disable integrated export to Google Cloud Monitoring specify a non-matching filter in the "collection.filter" field.
```

bwplotka (approved):
LGTM, just nits to commentary 💪🏽

PR #763, opened 2024-01-09
OperatorConfig compression setting is now also applied for rule-eval. configMap from *Rules CRs [0.9]

bwplotka:
Fixes https://github.com/GoogleCloudPlatform/prometheus-engine/issues/549

PR #891, opened 2024-03-14
feat(rule-evaluator): add runtime status endpoint

TheSpiritXIII:
This helps debug whether the configuration was reloaded successfully. The endpoint matches the Prometheus container's endpoint.

In the future, we will use this in e2e tests to help prevent situations such as: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/763


bwplotka (approved):
Thanks, super useful 💪🏽

Issue #765, opened 2024-01-10
rule-eval tests should check if rules are loaded to rule-eval, not only config map


bwplotka:
When checking testability of https://github.com/GoogleCloudPlatform/prometheus-engine/pull/763 I noticed we don't test on CI (when GCM option is off) if rules are ever loaded. Let's improve it e.g. by having a flag to skip readiness check of rule-eval for test purposes and checking API/metrics.

PR #892, opened 2024-03-14
test(e2e): check whether rule-evaluator reload was successful

TheSpiritXIII:
Resolves https://github.com/GoogleCloudPlatform/prometheus-engine/issues/765 by checking both rule-evaluator status and config-reloader logs.

Most code was simply revived from the original e2e tests: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/747

Help prevent situations such as: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/763


TheSpiritXIII:
Investigating test failures...

pintohutch (commented):
Fly-by (non-blocking):
1. It may ultimately be easier to e2e test if the operator gets the reload status from the rule-evaluator and populates a `.Status` section of a CRD, e.g. the `OperatorConfig`. It could also be a nice debug signal for users. Note: the reload failure mode is also possible in collectors and the alertmanager.
2. Is it simpler to just update the rules and see whether they landed in GCM? Obviously we don't have GCM tests in Github (I wonder if we should revisit that in general..), but it's ultimately the signal we care about.

TheSpiritXIII:
1. That's a really great idea, and it would help out Autopilot users who are forbidden from port-forwarding the pod and hitting the endpoint directly. I think this warrants a deeper design discussion, though. For example, why reload status in the OperatorConfig and why not rule status in the individual Rule objects which would be more granular?
2. That's indeed true. My concern is that if GCM testing is disabled in e2e tests by default, it will not be caught. The other benefit of these tests is that they're more specific: you know which component (rule-evaluator vs. config-reloader vs. gcm) was responsible for the failure.

Ultimately, I would still like to move forward with these tests to unblock testing the rule-evaluator compression changes. Once we make a more concrete decision, it will be trivial to remove these tests.

TheSpiritXIII:
Converting to draft until https://github.com/GoogleCloudPlatform/prometheus-engine/pull/894 is merged!

TheSpiritXIII:
Ready for review now!

e2e/ruler_test.go
```diff
@@ -441,3 +463,292 @@ func createRuleEvaluatorOperatorConfig(ctx context.Context, opClient versioned.I
 	}
 	return nil
 }
+
+func testRulesGeneration(ctx context.Context, restConfig *rest.Config, kubeClient client.Client, clientSet kubernetes.Interface, systemNamespace, userNamespace string) func(*testing.T) {
+	return func(t *testing.T) {
+		timeStart := time.Now()
+		replace := strings.NewReplacer(
```
bernot-dev, line 470: Cool, I haven't used `strings.Replacer` before.

bernot-dev (approved):
LGTM. Just a few non-blocking comments.

PR #894, opened 2024-03-15
fix(e2e): start rule-evaluator without auth in e2e tests

TheSpiritXIII:
Adds a new flag to the rule-evaluator `--query.debug.disable-auth` which mirrors `--export.debug.disable-auth` but for querying rather than writing.

This flag is then enabled on e2e tests which lets the rule-evaluator start without requiring credentials.

Previously, we hacked this by creating a default credentials file, but this was removed in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/813

### DIFF
--- a/e2e/deploy/deploy.go
+++ b/e2e/deploy/deploy.go
@@ -38,40 +38,9 @@ func CreateResources(ctx context.Context, kubeClient client.Client, deployOpts .
 	return createResources(ctx, kubeClient, func(obj client.Object) (client.Object, error) {
 		switch obj := obj.(type) {
 		case *appsv1.DaemonSet:
-			if !opts.disableGCM {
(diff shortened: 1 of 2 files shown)


e2e/deploy/deploy.go
```diff
@@ -51,7 +51,8 @@ func CreateResources(ctx context.Context, kubeClient client.Client, deployOpts .
 				}
 			}
 		case *appsv1.Deployment:
-			if obj.GetName() == operator.NameOperator {
+			switch obj.GetName() {
```
bernot-dev, line 54: suggestion(if-minor): Avoid nested `switch` statements. Maybe a helper function or something?

bernot-dev (approved):

PR #897, opened 2024-03-18
feat: apply OperatorConfig compression setting for rule-eval. configMap from *Rules CRs

TheSpiritXIII:
Fixes https://github.com/GoogleCloudPlatform/prometheus-engine/issues/549

Reopened https://github.com/GoogleCloudPlatform/prometheus-engine/pull/763

Depends on:
- [x] https://github.com/GoogleCloudPlatform/prometheus-engine/pull/896
- [x] https://github.com/GoogleCloudPlatform/prometheus-engine/pull/892

### DIFF
--- a/e2e/ruler_test.go
+++ b/e2e/ruler_test.go
@@ -15,10 +15,13 @@
 package e2e
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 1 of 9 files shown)


pkg/operator/collection.go
```diff
@@ -320,6 +320,31 @@ func gzipData(data []byte) ([]byte, error) {
 	return b.Bytes(), nil
 }
 
+func setConfigMapData(cm *corev1.ConfigMap, c monitoringv1.CompressionType, fn string, data string) error {
```
pintohutch, line 323: suggestion (non-blocking): `fn` threw me off for a minute as I was reading `function`. Maybe `filename` or `path`?

bernot-dev (approved):
LGTM

e2e/ruler_test.go
```diff
@@ -610,6 +638,22 @@ func testRulesGeneration(ctx context.Context, restConfig *rest.Config, kubeClien
...
+					}
+					uncompressed, err := io.ReadAll(r)
+					if err != nil {
+						t.Fatal(err)
+					}
+					data[fn] = string(uncompressed)
+				}
+			}
 
 			// The operator observes Rules across all namespaces. For the purpose of this test we
```
pintohutch, line 658: question: This seems risky, as we could end up deleting things we shouldn't have written to the configmap - can you explain why this is here?

pintohutch (commented):
Mostly LGTM % question and a nit

e2e/ruler_test.go
TheSpiritXIII: The explanation is that this is not my code so I didn't think much of it. It comes from the initial commit (see [main_test.go](https://github.com/GoogleCloudPlatform/prometheus-engine/commit/5498f6b1f8c6f7627ee3c72731b367e5b9cb2ac0)).

However, you make a great point so I removed that piece of code and consolidated this method with the other create rules method in this file to ensure that `data = want`.

pintohutch (approved):

PR #898, opened 2024-03-18
refactor: restart deployment

bernot-dev:
Restart deployment instead of deleting pods in webhook e2e test.


TheSpiritXIII (approved):
This looks like the right way. Thanks for the change!

PR #904, opened 2024-03-19
chore: enable boringcrypto

bernot-dev:
Enable boringcrypto flag for Go builds.

### DIFF
--- a/cmd/datasource-syncer/Dockerfile
+++ b/cmd/datasource-syncer/Dockerfile
@@ -17,8 +17,9 @@ WORKDIR /app
 COPY . ./
 
 FROM buildbase as appbase
-RUN CGO_ENABLED=0 go build -mod=vendor -o datasource-syncer cmd/datasource-syncer/*.go
+RUN CGO_ENABLED=0 GOEXPERIMENT=boringcrypto go build -mod=vendor -o datasource-syncer cmd/datasource-syncer/*.go
 
 FROM gcr.io/distroless/static-debian11:latest
 COPY --from=appbase /app/datasource-syncer /bin/datasource-syncer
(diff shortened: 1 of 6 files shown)


TheSpiritXIII (changes requested):
Should we add it `examples/instrumentation/go-synthetic/Dockerfile` since we publish that image too?

bernot-dev:
> Should we add it `examples/instrumentation/go-synthetic/Dockerfile` since we publish that image too?

Added.

cmd/datasource-syncer/Dockerfile
```diff
@@ -17,8 +17,8 @@ WORKDIR /app
 COPY . ./
 
 FROM buildbase as appbase
-RUN CGO_ENABLED=0 go build -mod=vendor -o datasource-syncer cmd/datasource-syncer/*.go
+RUN CGO_ENABLED=0 GOEXPERIMENT=boringcrypto go build -mod=vendor -o datasource-syncer cmd/datasource-syncer/*.go
 
 FROM gcr.io/distroless/static-debian11:latest
 COPY --from=appbase /app/datasource-syncer /bin/datasource-syncer
-ENTRYPOINT ["/bin/datasource-syncer"]
```
TheSpiritXIII, line 24: nit (non-blocking): was this newline removed? Let's ensure we have file-ending newlines to ensure tools behave properly (and because it's enforced within Google).

TheSpiritXIII (approved):

cmd/datasource-syncer/Dockerfile
bernot-dev: I don't think that's part of this change, but I added a newline.

pintohutch (approved):

PR #905, opened 2024-03-19
style: enforce apierrors alias

bernot-dev:
Always import "k8s.io/apimachinery/pkg/api/errors" as apierrors.


TheSpiritXIII (approved):
LGTM. I usually always do this, but I must have missed the e2e one! Thanks!

PR #910, opened 2024-03-21
chore: enable cgo and use google buildbase

pintohutch:
cgo needs to be enabled to link against boringcrypto, so we add that here.

In addition, we use the google-go.pkg.dev/golang image as the Go buildbase to ensure build-time requirements, like boringcrypto, are enabled.

We also use gke.gcr.io/gke-distroless/libc as our runtime image.

We add the "cryp/tls/fipsonly" import to ensure boringcrypto is linking properly at build time. We guard this with a build tag "boring" in a dedicated file.

### DIFF
--- a/cmd/rule-evaluator/Dockerfile
+++ b/cmd/rule-evaluator/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.22-bullseye AS buildbase
+FROM google-go.pkg.dev/golang:1.22.1@sha256:4463a06a8378eddc4efa36de9a5310ff087c03d3cc4278c87059209d3d95e4c0 as buildbase
 WORKDIR /app
 COPY go.mod go.mod
 COPY go.sum go.sum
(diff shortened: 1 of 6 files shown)


cmd/rule-evaluator/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.22-bullseye AS buildbase
+FROM google-go.pkg.dev/golang:1.22.0@sha256:564af112166fe661bf3b29e2ffd687868eaa0a7b9f02d7056e3e62f9df991b89 as buildbase
```
bernot-dev, line 15: Question: Is there a way to list the available images in that repo? 1.22.1 should be released, I think.

pintohutch: Yep - using the [`gcrane` tool](https://cloud.google.com/artifact-registry/docs/docker/copy-from-gcr#copy-gcrane), i.e. `gcrane ls google-go.pkg.dev/golang`.

Good catch, updated to 1.22.1

bwplotka (approved):
LGTM, but shouldn't we do the same for rest of our cmds?

There might be a bit different perf characteristics with CGO, but recent resources says it should be [minimal](https://shane.ai/posts/cgo-performance-in-go1.21/). Let's watch that with a bit more care (this is the most relevant to collector perhaps).

pintohutch:
> LGTM, but shouldn't we do the same for rest of our cmds?

Maybe? Though I don't think we need to rush as those are electively deployed by users and not part of the managed suite of binaries. IMO they can be done later.

> There might be a bit different perf characteristics with CGO, but recent resources says it should be [minimal](https://shane.ai/posts/cgo-performance-in-go1.21/). Let's watch that with a bit more care (this is the most relevant to collector perhaps).

Thanks for the link - and good to know! I agree the collector is the most risky. I did some `/prombench`-marking for the PRs there and they looked good for the most part. Agreed we should proceed cautiously to start.

PR #912, opened 2024-03-25
fix(operator): add duplicate job name validation within PM, CPM and CNM

bwplotka:
Related to https://github.com/GoogleCloudPlatform/prometheus-engine/issues/479

Without this, certainly malformed CR could trigger Prometheus config parsing error, distrupting the whole cluster collection config changes.

Also to provide better fix to potentially #479 (vs creating separate CR) we could add e.g. JobNameSuffix field for endpoint, but that's for another PR.

### DIFF
--- a/pkg/operator/apis/monitoring/v1/common_types.go
+++ b/pkg/operator/apis/monitoring/v1/common_types.go
@@ -181,3 +181,18 @@ var invalidLabelCharRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
 func sanitizeLabelName(name string) prommodel.LabelName {
 	return prommodel.LabelName(invalidLabelCharRE.ReplaceAllString(name, "_"))
 }
+
+// validateDistinctJobNames returns error if any job name duplicates.
+// This would also result in an error on Prometheus server config parsing.
+func validateDistinctJobNames(jobs []*promconfig.ScrapeConfig) error {
+	jobToIndex := map[string]int{}
+	for i, job := range jobs {
+		if dup, ok := jobToIndex[job.JobName]; ok {
+			// TODO(bwplotka): Consider adding JobSuffix field or so for https://github.com/GoogleCloudPlatform/prometheus-engine/issues/479
+			return fmt.Errorf("duplicated job name %v for endpoints with index %d and %d;"+
+				"consider creating a separate custom resource (PodMonitoring, etc.) for endpoints that share the same resource name, namespace and port name", job.JobName, dup, i)
+		}
+		jobToIndex[job.JobName] = i
+	}
+	return nil
+}
--- a/pkg/operator/apis/monitoring/v1/node_types_test.go
+++ b/pkg/operator/apis/monitoring/v1/node_types_test.go
@@ -68,11 +68,28 @@ func TestValidateClusterNodeMonitoring(t *testing.T) {
 			fail:        true,
 			errContains: "scrape timeout 2s must not be greater than scrape interval 1s",
 		},
+		{
+			// Regression test for https://github.com/GoogleCloudPlatform/prometheus-engine/issues/479
+			desc: "Duplicated job name",
+			eps: []ScrapeNodeEndpoint{
+				{
+					Interval: "10s",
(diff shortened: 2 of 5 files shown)


maxamins (approved):
LGTM

pintohutch (approved):
nice catch!

pkg/operator/apis/monitoring/v1/node_types_test.go
```diff
@@ -68,11 +68,29 @@ func TestValidateClusterNodeMonitoring(t *testing.T) {
...
+			errContains: "/r1/metrics for endpoints with index 0 and 1. Consider creating a separate resource for each job",
+		},
 	}
 
 	for _, c := range cases {
 		t.Run(c.desc+"", func(t *testing.T) {
 			nm := &ClusterNodeMonitoring{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "r1",
+					Namespace: "ns1",
```
TheSpiritXIII, line 92: chore: ClusterNodeMonitoring should not have a namespace.

pkg/operator/apis/monitoring/v1/pod_types_test.go
```diff
@@ -412,6 +433,10 @@ func TestValidatePodMonitoringCommon(t *testing.T) {
 
 		t.Run(c.desc+"_clusterpodmonitoring", func(t *testing.T) {
 			cm := &ClusterPodMonitoring{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "r1",
+					Namespace: "ns1",
```
TheSpiritXIII, line 438: chore: ClusterPodMonitoring cannot have a namespace.

pkg/operator/apis/monitoring/v1/common_types.go
```diff
@@ -181,3 +181,17 @@ var invalidLabelCharRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
...
 }
+
+// validateDistinctJobNames returns error if any job name duplicates.
+// This would also result in an error on Prometheus server config parsing.
+func validateDistinctJobNames(jobs []*promconfig.ScrapeConfig) error {
+	jobToIndex := map[string]int{}
+	for i, job := range jobs {
+		if dup, ok := jobToIndex[job.JobName]; ok {
+			// TODO(bwplotka): Consider adding JobSuffix field or so for https://github.com/GoogleCloudPlatform/prometheus-engine/issues/479
+			return fmt.Errorf("duplicated job name %v for endpoints with index %d and %d. Consider creating a separate resource for each job", job.JobName, dup, i)
```
TheSpiritXIII, line 192: nit(non-blocking): For users, a job name is abstracted away from them. We construct them rather arbitrarily. For example, why there's no reason why we couldn't encode scrape intervals into them to help make them unique and further abstract them away. As such, I believe this error message may be confusing for them.

What the user actually has to do is create a resource with either a different namespace, name or port depending on the resource type. I wonder if we could spell something like that out here?

TheSpiritXIII (changes requested):
This is a useful change! I left a couple of minor nits to address.

pkg/operator/apis/monitoring/v1/node_types_test.go
bwplotka: Yup, but it has such field, so wanted to ensure we don't rely on this. Happy to remove.

pkg/operator/apis/monitoring/v1/common_types.go
bwplotka: I agree I can give more constructive hints, but I disagree job name is abstracted away. While we generate, validate and sanitize it, users definitely has to be aware of what job means and how is constructed when using those metrics via PromQL/other means. I think it should be fine to mention the job abstraction here.

Essentially they might want to aggregate by job label when alering, checking it when querying, retrieving series to tell what *Monitoring CR configured it (in some way, what application we are monitoring).

Let me improve the message..

TheSpiritXIII (approved):
Thanks! Looks good.

PR #931, opened 2024-04-10
Add healthz and readyz to operator

TheSpiritXIII:
This PR improves the reliability of the operator.

This is useful for moving the operator to the control plane (if we ever do), for customers to avoid the 0-node problem themselves, and for us to simplify our e2e test logic (makes our guitar test more reliant too).

### DIFF
--- a/charts/operator/templates/deployment.yaml
+++ b/charts/operator/templates/deployment.yaml
@@ -68,6 +68,20 @@ spec:
             drop:
             - all
           privileged: false
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            # Note this should match the --probe-addr flag passed in to the operator args.
+            # Default is 18081.
+            port: 18081
+            scheme: HTTP
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            # Note this should match the --probe-addr flag passed in to the operator args.
+            # Default is 18081.
+            port: 18081
+            scheme: HTTP
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -69,6 +69,7 @@ func main() {
 		caCert      = flag.String("ca-cert-base64", "", "The base64-encoded certificate authority.")
 		webhookAddr = flag.String("webhook-addr", ":10250",
 			"Address to listen to for incoming kube admission webhook connections.")
+		probeAddr   = flag.String("probe-addr", ":18081", "Address to outputs probe statuses (e.g. /readyz and /healthz)")
 		metricsAddr = flag.String("metrics-addr", ":18080", "Address to emit metrics on.")
 
 		// Permit the operator to cleanup previously-managed resources that
@@ -103,6 +104,7 @@ func main() {
 		Cluster:           *cluster,
 		OperatorNamespace: *operatorNamespace,
 		PublicNamespace:   *publicNamespace,
+		ProbeAddr:         *probeAddr,
 		TLSCert:           *tlsCert,
 		TLSKey:            *tlsKey,
 		CACert:            *caCert,
--- a/manifests/operator.yaml
+++ b/manifests/operator.yaml
@@ -541,6 +541,20 @@ spec:
             drop:
             - all
           privileged: false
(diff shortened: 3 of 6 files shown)


cmd/operator/main.go
```diff
@@ -69,7 +69,8 @@ func main() {
 		caCert      = flag.String("ca-cert-base64", "", "The base64-encoded certificate authority.")
 		webhookAddr = flag.String("webhook-addr", ":10250",
 			"Address to listen to for incoming kube admission webhook connections.")
-		metricsAddr = flag.String("metrics-addr", ":18080", "Address to emit metrics on.")
+		healthProbeAddr = flag.String("health-probe-addr", ":8080", "Address to outputs health status")
```
bwplotka, line 72: What's wrong with using metric port for this?

bwplotka (changes requested):
Thanks! Generally looks good, but I would suggest using same endpoint as metric endpoint for simplicity (or at least name "heath-probe-address" to show that it also supports readiness.. but why not reusing metric addr?)

cmd/operator/main.go
TheSpiritXIII: Not possible -- we don't serve `metrics-addr` using controller-runtime and there is no way to expose a healthz/readyz handler through controller-runtime. If you feel strongly about this, I propose adding port consolidation to our backlog.

Do you have any suggestions for a name other than "heath-probe-address"? They're called "health endpoints" generally (https://kubernetes.io/docs/reference/using-api/health-checks/) and this is the same language used in controller-runtime.

bwplotka: What about ignoring that controller runtime flow and implementing our own handlers (using started checkers?) in metric-addr port? (maybe stupid suggestion, I didn't look deeper what else that "AddReady/Health*" methods do..

If it does not make sense, then "probes-addr" might do the work, WDYT?

charts/operator/templates/deployment.yaml
```diff
@@ -68,6 +68,20 @@ spec:
             drop:
             - all
           privileged: false
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            # Note this should match the --health-probe-addr flag passed in to the operator args.
```
bwplotka, line 74:
```suggestion
            # Note this should match the --probe-addr flag passed in to the operator args.
```

charts/operator/templates/deployment.yaml
```diff
@@ -68,6 +68,20 @@ spec:
...
+            # Default is 18081.
+            port: 18081
+            scheme: HTTP
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            # Note this should match the --health-probe-addr flag passed in to the operator args.
```
bwplotka, line 81:
```suggestion
            # Note this should match the --probe-addr flag passed in to the operator args.
```

manifests/operator.yaml
```diff
@@ -541,6 +541,20 @@ spec:
             drop:
             - all
           privileged: false
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            # Note this should match the --health-probe-addr flag passed in to the operator args.
```
bwplotka, line 547:
```suggestion
            # Note this should match the --probe-addr flag passed in to the operator args.
```

manifests/operator.yaml
```diff
@@ -541,6 +541,20 @@ spec:
...
+            # Default is 18081.
+            port: 18081
+            scheme: HTTP
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            # Note this should match the --health-probe-addr flag passed in to the operator args.
```
bwplotka, line 554:
```suggestion
            # Note this should match the --probe-addr flag passed in to the operator args.
```

pkg/operator/operator.go
```diff
@@ -114,6 +114,8 @@ type Options struct {
 	// Namespace to which the operator looks for user-specified configuration
 	// data, like Secrets and ConfigMaps.
 	PublicNamespace string
+	// Health serving address.
```
bwplotka, line 117:
```suggestion
	// Health and readiness serving address.
```

bwplotka (approved):
Thanks for addressing all comments! LGTM, added some nits after flag name change.

PR #936, opened 2024-04-17
chore: upgrade golangci-lint

bernot-dev:
A few lints were missed due to bugs in the linter that have been corrected in more recent versions.


TheSpiritXIII (approved):
Awesome!

PR #943, opened 2024-04-22
feat: configurable vpa

bernot-dev:


pkg/operator/operator.go
```diff
@@ -202,6 +208,81 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
...
+			Field: fields.SelectorFromSet(fields.Set{
+				"metadata.namespace": opts.OperatorNamespace,
+				"metadata.name":      NameRuleEvaluator,
+			}),
+		},
+	}
+
+	// In GKE, installing Vertical Pod Autoscaling requires a cluster restart, and therefore it also results in an operator restart.
+
+	// In other environments, the operator may need to be restarted to enable VPA to run the following check again and watch for the objects.
```
bwplotka, line 268: Plus we could somehow watch periodically (we should not for the first version). Let's add this fact to e.g. vpa.enabled field description?

pkg/operator/operator.go
```diff
@@ -202,6 +208,81 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
...
+	var vpaAvailable bool
+	clientset, err := apiextensions.NewForConfig(clientConfig)
+	if err != nil {
+		return nil, fmt.Errorf("create clientset: %w", err)
+	}
+	if _, err := clientset.ApiextensionsV1().CustomResourceDefinitions().Get(context.Background(), "verticalpodautoscalers.autoscaling.k8s.io", metav1.GetOptions{}); err != nil {
+		logger.Info("vertical pod autoscaling is not available")
```
bwplotka, line 275: Can we mention in this log line, what operator is doing with this fact (same below?) e.g.

```suggestion
		logger.Info("vertical pod autoscaling is not available, vpa.enabled config option will not work")
```

pkg/operator/operator.go
```diff
@@ -202,6 +208,81 @@ func New(logger logr.Logger, clientConfig *rest.Config, opts Options) (*Operator
...
+	} else {
+		logger.Info("vertical pod autoscaling available")
```
bwplotka, line 277:
```suggestion
		logger.Info("vertical pod autoscaling available; setting up watcher for VPA in operated namespace")
```

pkg/operator/operator_config.go
```diff
@@ -836,6 +837,9 @@ func (v *operatorConfigValidator) ValidateCreate(_ context.Context, o runtime.Ob
 	if err := validateRules(&oc.Rules); err != nil {
 		return nil, fmt.Errorf("invalid rules config: %w", err)
 	}
+	if oc.Scaling.VPA.Enabled && !v.vpaAvailable {
+		return nil, fmt.Errorf("vertical pod autoscaling is not available")
```
bwplotka, line 841:
```suggestion
		return nil, fmt.Errorf("vertical pod autoscaling is not available (if VPA was enabled and you still see this, than restart the operator)")
```

pkg/operator/scaling.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+func setupScalingController(op *Operator) error {
+	objFilterOperatorConfig := namespacedNamePredicate{
+		namespace: op.opts.PublicNamespace,
+		name:      NameOperatorConfig,
+	}
+
+	err := ctrl.NewControllerManagedBy(op.manager).
```
bwplotka, line 60: complexity nit: Opportunity to return here and wrap if needed on a caller side (no additional context is added here)

pkg/operator/scaling.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func applyVPA(ctx context.Context, c client.Client, namespace string) error {
+	vpa := autoscalingv1.VerticalPodAutoscaler{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      collectorVPAName,
+		},
+	}
+	if _, err := controllerutil.CreateOrUpdate(ctx, c, &vpa, func() error {
```
bwplotka, line 105: nit: if not needed here, let's just return?

pkg/operator/scaling.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func deleteVPA(ctx context.Context, c client.Writer, namespace string) error {
+	vpa := autoscalingv1.VerticalPodAutoscaler{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      collectorVPAName,
+			Namespace: namespace,
+		},
+	}
+	if err := c.Delete(ctx, &vpa); client.IgnoreNotFound(err) != nil {
```
bwplotka, line 145: nit: ditto, `return c.Delete` might be simpler

pkg/operator/scaling_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	autoscalingv1 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+)
+
+// Authorization test?
```
bwplotka, line 30: Left over comment?

bwplotka (approved):
Thanks! LGTM, modulo documentation suggestions and non blocking nits.

I am not an expert in VPA behaviour, so hard for me to validate if its works. Have you tested this vpa configuration manually? It feels difficult to test automatically.. If yes, do you mind sharing a manual test scenario you went through (only so it's easier for others or future yourself to verify this or compare to potential future troubleshooting) in GH description? 🤗 That would be amazing.

pkg/operator/scaling.go
bernot-dev: The error is wrapped, consistent with all of the other controllers in the operator.

pkg/operator/scaling.go
bernot-dev: The first return value is discarded because we don't particularly care about the OperationResult (whether it was created or updated), as long as it's not an error. However, the parent function does not return an OperationResult. So, we cannot return the result directly.

pkg/operator/scaling.go
bernot-dev: We don't want to return a "not found" error. The logic here ignores those.

pkg/operator/operator.go
bernot-dev: Moved this comment to the field description.

pkg/operator/operator.go
bernot-dev: Added detail.

PR #944, opened 2024-04-23
fix: cleanup security validation warnings

TheSpiritXIII:
See internal go/k8s-security-validation-service-user-guide

Two changes:
- The `readOnlyRootFilesystem` flag prevents an attacker from overriding the binary. For the GMP operator, I moved the certificate creation into an empty directory to avoid writing into the root filesystem.
- We were missing `seccompProfile` for rule-evaluator, but we have it everywhere else.

Other warnings are more involved (such as trimming update RBAC permissions).

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -38,6 +38,10 @@ import (
 	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator"
 )
 
+const (
+	defaultTLSDir = "/etc/tls/private"
+)
+
 func main() {
 	var (
 		defaultProjectID string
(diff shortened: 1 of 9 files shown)


cmd/operator/main.go
```diff
@@ -67,6 +71,7 @@ func main() {
 		tlsCert     = flag.String("tls-cert-base64", "", "The base64-encoded TLS certificate.")
 		tlsKey      = flag.String("tls-key-base64", "", "The base64-encoded TLS key.")
 		caCert      = flag.String("ca-cert-base64", "", "The base64-encoded certificate authority.")
+		certDir     = flag.String("cert-dir", defaultTLSDir, "The directory which contains TLS certificates for webhook server.")
```
bwplotka, line 74: You didn't mention this change in the PR description, did you mean opening another PR for this or you meant to essentially implement go/gmp:cert-args for operator? cc @bernot-dev

bwplotka: Hm, would't this change break user, if we upgrade image and accidently forgot to change this into existing dir? Are we sure `/etc/tls/private` always exists?

bwplotka (changes requested):
Thanks! LGTM to all changes mentioned in PR, but I found some "sneaky" changes (committed by accident or not mentioned?) I have questions about 🙈

cmd/operator/main.go
TheSpiritXIII: It doesn't fully implement go/gmp:cert-args. Versus @bernot-dev's solution this doesn't let you read a pre-existing secret (like the one we'd load from GKE). This is the change mentioned in the PR description about using an emptyDir. Due to the `readOnlyRootFilesystem` flag, we cannot write a self-signed certificate in the root directory anymore, so I settled on mounting the emptyDir to `/etc/tls/private`. As it's mounted this always exists.

bernot-dev (approved):
LGTM

bwplotka (approved):
Sweet, thanks!

PR #946, opened 2024-04-25
feat: Made config-reloader resilient to initial readiness server unavailability (5s default timeout).

bwplotka:
This should fix misleading restart of config-reloader which happeend in the past, yet giving us some extra guarantee when readiness url is misconfigured.

Alternative is to again - [remove startup check logic](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/526) and ensure our tests guards against misconfiguration of url in our manifests (or metric/alert on configuration not reloading, but's harder). Since the team was not comfortable with removal of readiness check, let's try something that will help in common case or config-reloader starting faster.

### DIFF
--- a/cmd/config-reloader/main.go
+++ b/cmd/config-reloader/main.go
@@ -45,8 +45,11 @@ func main() {
 		// management APIs, e.g.
 		// https://prometheus.io/docs/prometheus/latest/management_api/
 		// https://prometheus.io/docs/alerting/latest/management_api/
-		reloadURLStr  = flag.String("reload-url", "http://127.0.0.1:19090/-/reload", "reload endpoint triggers a reload of the configuration file")
-		readyURLStr   = flag.String("ready-url", "http://127.0.0.1:19090/-/ready", "ready endpoint returns a 200 when ready to serve traffic")
+		reloadURLStr                      = flag.String("reload-url", "http://127.0.0.1:19090/-/reload", "reload endpoint of the configuration target that triggers a reload of the configuration file")
+		readyURLStr                       = flag.String("ready-url", "http://127.0.0.1:19090/-/ready", "ready endpoint of the configuration target that returns a 200 when ready to serve traffic. If set, the config-reloader will probe it on startup")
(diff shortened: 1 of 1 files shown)


TheSpiritXIII (approved):
This might even help our config-reloader SLO. Looks great. Thanks!

cmd/config-reloader/main.go
```diff
@@ -45,8 +45,11 @@ func main() {
 		// management APIs, e.g.
 		// https://prometheus.io/docs/prometheus/latest/management_api/
 		// https://prometheus.io/docs/alerting/latest/management_api/
-		reloadURLStr  = flag.String("reload-url", "http://127.0.0.1:19090/-/reload", "reload endpoint triggers a reload of the configuration file")
-		readyURLStr   = flag.String("ready-url", "http://127.0.0.1:19090/-/ready", "ready endpoint returns a 200 when ready to serve traffic")
+		reloadURLStr                      = flag.String("reload-url", "http://127.0.0.1:19090/-/reload", "reload endpoint of the configuration target that triggers a reload of the configuration file")
+		readyURLStr                       = flag.String("ready-url", "http://127.0.0.1:19090/-/ready", "ready endpoint of the configuration target that returns a 200 when ready to serve traffic")
+		readyProbingInterval              = flag.Duration("ready-probing-interval", 1*time.Second, "how often to poll ready URL during start-up")
```
pintohutch, line 50: suggestion (nit, non-blocking): rename flags to "startup-probing-interval" and "startup-probing-no-connection-threshold"?

That reads more clearly to me that these parameters only control startup behavior, rather than continually probing during the lifetime of the process.

pintohutch (approved):
% nit comment

maxamins (approved):
NIce!

cmd/config-reloader/main.go
bwplotka: addressed (moved to `ready-startup-probing-interval` and `ready-startup-probing-no-conn-threshold`)

PR #949, opened 2024-04-25
refactor: move monitoring and rules defaulters and validators to types

TheSpiritXIII:
## Changes

1. Merge rules package into rules type, and move rules validators to the rules type.

In my opinion, this is where rules generation should live. In my ideal world, types would be a separate package than the operator, and external tools could import the types package and validate them. Two examples of where this would be useful is a prometheus-operator converter (which would even be able to compare resulting Prometheus configurations this way) and a dedicated webhook package.

This is also consistent with the monitoring types. `RuleGroupsConfig` is analogous to `ScrapeConfigs`.

2. Expand rules generation tests to validate that forbidden labels return an error.

3. Move monitoring defaulters to inline in the types.

4. Moved around monitoring types methods to be local to each class (e.g. PodMonitoring methods next to other PodMonitoring methods, instead of interleaved methods all over).

5. The monitoring and rules validate methods no longer take dummy projects, names and locations. This makes the code less confusing.


bwplotka (approved):
Nice, perhaps we could split rule_type to _types.go _validation.go or sth as the file is bit long-ish (not a blocker). I like we removed that rules package that was used only in one place and unnecessary exported. Consistency with scraping makes sense.

Thanks!

bwplotka (approved):
Thanks! LGTM

PR #950, opened 2024-04-25
refactor: don't use and generate CRD clientsets

TheSpiritXIII:
Some motivators:
1. The potential need to split webhooks into its own package: the webhook package should not depend on the entire operator. It would be great if we could split types into its own package. Removing this code creates a clear declination of what should go into the types package.
2. Our clientset code generator is deprecated and slow.
3. The controller-runtime client is easier to use on e2e tests.


pintohutch (approved):
That's a handsome diff! Cool with me - as if there are any issues with g3 imports, we have the clientset version in source-control. 🚢

PR #953, opened 2024-04-29
refactor(helm): remove updateversion and centralized one value file.

bwplotka:
### Changes
* Removed updateversion Make cmd
* Only use helm template once in a consistenst way.
* Added datasyncer (minimal templating)
* Added one values.global.yaml that contains all variables we care to change e.g. every release.
  * Moved all uninteresting settings to local values.yaml if applicable, wanted to make sure global file is changed on release only.

### Rationales
* Avoid confusion. Ideally there is a single source of truth for images, resources etc for our OSS manifests. Right now it will be charts/values.global.yaml
* Avoid make updateversion unsafe templating. Use helm so we are a bit safer on templating.
  * It's easy to make mistake and sed something wrongly. For example the app Version module
that perhaps suppose to change mainModuleVersion variable (I suppose?) `@$(SED) -i -r 's#	Version    = .*#	Version    = "$(LABEL_API_VERSION)"#g' pkg/export/export.go` was not matching anything.
  * We use Go template var where we expect things to be templated.

### Open Questions
* Do we need such complex helpers? I don't get why we template a lot? Do we do something special on release/internally to explain this?
* We still don't template mainModuleVersion file, but we should likely do that differntly (e.g build tag) and updateversion was not working already.


helm/datasource-syncer.yaml
maxamins, line 1: why is there two datasouce-syncer.yaml? Do we need this one?

bwplotka: Removed.

manifests/opconfig.yaml
```diff
@@ -0,0 +1,38 @@
+# Copyright 2024 Google LLC
+#
```
TheSpiritXIII, line 2: nit: where did this file come from? I don't see us generating it in `presubmit.sh`.

TheSpiritXIII (approved):
Looks good thanks! Just leaving some minor nits.

We're banking on release-please to automate the `mainModuleVersion` eventually so this is great! Perhaps @bernot-dev would have better Helm-related feedback since I'm by far not an expert.

charts/README.md
```diff
@@ -0,0 +1,24 @@
...
+Each chart requires its local `values.yaml` as we as `values.global.yaml` in this directory.
+The global value file allows us to have a single source of truth file to variables
+we use for our example manifests for `/manifests/operator.yaml`, standalone
+`/manifests/rule-evaluator.yaml` and `/cmd/datasource-syncer/datasource-syncer.yaml`.
+
+To render them, use `helm template` command with multiple `-f` flags (scripted in `make regen` command):
+
+```bash
+  ${HELM} template "${REPO_ROOT}/charts/rule-evaluator" \
+   -f "${REPO_ROOT}/charts/rule-evaluator/values.yaml" \
```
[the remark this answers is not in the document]
bwplotka: Correct. The global values are now in custom place (outside of charts)

manifests/opconfig.yaml
bwplotka: Ah.. copy pasta again, sorry.

charts/README.md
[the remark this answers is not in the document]
TheSpiritXIII: Makes sense. Thanks!

charts/values.global.yaml
```diff
@@ -0,0 +1,79 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+commonLabels: false
+namespace:
+  public: gmp-public
+  system: gmp-system
+version: 0.12.0
```
bernot-dev, line 19: Why is "version" at the top level of the new global values file? Are we saying we're going to synchronize this version across the operator, datasource-syncer and rule-evaluator?

Makefile
```diff
@@ -138,7 +138,7 @@ bin: $(GOCMDS) $(GOINSTR)
 regen:       ## Refresh autogenerated files and reformat code.
              ## Use CHECK=1 to only validate clean repo after run.
              ##
-regen: $(ADDLICENSE)
+regen: $(ADDLICENSE) $(HELM)
```
bernot-dev, line 141: Helm isn't used in this target? At least not directly...and this won't install it in the Docker container.

hack/presubmit.sh
```diff
@@ -84,8 +84,7 @@ update_crdgen() {
   which controller-gen || go install github.com/TheSpiritXIII/controller-tools/cmd/controller-gen@v0.14.1-gmp
```
bernot-dev, line 84: Thought: Irrelevant to this PR, butt this should probably be moved to Bingo.

charts/datasource-syncer/templates/datasource-syncer.yaml
```diff
@@ -0,0 +1,59 @@
...
+        args:
+        - "--datasource-uids=$DATASOURCE_UIDS"
+        - "--grafana-api-endpoint=$GRAFANA_API_ENDPOINT"
+        - "--grafana-api-token=$GRAFANA_API_TOKEN"
+        - "--project-id=$PROJECT_ID"
+      restartPolicy: Never
+---
+
+apiVersion: batch/v1
+kind: CronJob
```
bernot-dev, line 39: By convention, each resource should be in its own file and named according to the kind of resource it creates.

https://helm.sh/docs/chart_best_practices/templates/#structure-of-templates

bernot-dev:
Completely agree that we now have better options than `sed` to update versions more safely.

Overall, I like this change. It's not clear to me that putting all of the images in a global file is the best solution, because it has to be manually included, and not all of the charts use all of those images. But it's probably fine because we're reducing the number of calls to `helm template`. And centralizing versions into one file has advantages, as well.

Since the versions are not included in the default values file now, I would consider making them [required](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) to make it noisier if the version is not supplied.

Makefile
bwplotka: It is used, but you are right, it's done via importing env bingo source, thanks! Will kill.

hack/presubmit.sh
bwplotka: 👍🏽  let's do in the next one

charts/datasource-syncer/templates/datasource-syncer.yaml
bwplotka: Happy to split

charts/values.global.yaml
bwplotka: This comment is connected to another statement, so let's discuss both here

> Since the versions are not included in the default values file now, I would consider making them [required](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) to make it noisier if the version is not supplied.

We used to have `appVersion` in the chart yaml for each operator/pe and rule-eval. For this purpose we had `make updateversion` to update appVersion for  operator `@$(YQ) -i '.appVersion = strenv(LABEL_API_VERSION)' ./charts/operator/Chart.yaml`. I think we meant to add same for rule-eval but forgot (?). This field was in rule-eval too:

image

I think it only changes labels like `app.kubernetes.io/version` (with and without common), plus part of common label `app.kubernetes.io/instance`.

>  Are we saying we're going to synchronize this version across the operator, datasource-syncer and rule-evaluator?

Given what I saw I assumed we wanted that. We didn't want to synchronize those `app.kubernetes.io/version` for all manifests?

I assumed we want to template this (given sedding), thus moved this to version field in values. Datasource-syncer might need to add that label, but didn't want to change output in refactor PR yet.

bwplotka: Discussed offline, sounds like we can iterate and we can keep what I have in this PR for now.

bwplotka:
Addressed comments and simplified to one `values.global.yaml` -- simpler and consistent.

PR #956, opened 2024-05-01
fix: configurable vpa

bernot-dev:
Only load the scaling controller when VPA is enabled in the cluster.


TheSpiritXIII (approved):
Thank you!

Issue #713, opened 2023-12-13
Add tests for caBundle injection code


bwplotka:
### AC

* We have tests ensuring we update caBundle on self-deploy
* We have tests ensuring we don't try updating caBundle if those are updated by GKE on managed GMP

PR #959, opened 2024-05-02
refactor: consolidate webhook code and add CA bundle unit test

TheSpiritXIII:
Fixes https://github.com/GoogleCloudPlatform/prometheus-engine/issues/713

I've consolidated all webhook-related code into its own file. Furthermore, the `TestWebhookCABundleUpdate` unit test ensures that the webhook CA is updated correctly. Since the CA update loop sleeps for 1 minute, I made the tests run in parallel.


bernot-dev:
I'm going to close #663 because there will be too many merge conflicts for that PR to make sense.

pkg/operator/webhook.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(time.Minute):
+		}
+	}
+}
+
+type operatorConfigValidator struct {
```
bernot-dev, line 211: Why is this validator being moved?

bernot-dev (approved):
This change is fine, but I personally wouldn't invest a lot of energy in this area right now because most of the logic to handle certs will change soon.

PR #961, opened 2024-05-06
fix: helm chart to run datasource-syncer to run on arm64 nodes

maxamins:
Change https://github.com/GoogleCloudPlatform/prometheus-engine/pull/960 was missing the helm change.


pintohutch (approved):
Whoop - good catch!

PR #972, opened 2024-05-14
chore: add requests and limits to initContainer

pintohutch:
Avoids issues with Gatekeeper policies rejecting workloads that do not specify resource requests.

Note - the scheduler should not be affected, as the maximum of initContainers and containers is [considered](https://github.com/kubernetes/enhancements/tree/6394dde76f9f6993293ee8c944da4550c6fdf9bd/keps/sig-node/753-sidecar-containers#exposing-pod-resource-requirements) the effective request amount.

### DIFF
--- a/charts/values.global.yaml
+++ b/charts/values.global.yaml
@@ -49,11 +49,11 @@ resources:
       memory: 16M
   bash:
     limits:
-      memory: 32M
+      memory: 6M
     requests:
       cpu: 1m
-      memory: 4M
-  collector:
+      memory: 6M
+  prometheus:
     limits:
       memory: 2G
     requests:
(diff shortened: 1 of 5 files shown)


bernot-dev (approved):
LGTM

charts/values.global.yaml
```diff
@@ -49,11 +49,11 @@ resources:
       memory: 16M
   bash:
     limits:
-      memory: 32M
+      memory: 1M
```
bwplotka, line 52: I hope `touch` CLI does not use more than 1M.. but it's likely we see occasional OOM here, no? Is it worth a risk?

bwplotka (approved):
Nice, thanks! Do you mind rebasing your second commit to one for conform? CI?

pintohutch:
Yea - I may hold off on merging. For some reason E2E tests are timing out with this change...

PR #974, opened 2024-05-16
feat: allow updating export options at runtime

TheSpiritXIII:
Pre-req: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001

1. `ApplyConfig` now takes `ExporterOpts`, which allows runtime updates.
2. `exporter.New` now takes a context and `Run` no longer does, which means the `Exporter` uses the same context throughout its lifetime.

These changes allow us to update the metric client at runtime. When a metric client is updated, it may take:
- 50ms worse case, when not enough samples are being exported to fill a batch.
- <50ms best case, when many samples (>200) are being written


pkg/export/export.go
```diff
@@ -262,7 +310,7 @@ func (alwaysLease) OnLeaderChange(func()) {
 	// We never lose the lease as it's always owned.
 }
 
-func newMetricClient(ctx context.Context, opts ExporterOpts) (*monitoring.MetricClient, error) {
+func defaultNewMetricClient(ctx context.Context, opts ExporterOpts) (metricServiceClient, error) {
```
pintohutch, line 313: question: Do we need to support _every_ command line argument for hot-reloading? IIUC we only need to support options configurable in `OperatorConfig.Collection`, which would be limited to:
- export.match
- export.credentials

after #965 is merged.

I was hoping we could avoid recreating the metric client, but I don't think we can if we want to support resetting `export.credentials`..

pintohutch (commented):
Nice!

Partial review - will finish up later this week.

pkg/export/setup/setup.go
```diff
@@ -82,138 +83,147 @@ func SetGlobal(exporter *export.Exporter) (err error) {
...
-// Potential risky logic can be moved to the returned function we return here.
-// See b/344740239 on how hard is to debug regressions here.
-func FromFlags(a *kingpin.Application, userAgentProduct string) func(context.Context, log.Logger, prometheus.Registerer) (*export.Exporter, error) {
-	var (
-		metadataFetchTimeout time.Duration
-		opts                 export.ExporterOpts
-	)
-	opts.UserAgentProduct = userAgentProduct
-
+func ExporterOptsFlags(a *kingpin.Application, opts *export.ExporterOpts) {
```
pintohutch, line 99: suggestion: It feels brittle to rely on `opts` being `.Default()`ed before this is called every time.

What if we did the defaulting of `opts` in this routine, like before? Tying the flag/config definitions and default behavior centrally makes a lot of sense to me.

I think this also comes back to my previous comment of around how we envision clients handling the `ExporterOpts` struct.

pkg/export/setup/setup.go
```diff
@@ -82,138 +83,147 @@ func SetGlobal(exporter *export.Exporter) (err error) {
...
-			// use if labels or external label settings does not have those set. Those will
-			// be used as crucial labels for export to work against GCM's Prometheus target.
-			//
-			// Set a hard time limit due to readiness and liveliness probes during this stage.
-			mctx, cancel := context.WithTimeout(context.Background(), metadataFetchTimeout)
-			tryPopulateUnspecifiedFromMetadata(mctx, logger, &opts)
-			cancel()
-			_ = level.Debug(logger).Log("msg", "best-effort on-GCE metadata gathering finished")
-		}
+type HAOptions struct {
```
pintohutch, line 172: question: Why separate the haBackend config?

And either way, if want to do that, it would be much easier to review in a dedicated PR with a justification provided as there's a lot going on here already :)

pkg/export/setup/setup.go
```diff
@@ -239,6 +249,43 @@ func ExtraArgs() ([]string, error) {
 	return shlex.Split(os.Getenv(ExtraArgsEnvvar))
 }
 
+type MetadataOpts struct {
```
pintohutch, line 252: question: Why separate the MetadataOpts config?

And either way, if want to do that, it would be much easier to review in a dedicated PR with a justification provided as there's a lot going on here already :)

pkg/export/setup/setup_test.go
```diff
@@ -35,19 +35,34 @@ func TestFromFlags_NotOnGCE(t *testing.T) {
...
 	// Fake app invocation.
 	if _, err := fake.Parse(nil); err != nil {
 		t.Fatal(err)
 	}
 
-	// Make sure constructor does is not stuck forever.
-	_, _ = newExport(ctx, log.NewLogfmtLogger(os.Stderr), nil)
+	done := make(chan struct{}, 1)
+	go func() {
+		metadataOpts.ExtractMetadata(logger, &exporterOpts)
```
pintohutch, line 56: question: If we do want to make this change, wouldn't we also need to update the tests in `TestTryPopulateUnspecifiedFromMetadata` to ensure timeouts are correctly updated and respected in the config?

cmd/rule-evaluator/main.go
```diff
@@ -96,155 +96,110 @@ func main() {
...
 
-	configFile := a.Flag("config.file", "Prometheus configuration file path.").
-		Default("prometheus.yml").String()
+	haOpts := exportsetup.HAOptions{}
+	haOpts.Default()
+	haOpts.SetupFlags(a)
 
-	a.Flag("alertmanager.notification-queue-capacity", "The capacity of the queue for pending Alertmanager notifications.").
-		Default("10000").IntVar(&notifierOptions.QueueCapacity)
+	evaluatorOpts := evaluatorOptions{
```
pintohutch, line 117: suggestion: Again, it seems a lot of the code from #1001 was folded in here, which makes this change hard to evaluate on its own merit.

Can we back this out, or perhaps get the other one merged first before proceeding with this review?

pintohutch (changes requested):
Apologies for the delay, finally finished review.

Looks like this is a combination change that depends on #1001. Is there any way this can be separated out? It's difficult to review the combined changes at once.

TheSpiritXIII:
> Is there any way this can be separated out?

It already is -- go to the "Files changed" tab and on the top right directly underneath the tabs, there's a drop-down that says "All commits" where you can select which commit to review. Or do you mean you want this change without being rebased on top of the other change?

pintohutch:
Ah ok - it wasn't clear to me that I should just review that [commit](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974/commits/c09542ecaeac2e65937dc29254a5ef8f3383db4f).

If that's the case, I think most of my review comments outside of the rule-evaluator ones still apply - thanks for clarifying :)

cmd/rule-evaluator/main.go
TheSpiritXIII: Addressing in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001

pkg/export/setup/setup.go
TheSpiritXIII: Addressing in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001

pkg/export/setup/setup.go
TheSpiritXIII: Addressing in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001

pkg/export/setup/setup.go
TheSpiritXIII: Addressing in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001

pkg/export/setup/setup_test.go
TheSpiritXIII: Addressing in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001

pkg/export/export.go
```diff
@@ -279,11 +327,19 @@ func newMetricClient(ctx context.Context, opts ExporterOpts) (*monitoring.Metric
...
+	// Disable auth when the exporter is disabled because we don't want a panic when default
+	// credentials are not found.
+	if opts.DisableAuth || opts.Disable {
 		clientOpts = append(clientOpts,
 			option.WithoutAuthentication(),
 			option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())),
 		)
+	} else if opts.CredentialsFile == "" && len(opts.CredentialsFromJSON) == 0 {
+		// If no credentials are found, gRPC panics so we check manually.
+		_, err := google.FindDefaultCredentials(ctx)
```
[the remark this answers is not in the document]
TheSpiritXIII: The gRPC client looks for them automatically and panics if it doesn't find them so we don't need to pass it in. I'll look for the exact place this happens...

pkg/export/export.go
```diff
@@ -647,26 +710,37 @@ func (e *Exporter) Run(ctx context.Context) error {
 	}
 	defer stopTimer()
 
-	curBatch := newBatch(e.logger, e.opts.Efficiency.ShardCount, e.opts.Efficiency.BatchSize)
+	e.mtx.RLock()
+	opts := e.opts
+	e.mtx.RUnlock()
+
+	curBatch := newBatch(e.logger, opts.Efficiency.ShardCount, opts.Efficiency.BatchSize)
```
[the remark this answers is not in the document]
pintohutch: Nevermind - makes sense.

pkg/export/export.go
```diff
@@ -647,26 +710,37 @@ func (e *Exporter) Run(ctx context.Context) error {
...
 
 	// Send the currently accumulated batch to GCM asynchronously.
 	send := func() {
+		e.mtx.RLock()
+		opts := e.opts
+		sendFunc := e.metricClient.CreateTimeSeries
```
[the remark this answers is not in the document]
pintohutch: Makes sense - thanks!

pkg/export/storage.go
```diff
@@ -55,12 +55,12 @@ func NewStorage(exporter *Exporter) *Storage {
 
 // ApplyConfig applies the new configuration to the storage.
 func (s *Storage) ApplyConfig(cfg *config.Config) error {
-	return s.exporter.ApplyConfig(cfg)
+	return s.exporter.ApplyConfig(cfg, nil)
```
pintohutch, line 58: question: Doesn't the rule evaluator need to refresh export options at runtime as well?

pkg/export/export.go
```diff
@@ -379,20 +402,24 @@ func New(ctx context.Context, logger log.Logger, reg prometheus.Registerer, opts
 		lease = NopLease()
```
pintohutch, line 402: question: Why do we need the `Default()` call here?

Isn't all of this being done as part of the flags setup already?

In general, if clients of the struct are expected to do some operations on it before using it, I think it makes sense to wrap those operations in a constructor, i.e. `newExporterOpts`.

pkg/export/export.go
```diff
@@ -391,20 +433,38 @@ const (
...
 // Must be called at least once before Export() can be used.
-func (e *Exporter) ApplyConfig(cfg *config.Config) (err error) {
+func (e *Exporter) ApplyConfig(cfg *config.Config, opts *ExporterOpts) (err error) {
+	if e.seriesCache == nil {
+		return errors.New("nop exporter")
+	}
+	optsChanged := false
+	e.mtx.RLock()
+	if opts != nil {
+		opts.Default()
```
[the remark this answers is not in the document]
pintohutch: Actually - why is the `Default()` call necessary here? What are clients of `ApplyConfig` expected to be doing with the opts beforehand?

pkg/export/export.go
```diff
@@ -416,63 +443,96 @@ const (
...
+	if e.isNopExporter() {
+		return errors.New("nop exporter")
 	}
-	if !cfg.GlobalConfig.ExternalLabels.Has(KeyCluster) {
-		builder.Set(KeyCluster, e.opts.Cluster)
+
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+
+	optsChanged := false
```
pintohutch, line 454: suggestion: Don't check for changed status.

Why not just always overwrite the opts?

pkg/export/export.go
```diff
@@ -416,63 +443,96 @@ const (
...
+		}
+		optsChanged = !reflect.DeepEqual(e.opts, opts)
+		if optsChanged {
+			e.opts = *opts
+		}
 	}
-	lset := builder.Labels()
+
+	lset := createLabelSet(cfg, &e.opts)
+	labelsChanged := !labels.Equal(e.externalLabels, lset)
```
pintohutch, line 467: suggestion: Don't check for changed labels

We weren't doing it before. What's the motivation to add it now?

pkg/export/export.go
TheSpiritXIII: I don't think we should be recreating the metric client each time. If the metric client is recreated, it has to potentially redo the TCP handshake. With HTTP/2, TCP connections are kept alive for a small amount of time to reduce load when multiple requests are made to the same server in succession. In our case, we might send a CMP call every 50ms at the worst case, which is highly likely to benefit from the persistent TPC connection.

pkg/export/export.go
```diff
@@ -416,63 +443,96 @@ const (
...
-	// environment-specific value on GCE/GKE, is used.
-	builder := labels.NewBuilder(cfg.GlobalConfig.ExternalLabels)
-
-	if !cfg.GlobalConfig.ExternalLabels.Has(KeyProjectID) {
-		builder.Set(KeyProjectID, e.opts.ProjectID)
-	}
-	if !cfg.GlobalConfig.ExternalLabels.Has(KeyLocation) {
-		builder.Set(KeyLocation, e.opts.Location)
+func (e *Exporter) ApplyConfig(cfg *config.Config, opts *ExporterOpts) (err error) {
+	if e.isNopExporter() {
```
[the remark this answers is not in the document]
TheSpiritXIII: Only main.go calls this,  which does not use the no-op exporter. I added a comment here and other places I early-exit.

TheSpiritXIII (commented):
@pintohutch thanks! I made some changes you requested. Please take another look!

pkg/export/export_test.go
```diff
@@ -332,64 +330,179 @@ type testMetricService struct {
...
+	}
+	sendAndTestSamples("project-abc")
+
+	if err := e.ApplyConfig(&config.Config{}, &ExporterOpts{ProjectID: "project-xyz"}); err != nil {
+		t.Fatalf("Initial apply: %s", err)
+	}
+	sendAndTestSamples("project-xyz")
+}
+
+func TestDisabledExporter(t *testing.T) {
```
pintohutch, line 483: question: What's being tested in this function?

pintohutch (commented):
Awesome - I think it mostly LGTM, though with a change of this scope I'd love some co-review love from @bwplotka if possible.

A few comments and questions. Once those are resolved, I think it's good to :shipit:

pkg/export/export_test.go
TheSpiritXIII: Added more to the test to make it clearer.

pkg/export/export.go
TheSpiritXIII: Done.

pintohutch (approved):
LGTM!

pkg/export/export.go
```diff
@@ -416,63 +442,103 @@ const (
...
```
bwplotka, line 446: Do we really need testing code be in prod code though? It's just assertion for test generally?

pkg/export/setup/setup.go
```diff
@@ -82,9 +83,15 @@ func SetGlobal(exporter *export.Exporter) (err error) {
 // Global returns the global instance of the GCM exporter.
 func Global() *export.Exporter {
 	if globalExporter == nil {
-		// This should usually be a panic but we set an inactive default exporter in this case
-		// to not break existing tests in Prometheus.
+		if !testing.Testing() {
```
bwplotka, line 86: Ideally we get rid of all use of testing in prod code, this is getting worse and worse. Hinders readability and complexity for testability. Wonder if there are alternatives. How we can test prod code if it does things differently for tests?

pkg/export/export.go
```diff
@@ -657,10 +716,15 @@ func Version() (string, error) {
 // The per-shard overhead is minimal and thus a high number can be picked, which allows us
 // to cover a large range of potential throughput and latency combinations without requiring
 // user configuration or, even worse, runtime changes to the shard number.
-func (e *Exporter) Run(ctx context.Context) error {
-	defer e.metricClient.Close()
-	go e.seriesCache.run(ctx)
-	go e.lease.Run(ctx)
+func (e *Exporter) Run() error {
+	if e.isNopExporter() {
```
bwplotka, line 720: to kill IMO

pkg/export/export.go
```diff
@@ -416,63 +442,103 @@ const (
...
+	defer e.mtx.Unlock()
+
+	// Don't recreate the metric client each time. If the metric client is recreated, it has to
+	// potentially redo the TCP handshake. With HTTP/2, TCP connections are kept alive for a small
+	// amount of time to reduce load when multiple requests are made to the same server in
+	// succession. In our case, we might send a CMP call every 50ms at the worst case, which is
+	// highly likely to benefit from the persistent TPC connection.
+	optsChanged := false
+	if opts != nil {
+		opts.Default()
```
bwplotka, line 461: Should we ensure caller does that? Otherwise it's weird we modify this option on every apply. What if caller assume it's ready only (e.g. do something on unset event?)

plus can we call it `DefaultUnsetFields`?

bwplotka (commented):
Nice! Generally LGTM, it's nice as it enables collector improvement of no restart for exporter config / label setting. 

I would focus on killing testing code in prod code though, it's getting bit out of control, but I will understand if @TheSpiritXIII don't have time for that here. Plus I would NOT modify options (or config) in apply method, feels it can cause surprised going forward.

pkg/export/export.go
TheSpiritXIII: Good suggestion. Done. Thanks!

pkg/export/setup/setup.go
TheSpiritXIII: I agree, but I actually tried to get rid of this and it makes Prometheus unit tests extremely slow by an order of 10x-100x slower. To do this the proper way, it would be far too much code to set the global exporter to a NopExporter in all unit tests.

If you want to use the real exporter in unit tests, simply call `SetGlobal`. This is what `main.go` does in Prometheus.

pkg/export/export.go
TheSpiritXIII: Thanks, I got rid of this and added a comment instead.

pkg/export/export.go
TheSpiritXIII: Thanks, I got rid of this and added a comment instead.

pkg/export/export.go
```diff
@@ -316,11 +331,19 @@ func newMetricClient(ctx context.Context, opts ExporterOpts) (*monitoring.Metric
...
+	// credentials are not found.
+	if opts.DisableAuth || opts.Disable {
 		clientOpts = append(clientOpts,
 			option.WithoutAuthentication(),
 			option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())),
 		)
+	} else if opts.CredentialsFile == "" && len(opts.CredentialsFromJSON) == 0 {
+		// If no credentials are found, gRPC panics so we check manually.
+		_, err := google.FindDefaultCredentials(ctx)
+		if err != nil {
```
bwplotka, line 344:
```suggestion
		if _, err := google.FindDefaultCredentials(ctx); err != nil {
```

pkg/export/setup/setup.go
bwplotka: 👍🏽 We should be able to remove this part soonish with PRW too.

pkg/export/gcm/promtest/local_export.go
```diff
@@ -74,7 +74,9 @@ func (l *localExportWithGCM) Ref() string { return "export-pkg-with-gcm" }
 func (l *localExportWithGCM) start(t testing.TB, _ e2e.Environment) (v1.API, map[string]string) {
 	t.Helper()
 
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(signals.SetupSignalHandler())
```
bwplotka, line 77: woooo didn't know this, so good

bwplotka (approved):
Thanks for great work on this - it wasn't easy. What I like about this, is that it actually is more consistent with Prometheus OSS, PRW handling is hot-configurable 👍🏽 

Thanks for addressing all tricky comments and reducing testing code in prod code 💪🏽 

Great effort on tests too!

pkg/export/export.go
bwplotka: Small nit so merging

PR #976, opened 2024-05-16
test(e2e): debug pod logs

TheSpiritXIII:
Show debug Prometheus collector logs when an e2e test fails to startup.

I was running into an issue that only happened on GitHub Actions. How this works:
1. Use a context that times out right before the test. This ensures the test catches the timeout and can "cleanup" or in this case, show logs before exiting.
2. Show debug logs in the test. We use a new context in case the test timed out, and the original context cannot be reused.


bwplotka (approved):
Thanks! Let's try 💪🏽

PR #977, opened 2024-05-16
feat(export): allow GMP collectors to startup without metadata

TheSpiritXIII:
The way our e2e tests are written is usually we check for the GMP collector to start-up correctly.

The problem is that if we defer the metadata labeling to the configuration as opposed to flags, then the collector will never start-up correctly (and also goes against our goal which is to prevent the collector from restarting at initialization).

The simple solution here is to defer error handling until we have a configuration with scape or rule configurations.

See also: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/965

### DIFF
--- a/.conform.yaml
+++ b/.conform.yaml
@@ -42,6 +42,7 @@ policies:
           - configs
           - deps
           - e2e
+          - export
           - main
           - operator
           - prometheus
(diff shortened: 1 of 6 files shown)


pintohutch (approved):
LGTM % linting failures

pintohutch:
It also looks like there's some build and test failures as well - maybe a rebase from main will fix things?

PR #981, opened 2024-05-20
feat: populate rules status

TheSpiritXIII:
Add MonitoringStatus to all Rules CRDs.

### DIFF
--- a/charts/operator/crds/monitoring.googleapis.com_clusterrules.yaml
+++ b/charts/operator/crds/monitoring.googleapis.com_clusterrules.yaml
@@ -124,6 +124,44 @@ spec:
             type: object
           status:
             description: Most recently observed status of the resource.
+            properties:
+              conditions:
+                description: Represents the latest available observations of a podmonitor's
+                  current state.
+                items:
+                  description: MonitoringCondition describes the condition of a PodMonitoring.
+                  properties:
+                    lastTransitionTime:
+                      description: Last time the condition transitioned from one status
+                        to another.
+                      format: date-time
+                      type: string
+                    lastUpdateTime:
+                      description: The last time this condition was updated.
+                      format: date-time
+                      type: string
+                    message:
+                      description: A human-readable message indicating details about
+                        the transition.
+                      type: string
+                    reason:
+                      description: The reason for the condition's last transition.
+                      type: string
+                    status:
+                      description: Status of the condition, one of True, False, Unknown.
+                      type: string
+                    type:
+                      description: MonitoringConditionType is the type of MonitoringCondition.
+                      type: string
+                  required:
+                  - status
+                  - type
+                  type: object
+                type: array
+              observedGeneration:
+                description: The generation observed by the controller.
+                format: int64
+                type: integer
             type: object
         required:
         - spec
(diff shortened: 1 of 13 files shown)


pkg/operator/apis/monitoring/v1/rules_types.go
```diff
@@ -192,5 +204,5 @@ type Rule struct {
 
 // RulesStatus contains status information for a Rules resource.
 type RulesStatus struct {
-	// TODO: add status information.
+	MonitoringStatus `json:",inline"`
```
maxamins, line 207: I don't see in the CRD. Do you need to run `make regen`?

charts/operator/crds/monitoring.googleapis.com_clusterrules.yaml
```diff
@@ -124,6 +124,44 @@ spec:
             type: object
           status:
             description: Most recently observed status of the resource.
+            properties:
```
maxamins, line 127: What caused this piece to be generated? I don't see a matching change.

pkg/operator/apis/monitoring/v1/monitoring_types.go
```diff
@@ -77,17 +81,17 @@ type MonitoringStatus struct {
...
 	var (
 		specChanged              = status.ObservedGeneration != gen
 		statusTransition, update bool
 		conds                    = make(map[MonitoringConditionType]*MonitoringCondition)
 	)
 
-	if cond.Type == "" || cond.Status == "" {
-		return update, errInvalidCond
+	if !cond.IsValid() {
+		return false
```
maxamins, line 94: How often does this happen? Should we log when this happens?

maxamins (commented):
Nice job! few comments.

charts/operator/crds/monitoring.googleapis.com_clusterrules.yaml
TheSpiritXIII: This is the addition of the `MonitoringCondition` field.

pkg/operator/apis/monitoring/v1/monitoring_types.go
TheSpiritXIII: In practice, never. We always call `SetMonitoringCondition` with a valid condition.

If someone wants to build a `MonitoringCondition` dynamically and wants logging, they could call `IsValid` and then log.

pkg/operator/apis/monitoring/v1/rules_types.go
TheSpiritXIII: It's in there (the `properties` field). Correct me if I'm wrong.

charts/operator/crds/monitoring.googleapis.com_clusterrules.yaml
maxamins: `MonitoringCondition` seems to have been added in a prior release. So it's strange that this is showing up now.

pkg/operator/apis/monitoring/v1/rules_types.go
maxamins: I see it. Thanks!

pkg/operator/apis/monitoring/v1/monitoring_types.go
maxamins: OK!

maxamins (approved):

charts/operator/crds/monitoring.googleapis.com_clusterrules.yaml
TheSpiritXIII: The definition is copied into each place it's used. This is the first time it's used in `ClusterRules`!

PR #985, opened 2024-05-24
feat: default CRDs on the reconcile loop

TheSpiritXIII:
If webhooks are not running for some unknown reason, we will want to default objects.

### DIFF
--- a/pkg/operator/collection.go
+++ b/pkg/operator/collection.go
@@ -19,6 +19,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"path"
@@ -126,9 +127,8 @@ func setupCollectionControllers(op *Operator) error {
 }
 
 type collectionReconciler struct {
-	client        client.Client
-	opts          Options
-	statusUpdates []monitoringv1.MonitoringCRD
+	client client.Client
+	opts   Options
 }
 
 func newCollectionReconciler(c client.Client, opts Options) *collectionReconciler {
@@ -181,15 +181,6 @@ func (r *collectionReconciler) Reconcile(ctx context.Context, req reconcile.Requ
 	if err := r.ensureCollectorConfig(ctx, &config.Collection, config.Features.Config.Compression, config.Exports); err != nil {
 		return reconcile.Result{}, fmt.Errorf("ensure collector config: %w", err)
 	}
-
-	// Reconcile any status updates.
-	for _, obj := range r.statusUpdates {
-		if err := patchMonitoringStatus(ctx, r.client, obj, obj.GetMonitoringStatus()); err != nil {
-			logger.Error(err, "update status", "obj", obj)
-		}
-	}
-	// Reset status updates for next reconcile loop.
-	r.statusUpdates = r.statusUpdates[:0]
 
 	return reconcile.Result{}, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/operator/collection.go
```diff
@@ -364,7 +374,9 @@ type prometheusConfig struct {
 	SecretConfigs []secrets.SecretConfig `yaml:"kubernetes_secrets,omitempty"`
 }
 
-func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *monitoringv1.CollectionSpec, exports []monitoringv1.ExportSpec) (*prometheusConfig, error) {
+// makeCollectorConfig returns the Prometheus configuration, the list of objects to update, the list
+// of objects whose status need to be updated and any list of errors.
+func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *monitoringv1.CollectionSpec, exports []monitoringv1.ExportSpec) (*prometheusConfig, []monitoringv1.MonitoringCRD, []monitoringv1.MonitoringCRD, error) {
```
bernot-dev, line 379: suggestion: This may be a good place for named result parameters, since there are two parameters in a row with the same type.

go/go-style/decisions#named-result-parameters

maxamins: You can add `statusUpdates` and the new status back to the `collectionReconciler` struct.

pkg/operator/collection.go
```diff
@@ -354,7 +345,26 @@ func (r *collectionReconciler) ensureCollectorConfig(ctx context.Context, spec *
...
+
+	// Reconcile any status updates.
+	var errs []error
+	for _, obj := range fullUpdates {
+		// Update does not patch status. Must patch status first because `Update` would mutate
+		// the `obj` parameter and reset it before we can patch it.
+		if err := patchMonitoringStatus(ctx, r.client, obj, obj.GetMonitoringStatus()); err != nil {
+			errs = append(errs, err)
+		}
+		if err := r.client.Update(ctx, obj); err != nil {
```
maxamins, line 357: Why do we need to update here? is it because we add `c.Spec.TargetLabels.Metadata`?

bwplotka: Yea, "full" seems to mean "include non status updates" e.g. when defaulting.

I wonder if we can put better name here for those semantics, proposed in other comment.

pkg/operator/collection.go
```diff
@@ -328,7 +319,7 @@ func setConfigMapData(cm *corev1.ConfigMap, c monitoringv1.CompressionType, key
 
 // ensureCollectorConfig generates the collector config and creates or updates it.
 func (r *collectionReconciler) ensureCollectorConfig(ctx context.Context, spec *monitoringv1.CollectionSpec, compression monitoringv1.CompressionType, exports []monitoringv1.ExportSpec) error {
-	cfg, err := r.makeCollectorConfig(ctx, spec, exports)
+	cfg, fullUpdates, statusUpdates, err := r.makeCollectorConfig(ctx, spec, exports)
```
bwplotka, line 322: Nice, after reading whole PR I understand this, but this line alone was confusing (what full means, that one contain all updates, but second only status updates for same obj?).

Given whole PR context I would consider (just an idea):

* separate type for statuses
* method for ScheduleUpdate(obj CRD, patchStatusOnly bool)
* Then in this function do Update() which will hide non-trivial set of Update + patch calls.

Another benefit of the above is that I feel we risk a bit of efficiency regressions with it. Even if not critical for majority of cases, we might need to make it more complex for scalability (e.g. do async, batch it, backoff etc)

bwplotka (approved):
Generally LGTM, modulo some suggestions around readability, thanks!

The amount of different extra APIs calls for potentially 1k+ Monitoring Resources worries me. We might want to run/create some easy framework for Operator scale tests (aka operatorbench) one day (: I think fine for now

pkg/operator/collection.go
TheSpiritXIII: I now return a `[]update` instead.

pkg/operator/collection.go
TheSpiritXIII: Check out the new `update` struct!

bwplotka (approved):
Thanks! 💪🏽

PR #989, opened 2024-05-28
fix(e2e): reduce e2e flakiness when a deployment pod restarts

TheSpiritXIII:
Sometimes e2e tests fail if a deployment pod fails to startup, and restarts, or the GMP operator causes the pod to restart and we query the pod too quick.

The quick and easy fix is to only filter active pods.


bwplotka (approved):
Thanks!!

PR #991, opened 2024-05-30
fix: Use default shard count in export

pintohutch:
This reverts an accidental change in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/514


bwplotka (approved):
Eh, so silly mistake, sorry for that.

PR #994, opened 2024-06-03
chore: update linter to go1.22

bernot-dev:
Key language changes: https://tip.golang.org/doc/go1.22#language

Ranging over integers is expressed more concisely, and variables declared by a for loop are created anew each loop.

Linters now enforce that the new form of each of these changes.


TheSpiritXIII (approved):

pkg/operator/collection.go
```diff
@@ -494,8 +494,6 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 			logger.Info("ClusterNodeMonitoring job %s was not applied because OperatorConfig.collector.kubeletScraping is enabled. kubeletScraping already includes the metrics in this job.", "name", cm.Name)
 			continue
 		}
-		// Reassign so we can safely get a pointer.
```
maxamins, line 497: should 444 and 403 also be removed?

maxamins: Actually did you mean to remove this line?

bernot-dev: Yes, I meant to remove those. 444 and 403 can also be removed, but I'm going to leave them in place for this PR since the linter didn't catch them.

maxamins: Probably because a different name is used rather than `pm := pm`. I removed them in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1000

PR #995, opened 2024-06-03
fix(operator): support secrets outside default ns for scrape auth

fdaguin:
Fixes the issue shared through https://github.com/GoogleCloudPlatform/prometheus-engine/pull/776#issuecomment-2087606662.

The current implementation only supports a PodMonitoring selecting a Secret from the `default` namespace.

The existing E2E test cases were using a Secret deployed inside the `default` namespace therefore not covering this "edge" case:

https://github.com/GoogleCloudPlatform/prometheus-engine/blob/f21c2ba4c0a9c3108209c1e59ec75d52e388235e/e2e/authorization_test.go#L96-L177

### DIFF
--- a/pkg/operator/apis/monitoring/v1/http_types.go
+++ b/pkg/operator/apis/monitoring/v1/http_types.go
@@ -101,11 +101,8 @@ func (s *SecretKeySelector) toPrometheusSecretRef(m PodMonitoringCRD, pool Prome
 	}
 	if m.IsNamespaceScoped() {
 		monitoringNamespace := m.GetNamespace()
-		if monitoringNamespace == "" {
-			monitoringNamespace = metav1.NamespaceDefault
-		}
-		if ns != monitoringNamespace {
-			return "", fmt.Errorf("must use namespace %q, got: %q", m.GetNamespace(), ns)
+		if monitoringNamespace != "" {
+			ns = monitoringNamespace
 		}
 	}
 
--- a/pkg/operator/apis/monitoring/v1/http_types.go
+++ b/pkg/operator/apis/monitoring/v1/http_types.go
@@ -97,13 +97,13 @@ func (s *SecretKeySelector) toPrometheusSecretRef(m PodMonitoringCRD, pool Prome
 
 	ns := s.Namespace
 	if ns == "" {
-		ns = metav1.NamespaceDefault
-	}
-	if m.IsNamespaceScoped() {
-		monitoringNamespace := m.GetNamespace()
-		if monitoringNamespace != "" {
-			ns = monitoringNamespace
+		if m.IsNamespaceScoped() {
+			ns = m.GetNamespace()
+		} else {
+			ns = metav1.NamespaceDefault
 		}
+	} else if m.IsNamespaceScoped() && ns != m.GetNamespace() {
+		return "", fmt.Errorf("must use namespace %q, got: %q", m.GetNamespace(), ns)
 	}
 
 	ref := fmt.Sprintf("%s/%s/%s", ns, s.Name, s.Key)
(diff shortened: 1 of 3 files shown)


pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -101,11 +101,8 @@ func (s *SecretKeySelector) toPrometheusSecretRef(m PodMonitoringCRD, pool Prome
 	}
```
TheSpiritXIII, line 101: It would be more clear to move the whole if statement here. We should be defaulting to whatever the PodMonitoring's namespace is _instead_ of the default namespace, e.g:

```go
ns := s.Namespace
if ns == "" {
	if m.IsNamespaceScoped() {
		ns = m.GetNamespace()
	} else {
		ns = metav1.NamespaceDefault
	}
}
```

pkg/operator/apis/monitoring/v1/http_types.go
```diff
@@ -101,11 +101,8 @@ func (s *SecretKeySelector) toPrometheusSecretRef(m PodMonitoringCRD, pool Prome
...
 	if m.IsNamespaceScoped() {
 		monitoringNamespace := m.GetNamespace()
-		if monitoringNamespace == "" {
-			monitoringNamespace = metav1.NamespaceDefault
-		}
-		if ns != monitoringNamespace {
-			return "", fmt.Errorf("must use namespace %q, got: %q", m.GetNamespace(), ns)
```
TheSpiritXIII, line 108: I'd like to keep this error handling because it would help catch issues in incorrect unit tests or if somehow this field was set.

We can simply it. Combined with the last suggestion, we could do:

```go
if ns == "" {
	...
} else if m.IsNamespaceScoped() && ns != m.GetNamespace() {
	return "", fmt.Errorf("must use namespace %q, got: %q", m.GetNamespace(), ns)
}
```

pkg/operator/apis/monitoring/v1/http_types_test.go
```diff
@@ -24,10 +24,13 @@ import (
 	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/secrets"
 )
 
-type secretNamespaceTestCase struct {
+type secretNamespacePodMonitoringTestCase struct {
```
TheSpiritXIII, line 27: nit: if the objects are the same, I would just reuse the same struct.

pkg/operator/apis/monitoring/v1/http_types_test.go
```diff
@@ -51,56 +54,34 @@ func TestClusterSecretKeySelector_toPrometheusSecretRef_PodMonitoring(t *testing
 		}
 	})
 
-	testCases := []secretNamespaceTestCase{
```
TheSpiritXIII, line 54: If we add the error back, let's also keep these test cases. We can keep all of the ClusterPodMonitoring changes, though!

TheSpiritXIII (changes requested):
Wow great catch! I can't believe this slipped through. 😅 

Overall, this is a really solid PR. I left a couple of minor comments. I would be happy to merge after they're addressed. I'll make sure this change is backported to all GKE versions where 0.12 is present.

pkg/operator/apis/monitoring/v1/http_types.go
fdaguin: 1b5b60e7cc23cfd66f979c91bf2a63a3bc06d6b6

pkg/operator/apis/monitoring/v1/http_types_test.go
fdaguin: 42b36b7d419f0ac1a7f60942ac1e61ef026eda21

pkg/operator/apis/monitoring/v1/http_types_test.go
fdaguin: 42b36b7d419f0ac1a7f60942ac1e61ef026eda21

fdaguin:
Thank you for your detailed feedback @TheSpiritXIII 🙏

TheSpiritXIII (approved):
Thank you very much!

PR #1001, opened 2024-06-04
refactor: cleanup export flags

TheSpiritXIII:
This PR cleans up all export flags so that:
- The code is more readable as it's more broken up by function.
- We could move these options to YAML one day and update them at runtime instead of at startup.

Changes:
- Ensures all flag values have their own struct.
- Move flag validation and defaulting to their own methods.
- Ensure all flag structs have only serializable values (e.g. `ExporterOpts.Lease` was moved out).
- Use URLs with `kingpin` (flag library) to avoid needing to parse strings manually.

I also broke up the monstrosity that was the `setup` package exporter constructor which returned a function that would create an exporter. The reason this was done was because first we setup flags and then returned a method that would create the exporter once the flags were parsed. Now you can just call `SetupFlags` and then call `exporter.New`.

Tested a Prometheus version built with these changes with `e2e` tests, and they pass.

### DIFF
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -80,8 +80,7 @@ func main() {
 		var err error
 		defaultProjectID, err = metadata.ProjectID()
 		if err != nil {
-			//nolint:errcheck
-			level.Warn(logger).Log("msg", "Unable to detect Google Cloud project", "err", err)
+			_ = level.Warn(logger).Log("msg", "Unable to detect Google Cloud project", "err", err)
 		}
 	}
 
@@ -96,11 +95,20 @@ func main() {
 	// we reuse that constant.
 	version, err := export.Version()
 	if err != nil {
-		//nolint:errcheck
-		level.Error(logger).Log("msg", "Unable to fetch module version", "err", err)
+		_ = level.Error(logger).Log("msg", "Unable to fetch module version", "err", err)
 		os.Exit(1)
 	}
-	newExporter := exportsetup.FromFlags(a, fmt.Sprintf("rule-evaluator/%s", version))
+
+	exporterOpts := export.ExporterOpts{
+		UserAgentProduct: fmt.Sprintf("rule-evaluator/%s", version),
+	}
+	exportsetup.ExporterOptsFlags(a, &exporterOpts)
+
+	metadataOpts := exportsetup.MetadataOpts{}
+	metadataOpts.SetupFlags(a)
+
+	haOpts := exportsetup.HAOptions{}
+	haOpts.SetupFlags(a)
 
 	notifierOptions := notifier.Options{Registerer: reg}
 
@@ -131,22 +139,19 @@ func main() {
 
 	extraArgs, err := exportsetup.ExtraArgs()
 	if err != nil {
-		//nolint:errcheck
-		level.Error(logger).Log("msg", "Error parsing commandline arguments", "err", err)
+		_ = level.Error(logger).Log("msg", "Error parsing commandline arguments", "err", err)
 		a.Usage(os.Args[1:])
 		os.Exit(2)
 	}
 	if _, err := a.Parse(append(os.Args[1:], extraArgs...)); err != nil {
-		//nolint:errcheck
-		level.Error(logger).Log("msg", "Error parsing commandline arguments", "err", err)
+		_ = level.Error(logger).Log("msg", "Error parsing commandline arguments", "err", err)
 		a.Usage(os.Args[1:])
 		os.Exit(2)
 	}
 	startTime := time.Now()
 
 	if *projectID == "" {
-		//nolint:errcheck
-		level.Error(logger).Log("msg", "no --query.project-id was specified or could be derived from the environment")
+		_ = level.Error(logger).Log("msg", "no --query.project-id was specified or could be derived from the environment")
 		os.Exit(2)
 	}
 
@@ -157,8 +162,7 @@ func main() {
 		var err error
 		generatorURL, err = url.Parse(*generatorURLStr)
 		if err != nil {
-			//nolint:errcheck
-			level.Error(logger).Log("msg", "Invalid --query.generator-url", "err", err)
+			_ = level.Error(logger).Log("msg", "Invalid --query.generator-url", "err", err)
 			os.Exit(2)
 		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/export/setup/setup_test.go
```diff
@@ -35,19 +35,34 @@ func TestFromFlags_NotOnGCE(t *testing.T) {
...
 	// Fake app invocation.
 	if _, err := fake.Parse(nil); err != nil {
 		t.Fatal(err)
 	}
 
-	// Make sure constructor does is not stuck forever.
-	_, _ = newExport(ctx, log.NewLogfmtLogger(os.Stderr), nil)
+	done := make(chan struct{}, 1)
+	go func() {
+		metadataOpts.ExtractMetadata(logger, &exporterOpts)
```
TheSpiritXIII, line 56: > @pintohutch [`7fe3d0b` (#974)](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974/commits/7fe3d0b10a0a0258d55440ef27bc473102e02fe1#r1661217052)
> question: If we do want to make this change, wouldn't we also need to update the tests in `TestTryPopulateUnspecifiedFromMetadata` to ensure timeouts are correctly updated and respected in the config?

Can you clarify why we might need to? The default/logic hasn't been changed, only moved.

pkg/export/setup/setup.go
```diff
@@ -239,6 +243,43 @@ func ExtraArgs() ([]string, error) {
 	return shlex.Split(os.Getenv(ExtraArgsEnvvar))
 }
 
+type MetadataOpts struct {
```
TheSpiritXIII, line 246: > @pintohutch  [`7fe3d0b` (#974)](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974/commits/7fe3d0b10a0a0258d55440ef27bc473102e02fe1#r1661201261)
> question: Why separate the MetadataOpts config?

My goal was to split up configurations used in setup.go and export.go. Previously these values were just variables in `FromFlags`.

I can combine `MetadataOpts` and `HAOpts` into something like `SetupOpts` if that makes more sense, but the separation meant clearer separation of concerns.

pkg/export/setup/setup.go
```diff
@@ -90,130 +90,134 @@ func Global() *export.Exporter {
...
-			// use if labels or external label settings does not have those set. Those will
-			// be used as crucial labels for export to work against GCM's Prometheus target.
-			//
-			// Set a hard time limit due to readiness and liveliness probes during this stage.
-			mctx, cancel := context.WithTimeout(context.Background(), metadataFetchTimeout)
-			tryPopulateUnspecifiedFromMetadata(mctx, logger, &opts)
-			cancel()
-			_ = level.Debug(logger).Log("msg", "best-effort on-GCE metadata gathering finished")
-		}
+type HAOptions struct {
```
TheSpiritXIII, line 166: > @pintohutch [`7fe3d0b` (#974)](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974/commits/7fe3d0b10a0a0258d55440ef27bc473102e02fe1#r1661197690)
> question: Why separate the haBackend config?

See similar justification on `MetadataOpts`.

pkg/export/setup/setup.go
```diff
@@ -90,130 +90,134 @@ func Global() *export.Exporter {
...
-// Potential risky logic can be moved to the returned function we return here.
-// See b/344740239 on how hard is to debug regressions here.
-func FromFlags(a *kingpin.Application, userAgentProduct string) func(context.Context, log.Logger, prometheus.Registerer) (*export.Exporter, error) {
-	var (
-		metadataFetchTimeout time.Duration
-		opts                 export.ExporterOpts
-	)
-	opts.UserAgentProduct = userAgentProduct
-
+func ExporterOptsFlags(a *kingpin.Application, opts *export.ExporterOpts) {
```
TheSpiritXIII, line 93: > @pintohutch [`7fe3d0b` (#974)](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974/commits/7fe3d0b10a0a0258d55440ef27bc473102e02fe1#r1661194668)
> suggestion: It feels brittle to rely on opts being .Default()ed before this is called every time.
>
> What if we did the defaulting of opts in this routine, like before? Tying the flag/config definitions and default behavior centrally makes a lot of sense to me.
> 
> I think this also comes back to my previous comment of around how we envision clients handling the ExporterOpts struct.

That makes a lot of sense, but I'd still like the `Default` method so that unit tests can still use defaults without setting up flags or hardcoding them. To my shock, we were doing the former so this change prevents this. :)

pkg/export/export.go
```diff
@@ -310,7 +347,7 @@ func newMetricClient(ctx context.Context, opts ExporterOpts) (*monitoring.Metric
 }
 
 // New returns a new Cloud Monitoring Exporter.
-func New(ctx context.Context, logger log.Logger, reg prometheus.Registerer, opts ExporterOpts) (*Exporter, error) {
+func New(ctx context.Context, logger log.Logger, reg prometheus.Registerer, opts ExporterOpts, lease Lease) (*Exporter, error) {
```
TheSpiritXIII, line 350: > @pintohutch [`7fe3d0b` (#974)](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974/commits/7fe3d0b10a0a0258d55440ef27bc473102e02fe1#r1661116963)
> question: How will callers of the export library refresh ExporterOpts?
>
> This change looks to be focused on refreshing the Exporter only with a given ExporterOpts struct.

I will be adding this change to `ApplyConfig`. It will take an optional `ExporterOpts`. If `nil`, nothing changes: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974

pkg/export/gcm/promtest/local_export.go
```diff
@@ -93,18 +93,14 @@ func (l *localExportWithGCM) start(t testing.TB, _ e2e.Environment) (v1.API, map
...
-		UserAgentEnv:     "pe-github-action-test",
-		Endpoint:         "monitoring.googleapis.com:443",
-		Compression:      "none",
-		MetricTypePrefix: export.MetricTypePrefix,
-
-		Cluster:   cluster,
-		Location:  location,
-		ProjectID: creds.ProjectID,
-
+	exporterOpts := export.ExporterOpts{
```
TheSpiritXIII, line 96: > @pintohutch [`7fe3d0b` (#974)](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/974/commits/7fe3d0b10a0a0258d55440ef27bc473102e02fe1#r1661175990)
> question: Why the opts changes here?
>
> @bwplotka may be the better resource on reviewing this package, as I am less familiar with it.

They were hard-coded defaults, which I got rid of.

TheSpiritXIII:
For ease of making this PR easier to review, I split up the rule-evaluator refactors to this PR: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1058

cmd/rule-evaluator/main.go
```diff
@@ -131,22 +139,19 @@ func main() {
 
 	extraArgs, err := exportsetup.ExtraArgs()
 	if err != nil {
-		//nolint:errcheck
-		level.Error(logger).Log("msg", "Error parsing commandline arguments", "err", err)
+		_ = level.Error(logger).Log("msg", "Error parsing commandline arguments", "err", err)
```
bernot-dev, line 142: Question: Is throwing away the return better than nolint? I honestly don't have a strong preference, just curious if you have thoughts.

TheSpiritXIII: I typically have a difficult time remembering the nolint warning so I find this easier. Less boilerplate, clear and concise, and future-proof from other linters (when I was in the Java world, there were overlapping rules between linters and maybe one day could happen in Golang, who knows).

pkg/export/export.go
```diff
@@ -208,6 +208,38 @@ type ExporterOpts struct {
...
+	}
+	if opts.Efficiency.ShardBufferSize == 0 {
+		opts.Efficiency.ShardBufferSize = DefaultShardBufferSize
+	}
+
+	if opts.Endpoint == "" {
+		opts.Endpoint = "monitoring.googleapis.com:443"
+	}
+	if opts.Compression == "" {
+		opts.Compression = CompressionNone
```
bernot-dev, line 226: Question: What compression does this control? Should we use gzip?

TheSpiritXIII: This is an opt-in feature when users have many or large PodMonitorings and the corresponding ConfigMap reaches etcd limits of 1 MiB. The default is not changing here (the old code set the default when setting up the flag), and hopefully it's now clearer in the code what all of the defaults are.

bernot-dev (approved):
LGTM

PR #1002, opened 2024-06-05
fix: use clusterpodmonitoring for nvidia-dcgm

pintohutch:
NVIDIA dcgm-exporter captures and records the "pod", "namespace", and "container" labels. We should honor those in our relabeling.

Hence, we use a ClusterPodMonitoring with an empty .targetLabels.metadata (akin to what we do for KSM) to preserve them.


pintohutch:
cc @huygaa11

bernot-dev (approved):
LGTM

PR #1009, opened 2024-06-10
Update DCGM sample YAML so that /home/kubernetes/bin/nvidia is not created if it does not exist

yqlu:
Context: b/345755257

Verified via https://kubernetes.io/docs/concepts/storage/volumes/?spm=a2c63.p38356.0.0.28874dcbfWkCRx#hostpath-volume-types this is a well-defined option.

### DIFF
--- a/examples/nvidia-dcgm/exporter.yaml
+++ b/examples/nvidia-dcgm/exporter.yaml
@@ -91,6 +91,7 @@ spec:
       - name: nvidia-install-dir-host
         hostPath:
           path: /home/kubernetes/bin/nvidia
+          type: Directory
       - name: pod-resources
         hostPath:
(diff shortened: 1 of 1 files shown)


pintohutch (approved):

pintohutch (changes requested):
Actually - wdyt about using "DirectoryOrCreate" if someone is deploying outside of GKE where that directory may not exist?

yqlu:
Looking at b/343384034#comment42 (lots of history in that issue) it sounds like we explicitly do not want `DirectoryOrCreate` because of the race condition? Or am I misreading it.

pintohutch:
Ah - nice find! Yea then nvm my comment. Let's proceed with "Directory" then.

pintohutch (approved):

PR #1010, opened 2024-06-10
fix: use proper image versions in install manifest

pintohutch:
Bump the operator-related images up to v0.10.0-gke.6.

Bump the prometheus and alertmanager images as well.

### DIFF
--- a/charts/values.global.yaml
+++ b/charts/values.global.yaml
@@ -20,27 +20,27 @@ version: 0.13.0
 images:
   bash:
     image: gke.gcr.io/gke-distroless/bash
-    tag: "20220419" # NOTE: Has to be quoted otherwise it will be treated as a number.
+    tag: gke_distroless_20240207.00_p0
   alertmanager:
     image: gke.gcr.io/prometheus-engine/alertmanager
-    tag: v0.25.1-gmp.2-gke.0
+    tag: v0.25.1-gmp.3-gke.0
   prometheus:
     # TODO(bwplotka): Change to "v2.45.3-gmp.4-gke.0" once tags are cloned.
     image: gke.gcr.io/prometheus-engine/prometheus@sha256
-    tag: 7473d52f4a3e563e6377f8a6183091f25192b1e0705dd0933903e800bd69b7b2
+    tag: 3115cb238c9dd9abe4f4cf8ed3daeb1e6e9ee045bf24f8614e144b3a642f3464
   configReloader:
     image: gke.gcr.io/prometheus-engine/config-reloader
-    tag: v0.9.0-gke.1
+    tag: v0.10.0-gke.6
   operator:
     image: gke.gcr.io/prometheus-engine/operator
-    tag: v0.9.0-gke.1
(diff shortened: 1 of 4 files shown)


TheSpiritXIII (approved):
Thanks!

charts/values.global.yaml
```diff
@@ -20,27 +20,27 @@ version: 0.13.0
...
-    tag: v0.9.0-gke.1
+    tag: v0.10.0-gke.6
   ruleEvaluator:
     image: gke.gcr.io/prometheus-engine/rule-evaluator
-    tag: v0.9.0-gke.1
+    tag: v0.10.0-gke.6
   datasourceSyncer:
-    image: gcr.io/gke-release/prometheus-engine/datasource-syncer
+    image: gke.gcr.io/prometheus-engine/datasource-syncer
     #TODO(macxamin) Sync CURRENT_DATASOURCE_SYNCER_TAG with CURRENT_TAG
```
maxamins, line 42: This comment can be removed.

manifests/operator.yaml
```diff
@@ -534,7 +534,7 @@ spec:
       priorityClassName: gmp-critical
       containers:
       - name: operator
-        image: gke.gcr.io/prometheus-engine/operator:v0.9.0-gke.1
+        image: gke.gcr.io/prometheus-engine/operator:v0.10.0-gke.6
```
bwplotka, line 537: I doubt this works due to https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1013
 
(with readOnlyRootFilesystem: true)

Have you tested it?

pintohutch: Ah yes - indeed https://github.com/GoogleCloudPlatform/prometheus-engine/pull/944 breaks this.

@bernot-dev or @TheSpiritXIII do you know if we have an image published we could use that works with that change? Presumably HEAD worked (`v0.13.0-rc.0`).

OTOH maybe trying to "fix" our nightly manifest in this way isn't a great idea and I should abandon the PR as it isn't possible to always have a working manifest in HEAD anyway due to GKE build pipeline delays...

bwplotka: I mean, you are not making it worse so let's go =D

Good point that we could have issues going forward, but we have same problem with AM and Prometheus and for those we HAVE to make them workable in atomic way otherwise tests won't pass. It's just with PE images we are "allowed" to cheat a bit.

I wonder if splitting manifests to OSS official ones VS source for managed GMP should be different things e.g. for GMP we could simply use helm templates directly and have different values (: We could actually get rid of kustomize (or move kustomize here). cc @TheSpiritXIII

bwplotka (approved):
LGTM, mod comments.

manifests/operator.yaml
bwplotka: The alternative (or actually doing both ideas might work) would be to have official "dev" registry we will push ALWAYS and register the version here somehow (maybe "main" tag after all). Not tested, but it will be builded like any other image. Then we can promote to official prod registry and tag on the final release.

PR #1021, opened 2024-06-11
fix(export): make FromFlags & metadata call more reliable and observable

bwplotka:
TBD


pkg/export/setup/setup_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/export"
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/go-kit/log"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestFromFlags_NotOnGCE(t *testing.T) {
+	// Asserting there is actually no GCE underneath.
+	if metadata.OnGCE() {
+		t.Fatal("This test assumes we don't run on GCP")
```
TheSpiritXIII, line 35: Suggestion: `t.Skip`?

pkg/export/setup/setup_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Make sure constructor does is not stuck forever.
+	_, _ = newExport(ctx, log.NewLogfmtLogger(os.Stderr), nil)
+}
+
+// Regression test for b/344740239. We ensure that even stuck metadata servers
+// calls will timeout correctly (we propagate context properly).
+func TestTryPopulateUnspecifiedFromMetadata(t *testing.T) {
+	// Asserting there is actually no GCE underneath.
+	if metadata.OnGCE() {
```
TheSpiritXIII, line 57: Suggestion: What if we just mock the metadata server? e.g. we create an interface and pass that to `tryPopulateUnspecifiedFromMetadata`?

TheSpiritXIII (changes requested):
Thank you for the investigation and solid change!

pkg/export/setup/setup_test.go
bwplotka: Yea, what's the advantage of this versus what's now?

We can mock, but then we just test if we pass context correctly.

 Here we test the metadata client to also pass context correctly. I know we shouldn't generally test not our code.. but well, our product depends on this we just got regression because of that, thus let's maybe test this case?

pkg/export/setup/setup.go
```diff
@@ -230,3 +230,73 @@ func loadKubeConfig(kubeconfigPath string) (*rest.Config, error) {
 func ExtraArgs() ([]string, error) {
 	return shlex.Split(os.Getenv(ExtraArgsEnvvar))
 }
+
+// tryPopulateUnspecifiedFromMetadata assumes we are in GCP, with unknown state
+// of the metadata server. This is best-effort, so when the metadata server
+// is not accessible on GCP, because it was disabled (404 errors), or not accessible
+// (connection refused, slow network, e.g. sandbox + metadata disabled)
+// it's a noop. Make sure to pass context with a timeout.
+func tryPopulateUnspecifiedFromMetadata(ctx context.Context, logger log.Logger, opts *export.ExporterOpts) {
```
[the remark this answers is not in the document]
bwplotka: I doubt we hit the same problem, but that would be a good practice and little harm! I want to keep this PR small for bugfix though, let me check if this can be still "small" with operator changes.

pkg/export/setup/setup_test.go
bwplotka: Why? Do we expect to run in GCP? I don't think so, so let's fail fast (assert our assumption), WDYT? 🤔

pkg/export/setup/setup.go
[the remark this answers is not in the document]
bwplotka: Ok, I checked and indeed it would be nice, but I would postpone this PR until https://github.com/googleapis/google-cloud-go/pull/10370 is merged (if anything). Will be easier. For now we could merge this and patch 0.12.0?

bwplotka:
I tested this PR on my repro cluster.

* std node:  works
* sandbox node: works 
* no metadata node: works (we pass data via flags so it's fine)
* sandbox, no metadata: collector runs, but export is unauth as expected

Metrics flow for 3 nodes as expected:

image

From this PR POV it's rdy to go I think

bwplotka:
After I did https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-managed#explicit-credentials all works:

image

TheSpiritXIII (approved):
I'm okay with all of your reasons. Thank you so much for the metadata library change! Much appreciated and great idea.

pkg/export/setup/setup.go
```diff
@@ -184,7 +169,30 @@ func FromFlags(a *kingpin.Application, userAgentProduct string) func(context.Con
...
+			_ = level.Debug(logger).Log("msg", "detected we might run on GCE node; attempting metadata server access", "timeout", metadataFetchTimeout.String())
+			// When, potentially, on GCE we attempt to populate some, unspecified option entries
+			// like project ID, cluster, location, zone and user agent from GCP metadata server.
+			//
+			// This will be used to get *some* data, if not specified override by flags, to
+			// use if labels or external label settings does not have those set. Those will
+			// be used as crucial labels for export to work against GCM's Prometheus target.
+			//
+			// Set a hard time limit due to readiness and liveliness probes during this stage.
+			mctx, cancel := context.WithTimeout(context.Background(), metadataFetchTimeout)
```
pintohutch, line 190: nit: should this have been `ctx` instead of `context.Background()`?

pkg/export/setup/setup.go
```diff
@@ -230,3 +238,73 @@ func loadKubeConfig(kubeconfigPath string) (*rest.Config, error) {
 func ExtraArgs() ([]string, error) {
 	return shlex.Split(os.Getenv(ExtraArgsEnvvar))
 }
+
+// tryPopulateUnspecifiedFromMetadata assumes we are in GCP, with unknown state
+// of the metadata server. This is best-effort, so when the metadata server
+// is not accessible on GCP, because it was disabled (404 errors), or not accessible
+// (connection refused, slow network, e.g. sandbox + metadata disabled)
+// it's a noop. Make sure to pass context with a timeout.
```
pintohutch, line 246: suggestion: Could we add a TODO here referencing we can update/slim this code down once  https://github.com/googleapis/google-cloud-go/pull/10370 is merged?

PR #1028, opened 2024-06-13
feat: add max throughput option

bernot-dev:


charts/operator/templates/collector.yaml
```diff
@@ -178,6 +178,191 @@ spec:
...
+        ports:
+        - name: cfg-rel-metrics
+          containerPort: 19091
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: spec.nodeName
+        resources: {{- toYaml $.Values.resources.bash | nindent 10}}
```
pintohutch, line 257: suggestion: shouldn't this be `resources.configReloader`? Same with L76.

pintohutch (commented):
Thanks for putting this up!

I'm wondering: does this _have_ to live in Github? Could we use kustomize or helm to create the additional `max-collector` DaemonSet manifest as part of the GKE component internally by patching and duplicating the `collector` manifest?

The reason I'm asking is that users who self-install don't necessarily need two DaemonSet as they're free to adjust the one to meet their needs.

OTOH having two DaemonSets based on a node label is kind of a neat, resource optimization idea, even if the approach is rather coarse.

Thoughts?

bernot-dev:
> I'm wondering: does this _have_ to live in Github? Could we use kustomize or helm to create the additional `max-collector` DaemonSet manifest as part of the GKE component internally by patching and duplicating the `collector` manifest?
> 
> The reason I'm asking is that users who self-install don't necessarily need two DaemonSet as they're free to adjust the one to meet their needs.
> 
> OTOH having two DaemonSets based on a node label is kind of a neat, resource optimization idea, even if the approach is rather coarse.
> 
> Thoughts?

I think there are perfectly reasonable cases for self-install scenarios to use both DaemonSets, including heterogeneous node sizes. And the overhead of one additional DaemonSet that is often `Desired: 0` pods is a small price to pay. Self-install users who don't want the extra DaemonSet can easily delete it, as well. It may not be necessary for all users, but I don't think it hurts them.

charts/operator/templates/collector.yaml
bernot-dev: Yup. Looks like that was incorrect before. Good catch!

TheSpiritXIII:
I'm wondering if this has to live on GitHub as well. It's more effort for us to document this feature and have self-installed users set it up than it is for them to just edit the resources via Helm (which is already a variable on our configurations I belive) or via Kustomize.

bernot-dev:
Decided to move this feature to `examples` for OSS use cases.

PR #1031, opened 2024-06-17
refactor: use scale subresource to update replica counts

TheSpiritXIII:
This PR helps us get rid of our Deployment and StatefulSet RBAC policies one day (in favor of only allowing the scale sub-resource).

### DIFF
--- a/charts/operator/templates/role.yaml
+++ b/charts/operator/templates/role.yaml
@@ -89,12 +89,22 @@ rules:
   resourceNames: ["rule-evaluator"]
   verbs: ["get", "patch", "update"]
 - resources:
+  - deployments/scale
+  apiGroups: ["apps"]
+  resourceNames: ["rule-evaluator"]
+  verbs: ["get", "patch", "update"]
+- resources:
   - services
   apiGroups: [""]
   resourceNames: ["alertmanager"]
   verbs: ["get", "list", "watch"]
 - resources:
   - statefulsets
+  apiGroups: ["apps"]
+  resourceNames: ["alertmanager"]
+  verbs: ["get", "patch", "update"]
+- resources:
+  - statefulsets/scale
   apiGroups: ["apps"]
   resourceNames: ["alertmanager"]
   verbs: ["get", "patch", "update"]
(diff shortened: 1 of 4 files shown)


charts/operator/templates/role.yaml
```diff
@@ -88,6 +88,11 @@ rules:
   apiGroups: ["apps"]
   resourceNames: ["rule-evaluator"]
   verbs: ["get", "patch", "update"]
+- resources:
+  - deployments/scale
```
bernot-dev, line 92: I'm confused...why add these permissions now, which are a subset of the permissions that already exist? Why not just scale back the existing permissions when the rest of the necessary changes are ready?

TheSpiritXIII: I tried running this change without this permission and it failed. You would think they are a subset but that's not how it's implemented in Kubernetes...

pkg/operator/rules_test.go
```diff
@@ -473,3 +477,157 @@ func TestScaleRuleConsumers(t *testing.T) {
...
+		body = patchOptions.SubResourceBody
+	}
+
+	return c.client.Patch(ctx, body, patch, &patchOptions.PatchOptions)
+}
+
+func extractScale(obj client.Object) (autoscalingv1.Scale, error) {
+	switch obj := obj.(type) {
+	case *appsv1.Deployment:
+		var replicas int32 = 1
```
bernot-dev, line 553: go/go-style/best-practices#vardeclinitialization

pkg/operator/rules_test.go
```diff
@@ -473,3 +477,157 @@ func TestScaleRuleConsumers(t *testing.T) {
...
+	body := obj
+	if patchOptions.SubResourceBody != nil {
+		body = patchOptions.SubResourceBody
+	}
+
+	return c.client.Patch(ctx, body, patch, &patchOptions.PatchOptions)
+}
+
+func extractScale(obj client.Object) (autoscalingv1.Scale, error) {
+	switch obj := obj.(type) {
```
bernot-dev, line 551: Thought: I'm sad that this type switch is necessary. But it seems to be.

bernot-dev (approved):
LGTM

PR #1037, opened 2024-06-20
feat: generate max-throughput collector manifest

bernot-dev:
Generate an example manifest with the max-throughput resource specs.


manifests/operator.yaml
```diff
@@ -488,6 +488,10 @@ spec:
                 operator: In
                 values:
                 - linux
+              - key: cloud.google.com/gke-monitoring-variant
```
TheSpiritXIII, line 491: nit: Albeit extremely rare, I can envision this being problematic. If a user assumes that MAX_THROUGHPUT is enabled by default when you self-install and they start tagging their nodes, they won't get _any_ collectors running on those nodes anymore.

TheSpiritXIII (approved):
LGTM!

pintohutch (approved):

manifests/operator.yaml
bernot-dev: Agreed. However, I weighed this against the situation where a user does try to enable MAX_THROUGHPUT but forgets to add this selector and ends up running 2 instances of the collector per node.

Compared to the scenario you described, I believe this change is:

- More discoverable when misconfigured. Metrics will stopped being collected at the time of the node labeling. This should be a strong signal that GMP is not configured correctly. Two collector pods scraping the same targets would *kind of* work, with lots of out of order errors, I think.
- Less error-prone during setup. Simply apply an additional manifest, instead of also patching an existing resource (and keep the patches in sync). Likewise, this simplifies the modifications we make for GKE environments.

It seems like a user is unlikely to apply that specific label without some guidance/documentation they are referencing. And if they do, it should be relatively easy to troubleshoot because the collector pods would not be scheduled and the DaemonSet would show that the desired # of pods is 0.

PR #1052, opened 2024-07-02
feat: include Google Cloud links for alerts

TheSpiritXIII:
Pre-req: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001 (this helps me reuse the projectID flag from the --export.project_id flag).

See b/350551533 for details.

### DIFF
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -67,6 +67,12 @@ import (
 
 const projectIDVar = "PROJECT_ID"
 
+var googleCloudBaseURL = url.URL{
+	Scheme: "https",
+	Host:   "console.cloud.google.com",
+	Path:   "/monitoring/metrics-explorer",
+}
+
 func main() {
 	logger := log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
 	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
@@ -405,7 +411,7 @@ func (opts *evaluatorOptions) setupFlags(a *kingpin.Application) {
 		URLVar(&opts.TargetURL)
 
 	a.Flag("query.generator-url", "The base URL used for the generator URL in the alert notification payload. Should point to an instance of a query frontend that accesses the same data as --query.target-url.").
-		PlaceHolder("<URL>").
+		Default(googleCloudBaseURL.String()).
 		URLVar(&opts.GeneratorURL)
 
 	a.Flag("query.credentials-file", "Credentials file for OAuth2 authentication with --query.target-url.").
@@ -495,7 +501,7 @@ func QueryFunc(ctx context.Context, q string, t time.Time, v1api v1.API) (parser
 }
 
 // sendAlerts returns the rules.NotifyFunc for a Notifier.
-func sendAlerts(s *notifier.Manager, externalURL *url.URL) rules.NotifyFunc {
+func sendAlerts(s *notifier.Manager, projectID string, externalURL *url.URL) rules.NotifyFunc {
 	return func(_ context.Context, expr string, alerts ...*rules.Alert) {
 		var res []*notifier.Alert
 		for _, alert := range alerts {
@@ -510,7 +516,13 @@ func sendAlerts(s *notifier.Manager, externalURL *url.URL) rules.NotifyFunc {
 				a.EndsAt = alert.ValidUntil
 			}
 			if externalURL != nil {
-				a.GeneratorURL = externalURL.String() + strutil.TableLinkForExpression(expr)
+				if externalURL.String() == googleCloudBaseURL.String() {
+					// If it's a GCM link (default), create the full URL for the alert.
+					a.GeneratorURL = googleCloudLink(projectID, expr, alert.FiredAt, alert.FiredAt.Add(-time.Hour)).String()
+				} else {
+					// Otherwise, if it was specified we assume it points to a Prometheus frontend.
+					a.GeneratorURL = externalURL.String() + strutil.TableLinkForExpression(expr)
+				}
 			}
 			res = append(res, a)
 		}
@@ -518,6 +530,44 @@ func sendAlerts(s *notifier.Manager, externalURL *url.URL) rules.NotifyFunc {
 			s.Send(res...)
 		}
 	}
+}
+
+func googleCloudLink(projectID, expr string, endTime, startTime time.Time) *url.URL {
+	// Clone URL to avoid mutating the original.
+	url := googleCloudBaseURL
+	// Note: The URL API was reverse-engineered.
+	if !endTime.IsZero() {
+		url.Path += ";endTime=" + endTime.Format(time.RFC3339)
+	}
+	if !startTime.IsZero() {
+		url.Path += ";startTime=" + startTime.Format(time.RFC3339)
+	}
+
+	values := url.Query()
+	values.Set("project", projectID)
+
+	if expr != "" {
+		// These settings reflect the default on metrics explorer for majority of use-cases.
+		pageState := map[string]any{
+			"xyChart": map[string]any{
+				"dataSets": []map[string]any{
+					{
+						"prometheusQuery": expr,
+					},
+				},
+			},
+		}
+		// Note, this also escapes the JSON.
+		pageStateValue, err := json.Marshal(pageState)
+		if err != nil {
+			panic(err)
+		}
+		values.Set("pageState", string(pageStateValue))
+	}
+
+	// Escapes the query (which may have escaped JSON).
+	url.RawQuery = values.Encode()
+	return &url
 }
 
 type reloader struct {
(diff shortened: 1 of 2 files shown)


cmd/rule-evaluator/main.go
```diff
@@ -96,155 +106,107 @@ func main() {
...
-
-	queryCredentialsFile := a.Flag("query.credentials-file", "Credentials file for OAuth2 authentication with --query.target-url.").
-		Default("").String()
-
-	disableAuth := a.Flag("query.debug.disable-auth", "Disable authentication (for debugging purposes).").
-		Default("false").Bool()
+	exporterOpts := export.ExporterOpts{
+		UserAgentProduct: fmt.Sprintf("rule-evaluator/%s", version),
+	}
+	exportsetup.ExporterOptsFlags(a, &exporterOpts)
```
bwplotka, line 116: Hm, that changes the flag surface completely, break the users essentially, right? (`export.` prefix on all flags https://github.com/GoogleCloudPlatform/prometheus-engine/blob/afe2103347608bf63eb7555d208d638465dd3d7f/pkg/export/setup/setup.go)

Do we want this?

TheSpiritXIII: Please see https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001 first. On the rule-evaluator, we always had two sets of flags, query and export. Those have not changed here, but split up and exposed in the public API. The query options are now part of the `evaluatorOptions`.

bwplotka: PR: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1053

pintohutch:
suggestion: Change the merge base from `main` to `TheSpiritXIII/refactor-flags` for easier review-ability.

cmd/rule-evaluator/main.go
TheSpiritXIII: Done. Let me know what you think of it!

TheSpiritXIII:
> suggestion: Change the merge base from main to TheSpiritXIII/refactor-flags for easier review-ability.

@pintohutch done!

cmd/rule-evaluator/main.go
```diff
@@ -569,6 +581,34 @@ func sendAlerts(s *notifier.Manager, externalURL *url.URL) rules.NotifyFunc {
 	}
 }
 
+func googleCloudLink(projectID, expr string, endTime, startTime time.Time) string {
+	// Clone URL to avoid mutating the original.
+	url := googleCloudBaseURL
+	url.Path += ";endTime=" + endTime.Format(time.RFC3339)
+	url.Path += ";startTime=" + startTime.Format(time.RFC3339)
+
+	// These settings reflect the default on metrics explorer for majority of use-cases.
```
pintohutch, line 590: suggestion: Add a comment about URL format.

Is there a reference doc we can link here to inform the URL format and parameters? Or did you reverse-engineer this from a raw URL? Even that's the case, giving an example in a comment would be helpful.

cmd/rule-evaluator/main.go
```diff
@@ -67,6 +67,16 @@ import (
 
 const projectIDVar = "PROJECT_ID"
 
+var googleCloudBaseURL = urlMustParse("https://console.cloud.google.com/monitoring/metrics-explorer")
```
pintohutch, line 70: suggestion: store the url string as a const `string`.

Then we can use it as the default value for `query.generator-url` in the flag declaration along with the `projectID` defaulting like we do [elsewhere](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/0c597767c1883093107b0f875fca14bafd9480de/cmd/rule-evaluator/main.go#L111).

We can then build/parse the full URL struct [later](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/0c597767c1883093107b0f875fca14bafd9480de/cmd/rule-evaluator/main.go#L153-L158) to be passed to `sendAlerts` as-is without changing its function signature.

cmd/rule-evaluator/main.go
```diff
@@ -560,6 +570,8 @@ func sendAlerts(s *notifier.Manager, externalURL *url.URL) rules.NotifyFunc {
 			}
 			if externalURL != nil {
```
pintohutch, line 571: suggestion: Change the condition to check if the `externalURL` is prefixed with the (suggested) GCM `const`.

e.g.
```go
// If it's a GCM link (default), create the full URL for the alert.
if strings.HasPrefix(externalURL, gcmURL) {
	a.GeneratorURL = googleCloudLink(externalURL, expr, alert.FiredAt, alert.FiredAt.Add(-time.Hour))
// Otherwise, if it was specified we assume it points to a Prometheus frontend.
} else if externalURL != nil {
	a.GeneratorURL = externalURL.String() + strutil.TableLinkForExpression(expr)
}
```

cmd/rule-evaluator/main.go
```diff
@@ -569,6 +581,34 @@ func sendAlerts(s *notifier.Manager, externalURL *url.URL) rules.NotifyFunc {
...
+	url := googleCloudBaseURL
+	url.Path += ";endTime=" + endTime.Format(time.RFC3339)
+	url.Path += ";startTime=" + startTime.Format(time.RFC3339)
+
+	// These settings reflect the default on metrics explorer for majority of use-cases.
+	pageState := map[string]any{
+		"xyChart": map[string]any{
+			"dataSets": []map[string]any{
+				{
+					"prometheusQuery": expr,
```
pintohutch, line 595: question: Does `expr` need to be [escaped](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/vendor/github.com/prometheus/prometheus/util/strutil/strconv.go#L29)?

pintohutch (commented):
Nice - left some comments for ya.

IIUC @lyanco suggested making the GCM link the default.

If that's the case, we should update the `OperatorConfig` [docs](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/0c597767c1883093107b0f875fca14bafd9480de/pkg/operator/apis/monitoring/v1/operator_types.go#L152-L154) as well to highlight this as a potentially breaking change.

cmd/rule-evaluator/main.go
TheSpiritXIII: Yes. Using `json.Marshal` escapes the JSON quotes. Then, we use `Query().Encode()`, which encodes the JSON within the URL query.

cmd/rule-evaluator/main.go
TheSpiritXIII: Sadly, it was reverse-engineered.

cmd/rule-evaluator/main.go
TheSpiritXIII: I'm not sure if prefix makes sense. I made it check for equals, which would be the normalized URL since it's already been parsed. We can always change it to prefix if/when we add full templating support.

cmd/rule-evaluator/main.go
TheSpiritXIII: I'm not sure the `projectID` defaulting makes sense here because the `projectID` field is not enough. Should we encode the expression in the URL? How about the start and end times? More importantly, where does what gets encoded into the URL end? How do we document this API?

I personally feel like this would be a bit of a feature-creep to add when you consider all the possibilities. Fortunately, how this PR was designed, we always have the possibility to add it in, in a backwards-compatible fashion.

TheSpiritXIII (commented):
Thanks for the review @pintohutch! I'm opening up some discussions. Please let me know what you think!

cmd/rule-evaluator/main.go
pintohutch: Ah - agreed, this is better. Thanks!

cmd/rule-evaluator/main.go
pintohutch: cc @yqlu in case he knows of any docs around the ME URLs

cmd/rule-evaluator/main.go
pintohutch: SGTM

pintohutch (commented):
Thanks!

Can we also update the OperatorConfig [docs](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/0c597767c1883093107b0f875fca14bafd9480de/pkg/operator/apis/monitoring/v1/operator_types.go#L152-L154) to mention the change in default behavior?

cmd/rule-evaluator/main.go
TheSpiritXIII: That's a very good point. I still feel strongly against adding templating, so I went with a different approach: the default is still the raw URL but we add the project manually if we detect it.

cmd/rule-evaluator/main.go
```diff
@@ -67,6 +67,16 @@ import (
 
 const projectIDVar = "PROJECT_ID"
 
+var googleCloudBaseURL = urlMustParse("https://console.cloud.google.com/monitoring/metrics-explorer")
+
+func urlMustParse(s string) url.URL {
+	u, err := url.Parse(s)
+	if err != nil {
+		panic(err)
```
bernot-dev, line 75: Suggestion: Instead of using this helper that can panic, why not just define the literal?

```
url.URL{
	Scheme: "https",
	Host:   "console.cloud.google.com",
	Path:   "/monitoring/metrics-explorer",
}
```

TheSpiritXIII: Great idea. Thanks!

bernot-dev (approved):
LGTM

cmd/rule-evaluator/main.go
```diff
@@ -405,7 +411,7 @@ func (opts *evaluatorOptions) setupFlags(a *kingpin.Application) {
 		URLVar(&opts.TargetURL)
 
 	a.Flag("query.generator-url", "The base URL used for the generator URL in the alert notification payload. Should point to an instance of a query frontend that accesses the same data as --query.target-url.").
-		PlaceHolder("<URL>").
+		Default(googleCloudBaseURL.String()).
```
pintohutch, line 414: @TheSpiritXIII could you update the docs for the [OperatorConfig](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/55d40dccb88e518b8e4e705dedb94191edce12fb/pkg/operator/apis/monitoring/v1/operator_types.go#L152) highlighting this is now the default?

PR #1053, opened 2024-07-03
chore: add mdox formating and binary flag doc autogen

bwplotka:
Rationales:
* Flags are nice to have for OSS users to use, but it's extremely useful for us to double check if we don't accidently break some flag names/drop some flags (API surface). This is to help with https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1001 and co (:
* Mdox can help us with link check in future


TheSpiritXIII (approved):
Great work!

PR #1056, opened 2024-07-08
fix: show warning when cleaning up old resources without delete permissions

TheSpiritXIII:
Using the cleanup flag today causes the operator to crash-looping because we don't include the DELETE RBAC permission by default.

This happened because we used to do all resource creation/update/deletion on the operator but we moved it to YAML. This also makes the cleanup function consistent, as we warn when we can't delete everything else.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -389,8 +389,13 @@ func (o *Operator) cleanupOldResources(ctx context.Context) error {
 		return fmt.Errorf("get collector DaemonSet: %w", err)
 	}
 	if _, ok := ds.Annotations[o.opts.CleanupAnnotKey]; !ok {
-		if err := o.client.Delete(ctx, &ds); client.IgnoreNotFound(err) != nil {
-			return fmt.Errorf("delete collector DaemonSet: %w", err)
+		if err := o.client.Delete(ctx, &ds); err != nil {
+			switch {
+			case apierrors.IsForbidden(err):
+				o.logger.Info("delete collector was not allowed. Please remove it manually", "err", err)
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -404,8 +409,13 @@ func (o *Operator) cleanupOldResources(ctx context.Context) error {
...
 	}
 	if _, ok := deploy.Annotations[o.opts.CleanupAnnotKey]; !ok {
-		if err := o.client.Delete(ctx, &deploy); client.IgnoreNotFound(err) != nil {
-			return fmt.Errorf("delete rule-evaluator Deployment: %w", err)
+		if err := o.client.Delete(ctx, &deploy); err != nil {
+			switch {
+			case apierrors.IsForbidden(err):
+				o.logger.Info("delete rule-evaluator Deployment was not allowed. Please remove it manually")
+			case !apierrors.IsNotFound(err):
+				return fmt.Errorf("delete rule-evaluator Deployment failed: %w", err)
```
bernot-dev, line 417: Question: Doesn't the Not Found error include the type (Deployment) and the name (rule-evaluator)? I think that might be redundant.

pintohutch (approved):
LGTM % Adam's question.

Also, is this an urgent issue? What's preventing us from hitting this on GKE today or catching it in E2E tests, for example?

Can you elaborate in the PR description?

pkg/operator/operator.go
TheSpiritXIII: This is checking for _not_ not found. However, as the error message generally includes the type and name, I updated the wording.

pintohutch (approved):

PR #1059, opened 2024-07-08
feat: allow loading rule-evaluator query options from config

TheSpiritXIII:
~~Pre-req: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1104~~

The following are now settable in the rule-evaluator Prometheus config:

```yaml
google_cloud:
  query:
    project_id: []string
    generator_url: string
    credentials: string
```

When both CLI arguments and configurations are present, the CLI arguments will be the default and the set configurations will override the CLI arguments.

For now, this is only settable at startup time. https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1129 allows setting on runtime.

### DIFF
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"reflect"
 	"runtime"
 	"runtime/debug"
 	"strconv"
@@ -69,11 +70,29 @@ import (
 
 const projectIDVar = "PROJECT_ID"
 
-var googleCloudBaseURL = url.URL{
-	Scheme: "https",
-	Host:   "console.cloud.google.com",
-	Path:   "/monitoring/metrics-explorer",
-}
+var (
+	googleCloudBaseURL = url.URL{
+		Scheme: "https",
+		Host:   "console.cloud.google.com",
+		Path:   "/monitoring/metrics-explorer",
+	}
+
+	queryCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "rule_evaluator_query_requests_total",
+			Help: "A counter for query requests sent to GCM.",
+		},
+		[]string{"code", "method"},
+	)
+	queryHistogram = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "rule_evaluator_query_requests_latency_seconds",
+			Help:    "Histogram of response latency of query requests sent to GCM.",
+			Buckets: prometheus.DefBuckets,
+		},
+		[]string{"code", "method"},
+	)
+)
 
 func main() {
 	logger := log.NewJSONLogger(log.NewSyncWriter(os.Stderr))
@@ -98,6 +117,8 @@ func main() {
 		collectors.NewGoCollector(),
 		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
 		grpc_prometheus.DefaultClientMetrics,
+		queryCounter,
+		queryHistogram,
 	)
 
 	// The rule-evaluator version is identical to the export library version for now, so
@@ -115,7 +136,7 @@ func main() {
 	}
 	opts.SetupFlags(a)
 
-	evaluatorOpts := evaluatorOptions{
+	defaultEvaluatorOpts := evaluatorOptions{
 		TargetURL:     Must(url.Parse(fmt.Sprintf("https://monitoring.googleapis.com/v1/projects/%s/location/global/prometheus", projectIDVar))),
 		ProjectID:     defaultProjectID,
 		DisableAuth:   false,
@@ -123,7 +144,7 @@ func main() {
 		ConfigFile:    "prometheus.yml",
 		QueueCapacity: 10000,
 	}
-	evaluatorOpts.setupFlags(a)
+	defaultEvaluatorOpts.setupFlags(a)
 
 	extraArgs, err := exportsetup.ExtraArgs()
 	if err != nil {
@@ -137,7 +158,7 @@ func main() {
 		os.Exit(2)
 	}
 
-	if err := evaluatorOpts.validate(); err != nil {
+	if err := defaultEvaluatorOpts.validate(); err != nil {
 		_ = level.Error(logger).Log("msg", "invalid command line argument", "err", err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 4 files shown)


cmd/rule-evaluator/main.go
```diff
@@ -542,8 +584,12 @@ func sendAlerts(s *notifier.Manager, projectID string, externalURL *url.URL) rul
 			}
 			if externalURL != nil {
 				if externalURL.String() == googleCloudBaseURL.String() {
-					// If it's a GCM link (default), create the full URL for the alert.
-					a.GeneratorURL = googleCloudLink(projectID, expr, alert.FiredAt, alert.FiredAt.Add(-time.Hour)).String()
+					if projectID == "" {
```
bwplotka, line 587: Should we add a commentary on why project ID can be empty (it's for noop state when no rules are configured).

cmd/rule-evaluator/main.go
```diff
@@ -920,14 +962,72 @@ func newRuleEvaluator(
...
+		v1api, err := newAPI(e.ctx, evaluatorOpts, e.version, func(rt http.RoundTripper) http.RoundTripper {
+			return makeInstrumentedRoundTripper(rt)
+		})
+		if err != nil {
+			return fmt.Errorf("query client: %w", err)
+		}
+		queryFunc := newQueryFunc(e.logger, v1api)
+
+		externalURL := evaluatorOpts.GeneratorURL
+		if externalURL != nil && externalURL.String() == googleCloudBaseURL.String() {
```
bwplotka, line 994: It feels we could DRY this a bit, but not a blocker.

cmd/rule-evaluator/main.go
```diff
@@ -920,14 +962,72 @@ func newRuleEvaluator(
...
+		lastEvaluatorOpts: evaluatorOpts,
+		rerunManager:      false,
 	}
 
 	return &evaluator, nil
 }
 
-func (e *ruleEvaluator) ApplyConfig(cfg *config.Config) error {
+func (e *ruleEvaluator) ApplyConfig(cfg *config.Config, evaluatorOpts *evaluatorOptions) error {
+	if evaluatorOpts != nil && !reflect.DeepEqual(evaluatorOpts, e.lastEvaluatorOpts) {
```
bwplotka, line 982: This is a bit of work that could be replaced by simply detecting change AND os.Exit(0) (with clear log AND metric ideally, AND do this in main ofc).

This has consequence of tricky to debug if the "restart" was actual crash or consequence of valid config changes, but potentially it's fine?

cmd/rule-evaluator/main.go
```diff
@@ -951,10 +1051,26 @@ func (e *ruleEvaluator) Query(ctx context.Context, q string, t time.Time) (promq
 }
 
 func (e *ruleEvaluator) Run() {
-	e.rulesManager.Run()
+	for {
+		// Copy the rule-manager before running, so we don't hold the lock.
+		e.mtx.Lock()
+		oldRuleManager := e.rulesManager
```
bwplotka, line 1057: I will think about perhaps if there's anything to simplify here

cmd/rule-evaluator/main.go
```diff
@@ -855,59 +901,52 @@ func (db *queryAccess) Close() error {
...
+	if externalURL != nil && externalURL.String() == googleCloudBaseURL.String() {
+		// Project ID is empty when the rule-evaluator is instantiated, before config-reloader runs.
+		if projectID == "" {
+			externalURL = nil
+		} else {
+			// If it's a GCM link (default), create the full URL for the alert.
+			externalURL = googleCloudLink(projectID, "", time.Time{}, time.Time{})
+		}
+	}
+	return externalURL
```
bwplotka, line 928: Want to reverse a bit for readability?

```suggestion
        // Project ID is empty when the rule-evaluator is instantiated, before config-reloader runs.
	if externalURL == nil || projectID == ""  {
	     return nil
	}
	
	if externalURL.String() == googleCloudBaseURL.String() {
		// If it's a GCM link (default), create the full URL for the alert.
		externalURL = googleCloudLink(projectID, "", time.Time{}, time.Time{})
	}
	return externalURL
```

cmd/rule-evaluator/main.go
```diff
@@ -855,59 +901,52 @@ func (db *queryAccess) Close() error {
...
 
 type ruleEvaluator struct {
-	queryFunc    rules.QueryFunc
-	rulesManager *rules.Manager
+	queryFunc         rules.QueryFunc
+	rulesManager      *rules.Manager
+	lastEvaluatorOpts *evaluatorOptions
+}
+
+// Returns the URL that points to the rule-evaluator instance. By default, or if using a Google
```
bwplotka, line 915: How this URL points to rule-eval instance? I don't see any logic for this in this function 🤔

cmd/rule-evaluator/main.go
```diff
@@ -542,8 +583,13 @@ func sendAlerts(s *notifier.Manager, projectID string, externalURL *url.URL) rul
 			}
 			if externalURL != nil {
 				if externalURL.String() == googleCloudBaseURL.String() {
```
bwplotka, line 585: Want to rename `externalURL` with `generatorURL` (this is what we get from callers)

cmd/rule-evaluator/main.go
```diff
@@ -542,8 +583,13 @@ func sendAlerts(s *notifier.Manager, projectID string, externalURL *url.URL) rul
...
-					// If it's a GCM link (default), create the full URL for the alert.
-					a.GeneratorURL = googleCloudLink(projectID, expr, alert.FiredAt, alert.FiredAt.Add(-time.Hour)).String()
+					// Project ID is empty when the rule-evaluator is instantiated, before config-reloader runs.
+					if projectID == "" {
+						a.GeneratorURL = ""
+					} else {
+						// If it's a GCM link (default), create the full URL for the alert.
+						a.GeneratorURL = googleCloudLink(projectID, expr, alert.FiredAt, alert.FiredAt.Add(-time.Hour)).String()
+					}
```
bwplotka, line 592: So this `// Project ID is empty when the rule-evaluator is instantiated, before config-reloader runs.` comment is likely wrong. This is `sendAlert` that is unlikely to be called when no rules or alerts are set. Perhaps it's fine to ignore this check here? 

Or... if `googleCloudLink` has empty project ID we could simply set it to "<ProjectID>" so it's clear we don't know, but I am sceptical we would ever hit this here.

AND if you really want this check, maybe this is more readable: 

```suggestion
					if projectID != "" {
						// If it's a GCM link (default), and we have a projectID set, create the full URL for the alert.
						a.GeneratorURL = googleCloudLink(projectID, expr, alert.FiredAt, alert.FiredAt.Add(-time.Hour)).String()
					}
```

bwplotka (approved):
Readability items only, LGTM otherwise!

cmd/rule-evaluator/main.go
TheSpiritXIII: Note: This was split to other PR: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1129

cmd/rule-evaluator/main.go
TheSpiritXIII: Thanks! I'd like to keep this check because you need a project ID for a valid link, so I made the readability change.

cmd/rule-evaluator/main.go
TheSpiritXIII: The user would set it to the rule-eval instance. I clarified the comment. Thanks!

PR #1065, opened 2024-07-11
feat: workload labels

bernot-dev:
Implement go/gmp:expand-target-labels.

### DIFF
--- a/charts/operator/crds/monitoring.googleapis.com_clusterpodmonitorings.yaml
+++ b/charts/operator/crds/monitoring.googleapis.com_clusterpodmonitorings.yaml
@@ -138,8 +138,8 @@ spec:
                       description: |-
                         Relabeling rules for metrics scraped from this endpoint. Relabeling rules that
                         override protected target labels (project_id, location, cluster, namespace, job,
-                        instance, or __address__) are not permitted. The labelmap action is not permitted
-                        in general.
+                        instance, workload_controller, workload_controller_type, or __address__) are not
+                        permitted. The labelmap action is not permitted in general.
                       items:
                         description: RelabelingRule defines a single Prometheus relabeling
                           rule.
--- a/charts/operator/crds/monitoring.googleapis.com_podmonitorings.yaml
+++ b/charts/operator/crds/monitoring.googleapis.com_podmonitorings.yaml
@@ -128,8 +128,8 @@ spec:
                       description: |-
                         Relabeling rules for metrics scraped from this endpoint. Relabeling rules that
                         override protected target labels (project_id, location, cluster, namespace, job,
-                        instance, or __address__) are not permitted. The labelmap action is not permitted
-                        in general.
+                        instance, workload_controller, workload_controller_type, or __address__) are not
+                        permitted. The labelmap action is not permitted in general.
                       items:
                         description: RelabelingRule defines a single Prometheus relabeling
                           rule.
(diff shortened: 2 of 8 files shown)


pkg/operator/operator_config.go
```diff
@@ -646,11 +646,59 @@ func (r *operatorConfigReconciler) makeAlertmanagerConfigs(ctx context.Context,
 		if err != nil {
 			return nil, nil, fmt.Errorf("cannot build regex from service name %q: %w", am.Name, err)
 		}
-		cfg.RelabelConfigs = append(cfg.RelabelConfigs, &relabel.Config{
-			Action:       relabel.Keep,
-			SourceLabels: prommodel.LabelNames{"__meta_kubernetes_endpoints_name"},
-			Regex:        svcNameRE,
-		})
+		cfg.RelabelConfigs = append(cfg.RelabelConfigs,
```
bwplotka, line 649: Should we ensure there is a dynamic (e.g. config based) switch off for this? e.g. when we block some customer with this?

bwplotka (changes requested):
Relabelling rules looks generally good, but did you intend to change AlertingConfig? Relabeling in Alerting Config only adds labels to alerts, not metrics. Wonder if you meant to do this scrape config section instead? Or am I missing something? 🤔

Tests would be nice for metrics too (unless you meant to do this for alerts).

pkg/operator/operator_config.go
bernot-dev: The design rejected putting this change behind a feature flag, and also suggests that the creation of a feature flag would increase confusion and the maintenance burden for TargetLabels.

An MSA is planned to notify users of this upcoming change and allow them to adjust accordingly.

bernot-dev:
> did you intend to change AlertingConfig?

Nope. That was my mistake. Should be fixed now.

pkg/operator/operator_config.go
TheSpiritXIII: I believe the idea is that users before this change are unimpacted since there may be customers already using a label named `workload_controller`. The `PodMonitoring.targetLabels` field is essentially the switch. This PR is missing that.

@bernot-dev let's update the Defaulter, such that if the customer created a new PodMonitoring (e.g. did not set a `targetLabels` value), we add `workload_controller` in the list. Then, only add these additional configurations if it is specified (see `relabelingsForMetadata`).

pkg/operator/apis/monitoring/v1/operator_config.go
```diff
@@ -70,6 +70,51 @@ func (c *CollectionSpec) ScrapeConfigs() ([]*promconfig.ScrapeConfig, error) {
 			SourceLabels: prommodel.LabelNames{"__meta_kubernetes_node_name"},
 			TargetLabel:  "node",
 		},
+		// Capture controller name
```
TheSpiritXIII, line 73: Nice! As per the comment I left on Bartek's thread, let's move this and make it conditional (and therefore backwards-compatible).

TheSpiritXIII (changes requested):

bwplotka (approved):
Love it, thanks! I like this to be implicit for us, so we can iterate and own this (flexibility).

Can we add test for podmonitoring config that filters this out (to check if it's possible for users, I think the agreement was that we want it).

bernot-dev:
> FYI - we're still awaiting final word on if these labels are sufficient cc @xichen2020

@pintohutch Is this blocking?

pintohutch:
I don't think it's blocking this PR per se. We may want to wait final labels before a release though

TheSpiritXIII (approved):

PR #1074, opened 2024-07-16
feat: update alertmanager flags at runtime

TheSpiritXIII:
Uses the new config settings instead of CLI flags for alertmanager, and makes StatefulSet UPDATE permissions optional.


bwplotka (approved):
Thanks! (co-reviewing with @TheSpiritXIII )

PR #1075, opened 2024-07-16
feat: add config-reloader readiness

TheSpiritXIII:


pintohutch:
Can you fill out the description and describe what this change is doing and the motivation for it?

pintohutch:
@TheSpiritXIII - IIUC our discussion around supporting "empty" mandatory flag/config will make this change unnecessary. Is that your understanding as well?

TheSpiritXIII:
@pintohutch yes this change becomes non-critical. This PR may still provide a slightly better signal over when a workload is "ready" than we have today, but again, let's not prioritize it unless we need to.

bwplotka:
LGTM, but we need to resolve the conflicts. Thanks!

PR #1076, opened 2024-07-16
 feat: always output config reloader file even if application is not ready

TheSpiritXIII:
The problem is that the config-reloader never runs if the Prometheus instance starts crash-looping. This can create a small delay when the Prometheus instance eventually recovers and starts scraping metrics: the Prometheus instance will startup with the old configuration, and only then the config-reloader will update it.

If we move flags to the configuration-side, then being able to update the configuration before Prometheus starts up is critical. Without this, this PR fails: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1059


cmd/config-reloader/main.go
```diff
@@ -73,6 +74,35 @@ func main() {
...
+		return reloader.New(
+			logger,
+			reg,
+			&reloader.Options{
+				ReloadURL:     reloadURL,
+				CfgFile:       *configFile,
+				CfgOutputFile: *configFileOutput,
+				CfgDirs:       cfgDirs,
+				WatchedDirs:   watchedDirs,
+				// There are some reliability issues with fsnotify picking up file changes.
```
bernot-dev, line 95: fsnotify works fine if you set it up to [add a new watch every time it detects a remove](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/certwatcher/certwatcher.go#L184) (which is how edits end up showing up in Kubernetes volumes).

That said, polling is probably fine, too.

bernot-dev: Suggestion: Remove this part of the comment.

pintohutch:
I am a little confused, given configuration reloading can happen at any time duration of the Prometheus runtime. We also have the initContainer ensuring an empty config is present before Prometheus starts to prevent transient crashlooping.

Can you elaborate on why this is necessary or why the other PR is failing without this?

TheSpiritXIII:
@pintohutch thanks, I expanded on the PR description. Please let me know if it doesn't make sense!

cmd/config-reloader/main.go
```diff
@@ -110,6 +136,26 @@ func main() {
...
+	if err != nil {
+		//nolint:errcheck
+		level.Error(logger).Log("msg", "parsing listen URL failed", "err", err)
+		os.Exit(1)
+	}
+	// Don't pass a metrics registry, because config-reloader panics when we create our second.
+	tempReloader := newReloader(nil, &url.URL{
+		Scheme: "http",
+		// Since Prometheus is not ready, we won't be able to hit the reload URL so hit itself.
+		Host: listenAddr.String(),
```
bwplotka, line 150: Yea, I would advocate to stop hacking like that (: Bit too much. 

All the reloader code is upstream. I technically maintain it with Thanos team. Either we fork/vendor that code or we fix there, this is too hacky.

On top of that literally nothing technically fails (as is - stops working) if you just run normal reloader as we have now (without creating new) even if Prometheus is down. The [reloading will retry, increment metric and log error.](https://github.com/thanos-io/thanos/blob/main/pkg/reloader/reloader.go#L534) - expected if Prometheus is slowly starting.

To me all we need to do is either:
1) Actually do what has been proposed 1y+ ago, [just remove readiness check](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/526). Same arguments why it's not necessary apply now.
2) We could keep readiness check but start reloader early in the background if we want readiness crash (I don't think that's useful)
3) Change reloader upstream, so there is a lib function that only generates config, easy.
4) Embed config-reloader code to rule-eval.

bwplotka (changes requested):
Thanks a lot! 

I think the description is out-dated. We decided to reload by restarting internal rule manager, so this PR is not strictly blocking security hardening. HOWEVER I fully agree we should do this for other reasons (logic and faster scrape/recording).

I disagree with the way it's done here, I proposed some alternatives (: 

cc @dashpole for knowledge sharing 👍🏽 and @bernot-dev for opinions (:

Skipping this for 0.14 (if my understanding is correct, we don't need this)

PR #1078, opened 2024-07-17
refactor: remove Deployment update and patch permission

TheSpiritXIII:
Pre-req: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1105

### DIFF
--- a/charts/operator/templates/role.yaml
+++ b/charts/operator/templates/role.yaml
@@ -82,12 +82,8 @@ rules:
 - resources:
   - deployments
   apiGroups: ["apps"]
-  verbs: ["list", "watch"]
-- resources:
-  - deployments
-  apiGroups: ["apps"]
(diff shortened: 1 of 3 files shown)


vinayakankugoyal (approved):

charts/operator/templates/role.yaml
```diff
@@ -79,15 +79,11 @@ rules:
   apiGroups: ["apps"]
   resourceNames: ["collector"]
   verbs: ["get", "list", "watch", "patch", "update"]
-- resources:
-  - deployments
-  apiGroups: ["apps"]
-  verbs: ["list", "watch"]
```
bwplotka, line 85: oh wow, that was broad.. wonder why. Accident?

bwplotka (approved):
Let's go 💪🏽

PR #1087, opened 2024-07-23
chore: update to Prometheus v2.53.1 image

TheSpiritXIII:


bwplotka:
I think this is no longer needed?

PR #1091, opened 2024-07-24
test: support kind tests with images without tags

TheSpiritXIII:
Current sed regex breaks if there's no tag. We're moving to tag-less images, so this fixes that problem.


pintohutch (approved):
Nice catch!

PR #1092, opened 2024-07-25
fix: change data source syncer period to 10 minutes

lyanco:
fix: change data source syncer period to 10 minutes

Every 30 minutes meant that there could be at most 1 network timeout before losing connection. Running every 10 minutes instead makes it much less likely that a networking hiccup will cause Grafana to fail.

### DIFF
--- a/charts/datasource-syncer/templates/cronjob.yaml
+++ b/charts/datasource-syncer/templates/cronjob.yaml
@@ -18,7 +18,7 @@ kind: CronJob
 metadata:
   name: datasource-syncer
 spec:
-  schedule: "*/30 * * * *" # Run once every 30 minutes, must run at least once an hour.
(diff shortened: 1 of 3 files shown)


pintohutch:
@lyanco - looks like the presubmit is failing.

I believe this is due to the recent `helm` templating migration @bernot-dev introduced. I believe you want to make the change in https://github.com/GoogleCloudPlatform/prometheus-engine/blob/1cb4afb3a875c1dbdb764d4c5d986a6ab773b27b/charts/datasource-syncer/templates/cronjob.yaml#L21, and run `make regen`. Adam can keep me honest.

bernot-dev:
Interesting. Yes, looks like Bartek added datasource-syncer to the templating. I fixed this issue and pushed the updated version. Hopefully presubmits will pass this time.

maxamins (approved):
LGTM. In the future we can also have datasource-syncer retry failed requests.

PR #1102, opened 2024-08-06
refactor: add setup opts utility

TheSpiritXIII:
Makes it easier for clients to use the setup code.


pintohutch (approved):
Cool - thanks!

PR #1103, opened 2024-08-07
Revert "feat: generate max-throughput collector manifest"

bwplotka:
This reverts commit 15c6f7246367ee575517b6d2cf9bbe43abe820df, due to regression on GKE autopilot (b/357897990).

Let's fix and do in 0.14


pintohutch (approved):
Thanks - agreed we should approach this differently.

PR #1104, opened 2024-08-08
feat: allow loading rule-evaluator export options from config

TheSpiritXIII:
This is the rule-evaluator equivalent of https://github.com/GoogleCloudPlatform/prometheus/pull/180

The following are now settable in the rule-evaluator Prometheus config:

```yaml
google_cloud:
  export:
    match: []string
    compression: string
    credentials: string
```

When both CLI arguments and configurations are present, the CLI arguments will be the default and the set configurations will override the CLI arguments.

### DIFF
--- a/pkg/operator/operator_config.go
+++ b/pkg/operator/operator_config.go
@@ -287,6 +287,31 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorConfig(ctx context.Context
 	return secretData, nil
 }
 
+type RuleEvaluatorConfig struct {
+	promconfig.Config `yaml:",inline"`
+
+	// Google Cloud configuration. Matches our fork's configuration.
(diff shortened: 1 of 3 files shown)


pkg/operator/operator_config.go
```diff
@@ -286,6 +286,31 @@ func (r *operatorConfigReconciler) ensureRuleEvaluatorConfig(ctx context.Context
 	return secretData, nil
 }
 
+type RuleEvaluatorConfig struct {
+	promconfig.Config `yaml:",inline"`
+
+	// Google Cloud configuration. Matches our fork's configuration.
+	GoogleCloud GoogleCloudConfig `yaml:"google_cloud,omitempty"`
```
bwplotka, line 293: Can we import Prometheus type for the same?

bwplotka (approved):
LGTM, minor nits, we could merge anyway.

PR #1114, opened 2024-08-19
chore: add image tags to manifests

### LINKED ISSUE #1081: Release tags for docker images
Hi,

Manifests in the latest release branch (`v0.12.0`) don't use image tags with docker images and the corresponding tags are missing in the docker registry (the latest one available is `v0.11.0-gke.3`, it seems).

It would be great to tag the corresponding images with `v0.12.0` -- this would considerably simplify the upgrade process.

Thank you

pintohutch:
Our tagging pipeline was non-functional for a period, leading us to relying on using the image shas instead of tags.

Now that the pipeline is functional again, we can provide convenient image tags for reference and installations.

Fixes #1081

### DIFF
--- a/charts/operator/values.yaml
+++ b/charts/operator/values.yaml
@@ -19,24 +19,23 @@ namespace:
   system: gmp-system
 images:
   alertmanager:
-    image: gke.gcr.io/prometheus-engine/alertmanager@sha256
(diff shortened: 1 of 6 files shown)


TheSpiritXIII (approved):
Haven't reviewed if tags and SHAs match but looks great! :shipit:

dashpole (approved):

PR #1126, opened 2024-08-21
feat: add samples sent error counter

pintohutch:
We previously only logged an error when this happened, which can be hard to debug. Particularly in cases where we are writing to more than one project.

This change introduces a counter `gcm_export_samples_sent_errors_total` that is incremented every time a send call to GCM returns an error. We add the `project_id` as a label to help better inform the destination project that is experiencing the issue.

On a given collector or rule-evaluator, the cardinality of `project_id` values should be relatively small, O(100) in the more extreme cases.

### DIFF
--- a/pkg/export/export.go
+++ b/pkg/export/export.go
@@ -70,7 +70,7 @@ var (
 		Name: "gcm_export_samples_sent_total",
 		Help: "Number of exported samples sent to GCM.",
 	})
-	samplesSentErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
+	samplesSendErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "gcm_export_samples_send_errors_total",
 		Help: "Number of errors encountered while sending samples to GCM",
 	}, []string{"project_id"})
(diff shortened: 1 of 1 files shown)


pintohutch:
cc @lyanco - this may be a nice self-observability case for customers having issues.

pkg/export/export.go
```diff
@@ -70,6 +70,10 @@ var (
 		Name: "gcm_export_samples_sent_total",
 		Help: "Number of exported samples sent to GCM.",
 	})
+	samplesSentErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "gcm_export_samples_sent_errors_total",
```
dashpole, line 74:
```suggestion
		Name: "gcm_export_samples_send_errors_total",
```

First, this metric should not be compared with gcm_export_samples_sent_total, since one is samples, and the other is requests.  "send errors" is also a more reasonable unit.

dashpole (approved):

pkg/export/export.go
```diff
@@ -70,6 +70,10 @@ var (
 		Name: "gcm_export_samples_sent_total",
 		Help: "Number of exported samples sent to GCM.",
 	})
+	samplesSentErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
```
bwplotka, line 73: Let's not forget to register that metric

bwplotka (approved):
LGTM, thanks, just one important thing to do.

pkg/export/export.go
pintohutch: Good catch - done!

Issue #1123, opened 2024-08-20
Add operator to self-pod-monitoring


bwplotka:
We only have collector & rule-eval here: 
https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/examples/self-pod-monitoring.yaml

bwplotka:
Fixed by https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1127

Issue #793, opened 2024-01-29
[Design] Dynamic resource usage for GMP operator


pintohutch:
Can we find ways to avoid OOM crashes in the gmp-operator? Maybe using a [VPA](https://gist.github.com/pintohutch/65bc578f1ca7f9d07ad44ff944168bb6)?

Acceptance criteria:
- Proposal with design and trade-offs

clearclaw:
This weekend `gmp-operator` started crashlooping with, "no endpoints available for service gmp-operator".  While there were no OOMKill events, that turned out to be the cause.  Putting in a VPA addressed.

[Previous issue with rules-evaluator](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/971)

- **ALL** deploys were broken due to gmp-operator's webhook being dead. 
- Really bad visibility on the OOMKill.  (I don't know why scheduler events weren't propagating, but they weren't)
- As with rules-evaluator: VPA FTW.

I see that the [rules-evaluator VPA](https://github.com/GoogleCloudPlatform/prometheus-engine/issues/975) hasn't moved either.  I can live lacking VPAs, but having the webhook fail closed for gmp-operator was excruciating.

pintohutch:
Thanks for reporting @clearclaw - I apologize for the frustration.

With the operator being such a mission-critical binary in managed-collection, we can also explore failing open for our webhooks. Would that have helped in this situation over a VPA specifically?

clearclaw:
Yes, in that deploys would have worked despite the failure.  Being unable
to eg  ship a key security fix due to GMP indigestion is not right.

The lack of built-in observability/alerting around GMP remains a concern
(gmp-operator or rules-evaluator previously here, but there's more
components).  If the observability stack fails, all the alarms should be
going off.  It would be really nice to see/have default Rules objects for
all of GMP.
-- JCL

clearclaw:
Supplemental in case useful: 
![image](https://github.com/user-attachments/assets/9b123670-adb1-433c-955d-b7df6c317974)

bwplotka:
+1, some way of handling this would be important, let's prioritize it.

VPA/limits is one thing, but perhaps additionally it would be to prioritize the code optimization phase, to check for low hanging fruits and start to queue things up (trading latency vs memory).

> Additionally, the lack of default observability/alerting around GMP
remains a concern (gmp-operator or rules-evaluator previously here,
but there's more components).  If the observability stack fails, all the
[...]

PR #1127, opened 2024-08-22
feat: add more example self-podmonitorings

pintohutch:
The gmp-operator exposes Prometheus metrics that can be helpful to debug issues with managed-collection. This is also the case with the managed alertmanager.

We have examples of how to scrape metrics from other components, but not the operator or the alertmanager.

So here we provide examples to supplement the self-monitoring exporter documented in
https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/prometheus?hl=en.

Partly addresses #793.

### DIFF
--- a/examples/self-pod-monitoring.yaml
+++ b/examples/self-pod-monitoring.yaml
@@ -52,3 +52,37 @@ spec:
     interval: 30s
   - port: cfg-rel-metrics
     interval: 30s
+---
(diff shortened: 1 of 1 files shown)


bwplotka:
Fixes https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1123

bwplotka (approved):
Amazing, thanks!

PR #1129, opened 2024-08-23
feat: hot-reloader rule-evaluator query options from config

TheSpiritXIII:
Pre-req: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1059

### DIFF
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -29,6 +29,7 @@ import (
 	"runtime/debug"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -521,7 +522,7 @@ func (opts *evaluatorOptions) validate() error {
 	return nil
 }
 
-func newAPI(ctx context.Context, opts *evaluatorOptions, version string, roundTripperFunc func(http.RoundTripper) http.RoundTripper) (v1.API, error) {
+func newAPI(ctx context.Context, opts *evaluatorOptions, version string) (v1.API, error) {
 	clientOpts := []option.ClientOption{
 		option.WithScopes("https://www.googleapis.com/auth/monitoring.read"),
 		option.WithUserAgent(fmt.Sprintf("rule-evaluator/%s", version)),
@@ -539,7 +540,8 @@ func newAPI(ctx context.Context, opts *evaluatorOptions, version string, roundTr
 	if err != nil {
 		return nil, err
 	}
-	roundTripper := roundTripperFunc(transport)
+	roundTripper := promhttp.InstrumentRoundTripperCounter(queryCounter,
+		promhttp.InstrumentRoundTripperDuration(queryHistogram, transport))
 	client, err := api.NewClient(api.Config{
 		Address:      strings.ReplaceAll(opts.TargetURL.String(), projectIDVar, opts.ProjectID),
 		RoundTripper: roundTripper,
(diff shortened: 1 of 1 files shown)


cmd/rule-evaluator/main.go
```diff
@@ -959,17 +969,60 @@ func newRuleEvaluator(
...
 	return &evaluator, nil
 }
 
 func (e *ruleEvaluator) ApplyConfig(cfg *config.Config, evaluatorOpts *evaluatorOptions) error {
 	if evaluatorOpts != nil && !reflect.DeepEqual(evaluatorOpts, e.lastEvaluatorOpts) {
-		return fmt.Errorf("rule-config can only be changed at startup time")
+		e.lastEvaluatorOpts = evaluatorOpts
+
+		v1api, err := newAPI(e.ctx, evaluatorOpts, e.version, func(rt http.RoundTripper) http.RoundTripper {
+			return makeInstrumentedRoundTripper(rt)
```
bwplotka, line 993: This is now literally a one-liner. Want to inline and remove this function?

cmd/rule-evaluator/main.go
```diff
@@ -959,17 +969,60 @@ func newRuleEvaluator(
...
+			Metrics:    e.rulesMetrics,
+		})
+
+		// Set new rule-manager and flag before stopping, so we can rerun with the new one.
+		e.mtx.Lock()
+		oldRuleManager := e.rulesManager
+		e.rulesManager = rulesManager
+		e.rerunManager = true
+		oldRuleManager.Stop()
+		e.queryFunc = queryFunc
```
bwplotka, line 1019: I believe the user of `queryFunc` should also be behind the mutex in this scenario.

```
func (e *ruleEvaluator) Query(ctx context.Context, q string, t time.Time) (promql.Vector, error) {
	return e.queryFunc(ctx, q, t)
}
```

cmd/rule-evaluator/main.go
```diff
@@ -959,17 +969,60 @@ func newRuleEvaluator(
...
+			Logger:     e.logger,
+			NotifyFunc: sendAlerts(e.notifierManager, evaluatorOpts.ProjectID, evaluatorOpts.GeneratorURL),
+			Metrics:    e.rulesMetrics,
+		})
+
+		// Set new rule-manager and flag before stopping, so we can rerun with the new one.
+		e.mtx.Lock()
+		oldRuleManager := e.rulesManager
+		e.rulesManager = rulesManager
+		e.rerunManager = true
```
bwplotka, line 1017: I propose we replace this bool with a simple semantics of `Stop` nilling the `ruleManager`, wdyt?

```
func (e *ruleEvaluator) Stop() {
	e.mtx.Lock()
	defer e.mtx.Unlock()
	e.rulesManager.Stop()
        e.rulesManager = nil
}
```

This allows us to simplify Run:

```
func (e *ruleEvaluator) Run() {
	for {
		e.mtx.Lock()
		curr := e.rulesManager
                	e.mtx.Unlock()
                 // If rulesManager was nilled, it means shutdown process, otherwise it's an config update requiring 
                 // ruleManager to be restarted.
                 if curr == nil {
                     break
                 }
		curr.Run()
	}
}
```

bwplotka (approved):
Nice, thanks! As agreed offline, let's do the runtime magic happen. 

I propose a small simplification, otherwise LGTM! 💪🏽

PR #1141, opened 2024-09-12
chore: remove unused helper

damemi:
Remove `setContainerExtraArgs` helper as it is unused and causing lint failures


pintohutch:
Ah yes - this is no longer needed with the security-hardening changes. LGTM!

pintohutch (approved):

PR #1142, opened 2024-09-12
chore: update Alertmanager 0.27.0

damemi:


TheSpiritXIII (approved):
Approving, but we cannot merge until the image is in production: https://pantheon.corp.google.com/gcr/images/gke-release/global/prometheus-engine/alertmanager?mods=prod_coliseum

damemi:
@bwplotka we can close this now right?

PR #1146, opened 2024-09-17
Use protobuf encoding for core K8s APIs in Prometheus Engine

tosi3k:
For core K8s API objects like Pods, Nodes, etc., we can use protobuf encoding which, compared to the default JSON encoding, reduces CPU consumption related to (de)serialization, reduces overall latency of the API calls, reduces memory footprint and the work performed by the GC and results in quicker propagation of objects to event handlers of shared informers.

Core system components of K8s default their serialization method to protobuf for 8 years already: https://github.com/kubernetes/kubernetes/pull/25738.

Some benchmarks comparing JSON vs. protobuf showcasing how the latter data format (de)serializes faster and uses less memory:

- https://medium.com/@akresling/go-benchmark-json-v-protobuf-4ec3c62ec8d4
- https://www.codingexplorations.com/blog/performance-comparison-protobuf-marshaling-vs-json-marshaling-in-go
- https://medium.com/@david_turner/protocol-buffers-just-how-fast-are-they-9ec19ea580db

### DIFF
--- a/e2e/kube/pod.go
+++ b/e2e/kube/pod.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -31,6 +32,8 @@ import (
 
 // PodLogs returns the logs of the pod with the given name.
 func PodLogs(ctx context.Context, restConfig *rest.Config, namespace, name, container string) (string, error) {
+	coreConfig := rest.CopyConfig(restConfig)
+	coreConfig.ContentType = runtime.ContentTypeProtobuf
 	clientSet, err := kubernetes.NewForConfig(restConfig)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


e2e/kube/pod.go
```diff
@@ -24,13 +24,16 @@ import (
...
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // PodLogs returns the logs of the pod with the given name.
 func PodLogs(ctx context.Context, restConfig *rest.Config, namespace, name, container string) (string, error) {
+	coreConfig := rest.CopyConfig(restConfig)
```
bwplotka, line 35: do we need to copy?

bwplotka (approved):
Quick question about copying, but it looks ok. Curious about potential problems we could find. Is there a case where potential non-GKE k8s would ONLY support JSON and this will fail or client is able to fallback?

thanks!

e2e/kube/pod.go
tosi3k: I think it's safer - without a copy, we would change `restConfig` (passed by pointer here) that might be passed somewhere later to a dynamic clientset and CRs do not support protobuf encoding, unfortunately.

tosi3k:
>Is there a case where potential non-GKE k8s would ONLY support JSON and this will fail or client is able to fallback?

Only in K8s older than 1.3 in which protobuf support was added. It's embedded in the server and can't be turned off.

bwplotka:
Thanks!

PR #1148, opened 2024-09-20
chore: prepare v0.14.0 release candidate

bwplotka:
(also bumped DS go version)


bwplotka:
Manually built images and deployed to GKE std, shown no issue (details in b/367885490).

TheSpiritXIII (approved):
Looks good. I would just double check the alertmanager e2e test after the new image is built. Thanks!

PR #692, opened 2023-12-04
fix operator role for alertmanager externalURL

pintohutch:
also bump the version and images

Note: I had to add the get/list/watch abilities to the ClusterRole instead of the Role. This is obviously less preferable, however the controller-runtime library errors if I add those to the role:
```
E1204 22:39:49.710415       1 reflector.go:140] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:262: Failed to watch *v1.StatefulSet: failed to list *v1.StatefulSet: statefulsets.apps is forbidden: User "system:serviceaccount:gmp-system:operator" cannot list resource "statefulsets" in API group "apps" at the cluster scope
```

PR #1154, opened 2024-09-24
chore(operator): remove ensure AM and rule-eval resource

bwplotka:
After security hardening, those serve no purpose at the moment as AFAIK. I also removed watches for those which is not needed.

I noticed this thanks to `cl/677695994` discussion and when I was exploring past regressions in related flows https://github.com/GoogleCloudPlatform/prometheus-engine/pull/692


charts/operator/templates/role.yaml
```diff
@@ -83,7 +83,7 @@ rules:
   - deployments
   apiGroups: ["apps"]
   resourceNames: ["rule-evaluator"]
-  verbs: ["get", "list", "watch"]
+  verbs: ["get", "list"]
```
TheSpiritXIII, line 86: I think we should be able to get rid of them all? Any other places/reason we look them up?

TheSpiritXIII (approved):

PR #1159, opened 2024-09-27
feat: workload labels

bernot-dev:
Reimplement #1065 with new requirements, as revised in go/gmp:expand-target-labels.

### DIFF
--- a/pkg/operator/apis/monitoring/v1/pod_config.go
+++ b/pkg/operator/apis/monitoring/v1/pod_config.go
@@ -30,21 +30,88 @@ import (
 	"github.com/prometheus/prometheus/model/relabel"
 )
 
+const (
+	labelCluster   = "cluster"
+	labelLocation  = "location"
+	labelProjectID = "project_id"
+
+	labelContainer              = "container"
+	labelNamespace              = "namespace"
+	labelNode                   = "node"
+	labelPod                    = "pod"
+	labelTopLevelControllerName = "top_level_controller_name"
+	labelTopLevelControllerType = "top_level_controller_type"
+)
+
 var (
 	allowedClusterPodMonitoringLabel = map[string]bool{
-		"container": true,
-		"namespace": true,
-		"node":      true,
-		"pod":       true,
+		labelContainer:              true,
+		labelNamespace:              true,
+		labelNode:                   true,
+		labelPod:                    true,
+		labelTopLevelControllerName: true,
+		labelTopLevelControllerType: true,
 	}
 	allowedClusterPodMonitoringLabels = slices.Sorted(maps.Keys(allowedClusterPodMonitoringLabel))
 
 	allowedPodMonitoringLabel = map[string]bool{
-		"container": true,
-		"node":      true,
-		"pod":       true,
+		labelContainer:              true,
+		labelNode:                   true,
+		labelPod:                    true,
+		labelTopLevelControllerName: true,
+		labelTopLevelControllerType: true,
 	}
 	allowedPodMonitoringLabels = slices.Sorted(maps.Keys(allowedClusterPodMonitoringLabel))
+
+	topLevelControllerNameRules = []*relabel.Config{
+		// First, capture the controller name from the pod manifest.
+		{
+			Action:       relabel.Replace,
+			SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_controller_name"},
+			TargetLabel:  labelTopLevelControllerName,
+		},
+		// If the controller kind is a ReplicaSet and it has a pod template hash, it belongs to a deployment.
+		// The name of the deployment is the name of the ReplicaSet with the hash truncated.
+		{
+			Action:       relabel.Replace,
+			SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_controller_kind", "__meta_kubernetes_pod_labelpresent_pod_template_hash", "__meta_kubernetes_pod_controller_name"},
+			Regex:        relabel.MustNewRegexp("ReplicaSet;true;(.+)-[a-z0-9]+"),
+			TargetLabel:  labelTopLevelControllerName,
+		},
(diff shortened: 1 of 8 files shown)


dashpole (commented):
looks reasonable to me, and i'll let bartek approve.  I assume there no good way to test this against actual k8s objects?

pkg/operator/apis/monitoring/v1/pod_config.go
```diff
@@ -382,6 +386,76 @@ func relabelingsForMetadata(keys map[string]struct{}) (res []*relabel.Config) {
 			TargetLabel:  "node",
 		})
 	}
+	if _, ok := keys["top_level_controller_name"]; ok {
+		res = append(res,
+			// Default to pod name
```
bwplotka, line 391: nit: Can we use full sentence and trailing `.` for comments in this file, for clarity? thanks!

pkg/operator/apis/monitoring/v1/pod_config.go
```diff
@@ -32,17 +32,21 @@ import (
...
 	allowedClusterPodMonitoringLabel = map[string]bool{
-		"container": true,
-		"namespace": true,
-		"node":      true,
-		"pod":       true,
+		"container":                 true,
+		"namespace":                 true,
+		"node":                      true,
+		"pod":                       true,
+		"top_level_controller_name": true,
```
bwplotka, line 39: Do you mind switching to constants in this file? Bit of work, but avoid typos 🙈

pkg/operator/apis/monitoring/v1/pod_config.go
```diff
@@ -382,6 +386,76 @@ func relabelingsForMetadata(keys map[string]struct{}) (res []*relabel.Config) {
 			TargetLabel:  "node",
 		})
 	}
+	if _, ok := keys["top_level_controller_name"]; ok {
```
bwplotka, line 389: I would say this deserves a simple unit test that puts what we expect to see in those `__meta` labels and do the resulting relabelling. We can then track regressions or weird __meta label states over time. WDYT? (:

bwplotka (commented):
Thanks! This is great and super important work.

1. Added some nits and unit test suggestions.
2. I checked the past discussions, and there is one more thing we noticed:
   * Is it worth to mention those new labels in [`TargetLabels.Metadata`](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/b1a752822b70eaf672ff69a961dbb441d5f7a749/pkg/operator/apis/monitoring/v1/pod_types.go#L304) for both cluster and pod monitorings?
3. Did you manage to self-deploy GMP with operator from this PR, on GKE std cluster? It would be epic to have at least basics like daemonset, deployment and statefulset workloads and PodMonitorings to see if the labels appear correctly? Would be epic to do this, given no e2e for this (the alternative is to write an e2e test, that would be probably even better).

Otherwise perfect. I would love to propose same labels to Prometheus SD one day (:

pkg/operator/apis/monitoring/v1/pod_config.go
```diff
@@ -382,6 +386,76 @@ func relabelingsForMetadata(keys map[string]struct{}) (res []*relabel.Config) {
...
+			&relabel.Config{
+				Action:       relabel.Replace,
+				SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_name"},
+				TargetLabel:  "top_level_controller_name",
+			},
+			// For non-ReplicaSet and non-Job controllers
+			&relabel.Config{
+				Action:       relabel.Replace,
+				SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_controller_name"},
+				Regex:        relabel.MustNewRegexp("(.+)"),
```
bwplotka, line 401: Do we need this (and `Replacement`) in this entry?

That seems to be [a Prometheus default anyway](https://github.com/prometheus/prometheus/blob/main/model/relabel/relabel.go#L32)?

pkg/operator/apis/monitoring/v1/pod_config.go
```diff
@@ -382,6 +386,76 @@ func relabelingsForMetadata(keys map[string]struct{}) (res []*relabel.Config) {
...
+				SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_name"},
+				Regex:        relabel.MustNewRegexp("(.*)"),
+				TargetLabel:  "top_level_controller_type",
+				Replacement:  "Pod",
+			},
+			// Capture controller type
+			&relabel.Config{
+				Action:       relabel.Replace,
+				SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_controller_kind"},
+				Regex:        relabel.MustNewRegexp("(.+)"),
```
bwplotka, line 437: ditto

pkg/operator/apis/monitoring/v1/pod_config.go
```diff
@@ -382,6 +386,76 @@ func relabelingsForMetadata(keys map[string]struct{}) (res []*relabel.Config) {
...
+	}
+	if _, ok := keys["top_level_controller_type"]; ok {
+		res = append(res,
+			// Default to Pod
+			&relabel.Config{
+				Action:       relabel.Replace,
+				SourceLabels: prommodel.LabelNames{"__meta_kubernetes_pod_name"},
+				Regex:        relabel.MustNewRegexp("(.*)"),
+				TargetLabel:  "top_level_controller_type",
+				Replacement:  "Pod",
```
bwplotka, line 431: Curious, why `pod` and not `unknown` (clear distinction that we can't / have bug in the workload discovery?) . Not a bad choice, just curious of pros & cons, can't see this explained in the design.

Same for controller_name 🤔

bwplotka (approved):
Wow, addressed so quickly! Amazing, LGTM, thanks 💪🏽 

Unit test would be nice for relabel again, but not strictly now in this PR!

bwplotka:
Nice, thanks!
 
image

PR #1161, opened 2024-10-03
feat: add /api/v1/alerts endpoint

yama6a:
Adding a prometheus-compatible API endpoint to rule-evaluator: /api/v1/alerts

### DIFF
--- a/cmd/rule-evaluator/internal/alerts.go
+++ b/cmd/rule-evaluator/internal/alerts.go
@@ -1,0 +1,38 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import (
+	"net/http"
+
+	"github.com/prometheus/prometheus/rules"
+	apiv1 "github.com/prometheus/prometheus/web/api/v1"
+)
+
+type alertsEndpointResponse struct {
+	Alerts []*apiv1.Alert `json:"alerts"`
+}
+
+func (api *API) HandleAlertsEndpoint(w http.ResponseWriter, _ *http.Request) {
+	activeAlerts := []*rules.Alert{}
+	for _, rule := range api.rulesManager.AlertingRules() {
+		activeAlerts = append(activeAlerts, rule.ActiveAlerts()...)
+	}
+
+	alertsResponse := alertsEndpointResponse{
+		Alerts: alertsToAPIAlerts(activeAlerts),
+	}
+	api.writeSuccessResponse(w, http.StatusOK, "/api/v1/alerts", alertsResponse)
+}
--- a/cmd/rule-evaluator/internal/alerts_test.go
+++ b/cmd/rule-evaluator/internal/alerts_test.go
@@ -1,0 +1,62 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/prometheus/prometheus/rules"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAPI_HandleAlertsEndpoint(t *testing.T) {
+	t.Parallel()
+
+	api := &API{
+		rulesManager: RuleGroupsRetrieverMock{
+			AlertingRulesFunc: func() []*rules.AlertingRule {
+				return []*rules.AlertingRule{
+					rules.NewAlertingRule("test-alert-1", &parser.NumberLiteral{Val: 33}, time.Hour, time.Hour*4, []labels.Label{{Name: "instance", Value: "localhost:9090"}}, []labels.Label{{Name: "summary", Value: "Test alert 1"}, {Name: "description", Value: "This is a test alert"}}, nil, "", false, log.NewNopLogger()),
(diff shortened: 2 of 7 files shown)


penekk:
Excellent @yama6a 

Would this finally be enough to pull the rules into Grafana (from the evaluator), like it is with vanilla prometheus?

yama6a:
> Excellent @yama6a
> 
> Would this finally be enough to pull the rules into Grafana (from the evaluator), like it is with vanilla prometheus?

Probably yes-ish. 

For now, the endpoints will only be available on the `rules-evaluator` pod, not the `prometheus-frontend` pod, the latter of which you will probably use as a Prometheus-Datasource at the moment (running promQL queries, and fetching metrics/labels, etc).

A temporary hack could be to add the rule-evaluator pod as a separate Prometheus-Datasource in Grafana (once this is released), but I hope to make a PR for a proper solution soon-ish.

Currently, the frontend will blindly proxy all calls to `/api/*` to Google-Monitoring's prometheus endpoint, which doesn't work for these two new endpoints. I'm working on fixing that, and hope to have a solution soon. Ideally before the next release, if possible.

yama6a:
> Excellent @yama6a
> 
> Would this finally be enough to pull the rules into Grafana (from the evaluator), like it is possible with vanilla prometheus?

@penekk 
Probably yes-ish. 

For now, the endpoints will only be available on the `rules-evaluator` pod, not the `prometheus-frontend` pod, the latter of which you will probably use as a Prometheus-Datasource at the moment (running promQL queries, and fetching metrics/labels, etc).

A temporary hack could be to add the rule-evaluator pod as a separate Prometheus-Datasource in Grafana (once this is released), but I hope to make a PR for a proper solution soon-ish.

Currently, the frontend will blindly proxy all calls to `/api/*` to Google-Monitoring's prometheus endpoint, which doesn't work for these two new endpoints. I'm working on fixing that, and hope to have a solution soon. Ideally before the next release, if possible.

penekk:
> > Excellent @yama6a
> > Would this finally be enough to pull the rules into Grafana (from the evaluator), like it is possible with vanilla prometheus?
> 
> @penekk Probably yes-ish.
> 
> For now, the endpoints will only be available on the `rules-evaluator` pod, not the `prometheus-frontend` pod, the latter of which you will probably use as a Prometheus-Datasource at the moment (running promQL queries, and fetching metrics/labels, etc).
> 
@yama6a 
Actually the frontend is sort of 'out of the picture' because of datasyncer setup but I get your point.

> A temporary hack could be to add the rule-evaluator pod as a separate Prometheus-Datasource in Grafana (once this is released), but I hope to make a PR for a proper solution soon.
> 
That's what I was planning to do anyway :)

> Currently, the frontend will blindly proxy all calls to `/api/*` to Google-Monitoring's prometheus endpoint (https://monitoring.googleapis.com/v1/projects/PROJECT_ID/location/global/prometheus), which doesn't work for these two new endpoints. I'm working on fixing that, and hope to have a solution soon. Hopefully before the next release.
Good to know. If there is any way I can help with the whole thing (testing and what not, not that versed with go yet) just let me know.

And thank you a lot for working on this.

bwplotka (approved):
Thank you so much! LGTM. Presubmit failed with some cryptic (no error) log. Retried let's see.

bwplotka:
> Currently, the frontend will blindly proxy all calls to /api/* to Google-Monitoring's prometheus endpoint (https://monitoring.googleapis.com/v1/projects/PROJECT_ID/location/global/prometheus), which doesn't work for these two new endpoints. I'm working on fixing that, and hope to have a solution soon. Hopefully before the next release.

Nice! Yea, perhaps it's a good feature for `frontend` proxy or some separate proxy like thanos-querier or promxy. One day in Cloud Alerting too? 🤞🏽

cmd/rule-evaluator/internal/alerts.go
```diff
@@ -0,0 +1,24 @@
+package internal
```
bwplotka, line 1: Can you add same license as in other files? I think this is why presubmit fails 🤗

bwplotka:
Presubmit error is cryptic but it's detecting files without license:

```
#30 DONE 21.4s
/home/runner/go/bin/addlicense-v1.1.1 -check -ignore 'third_party/**' -ignore 'vendor/**' .
cmd/rule-evaluator/internal/alerts.go
cmd/rule-evaluator/internal/alerts_test.go
cmd/rule-evaluator/internal/api_test.go
make: *** [Makefile:144: regen] Error 1
```

cmd/rule-evaluator/internal/alerts.go
yama6a: done

yama6a:
@penekk 
> @yama6a Actually the frontend is sort of 'out of the picture' because of datasyncer setup but I get your point.
> 
> > A temporary hack could be to add the rule-evaluator pod as a separate Prometheus-Datasource in Grafana (once this is released), but I hope to make a PR for a proper solution soon.
> 
> That's what I was planning to do anyway :) (after seeing your previous contribution from a while back)

If you add the rule-evaluator as a data-source I would expect a bunch of UI and log errors, due to the fact that Grafana is going to try to use this as a query-datasource (which of course won't work). and Grafana-users can pick it as a datasource for their Dashboards and in the Metrics-Explorer and such. Not the end of the world of course, and perhaps the lesser evil in your case, just a fair warning.

lyanco:
Thank you for the contribution!! Should be a nice addition for debugging and general visibility.

Question for you - given that managed collection deploys one rule evaluator per cluster, there's no way to get a global view of your rules with this endpoint. You could only get a global view if there was only one rule evaluator, such as if you are using self-deployed evaluation or if you only use GlobalRules. 

Cloud Alerting's promql alerts work the same way as global rules but they can give you a better view in a centralized place. Have you looked into that? Is there some sort of feature deficit there that is stopping you from using it?

yama6a:
@lyanco 
> Question for you - given that managed collection deploys one rule evaluator per cluster, there's no way to get a global view of your rules with this endpoint. You could only get a global view if there was only one rule evaluator, such as if you are using self-deployed evaluation or if you only use GlobalRules.

I see what you mean; I suppose it depends on the individual setup and use-case.

In our case, we have one cluster (and hence one rule-evaluator) with one Grafana, per project. So, the global view is no problem if we pipe the rules via rule-evaluator into Grafana as a datasource. If there would be multiple clusters with one rule-evaluator respectively, we could still opt for adding all of those as datasources to a central Grafana, and use that as a place of consolidation to achieve a global view, or something else that ingests multiple rule-evaluator endpoints.

I'm not sure how using only GlobalRules would help, as you mentioned. Could you elaborate how this would affect a setup with multiple clusters? Wouldn't we still have multiple rule-evaluators which ingest the GlobalRules from the different clusters?

> Cloud Alerting's promql alerts work the same way as global rules but they can give you a better view in a centralized place. Have you looked into that? Is there some sort of feature deficit there that is stopping you from using it?

We did look into it, and it essentially boils down to developer experience, with GlobalRules adhering to the per-se industry standard of PrometheusRule resources (=> familiarity + ease of migration), and a IaC-decision of Terraform vs. Kubernetes or Atlantis vs ArgoCD. The full answer here is a bit more involved. 

Do you guys have a Slack or a similar communications channel? Would be happy to run it by you if you want, as well as see what architectural solution could be beneficial to everyone using GMP (With reference to https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1162). Would be happy to contribute more to this project.

bwplotka (approved):
Thanks!

lyanco:
@yama6a 

Got it, thanks for the explanation. If you have 1:1:1 cluster:project:grafana then yes this should work for you. If you have more than one cluster, you'd have to have multiple datasources in Grafana connecting to multiple rule-evaluators to show you all your rules, and it wouldn't be a single pane of glass (you'd have to tab between datasources).

In this scenario, designating one cluster as your Rules cluster and only using GlobalRules would allow you to connect Grafana to that rule-evaluator only and get a single pane of glass that way. But then you lose the tenancy benefits of Rules/ClusterRules resources. It wouldn't be recommended. FWIW if you use GlobalRules we recommend only putting them in one rule-evaluator, otherwise you end up getting collisions.

Noted on the cloud alerting experience, that makes sense. We're not getting rid of the cluster-local rule path, although we are investing in it less.

We made a chat room in Google Chat for you... I think bartek might have invited you?

PR #1162, opened 2024-10-03
feat: add rules-evaluator proxy calls to frontend

yama6a:
Enabling to query `/api/v1/rules` and `/api/v1/alerts` from the `frontend` by proxying these calls to the `rule-evaluator` pod.

This PR also creates a k8s service manifest for the `rule-evaluator` in order to be able to reach it from `frontend`

Note:
~~There was no proper way to test the k8s manifests in an actual cluster. If there is a way to test this setup properly before merging, that would probably be advisable.~~
Update: Added e2e tests

### DIFF
--- a/cmd/frontend/main.go
+++ b/cmd/frontend/main.go
@@ -34,6 +34,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/GoogleCloudPlatform/prometheus-engine/cmd/frontend/internal/rule"
+	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/ui"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/oklog/run"
@@ -42,8 +44,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"google.golang.org/api/option"
 	apihttp "google.golang.org/api/transport/http"
-
-	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/ui"
 )
 
 const projectIDVar = "PROJECT_ID"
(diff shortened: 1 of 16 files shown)


pintohutch:
Thanks @yama6a - for e2e testing, see our tests in https://github.com/GoogleCloudPlatform/prometheus-engine/tree/main/e2e. 

They deploy a full stack running to a kind cluster and do assertions on the workloads therein. You could add some testing there to double-check the manifests etc

bwplotka (approved):
Nice, let's add some e2e test, otherwise I'm supportive, code looks good 👍🏽

HThorsell:
Hi @bwplotka! Im Henrik, colleague of Yama here. We have added to the e2e tests as requested, please have another look at our PR.

cmd/frontend/main.go
```diff
@@ -65,6 +65,8 @@ var (
 
 	targetURLStr = flag.String("query.target-url", fmt.Sprintf("https://monitoring.googleapis.com/v1/projects/%s/location/global/prometheus", projectIDVar),
 		fmt.Sprintf("The URL to forward authenticated requests to. (%s is replaced with the --query.project-id flag.)", projectIDVar))
+
+	rulesEvaluatorURLStr = flag.String("rules.target-url", "http://rule-evaluator.gmp-system.svc.cluster.local:19092", "The URL to forward requests to the rules evaluator.")
```
bwplotka, line 69: Question for future compatibility. Should we make it open ended and... accept list one day? (aka repeated string flag)

It's ~trivial to merge rules, we added same logic in Thanos Querier years ago (we could even import/copy the exact code): https://github.com/thanos-io/thanos/blob/main/pkg/rules/proxy.go

The reason is that truly scalable rule-eval is multi instance, so this will come handy sooner or later.

bwplotka (approved):
Thanks one question, otherwise LGTM!

cmd/frontend/main.go
yama6a: @bwplotka 
That's a neat idea! I like it. Give us a day or three to change it, before we merge this :)

yama6a:
@bwplotka @pintohutch 
We added an e2e test. We also deployed the changes to the rule-evaluator, the k8s service, and the frontend to a GKE test-cluster and smoke-tested the endpoints. Everything works as expected: rule-evaluator endpoints, the service, and the endpoint-forwarding: `frontend-pod -> rule-evaluator-service -> rule-evaluator-pod`

We refactored quite a lot in this PR, improving on the structure and testability of the code, so you might want to re-review.

bwplotka (changes requested):
Nice! I had some readability comments... ended up doing a PR to fix this, might be easier. I left the functionality untouched really -- great job!

My recommendations are here https://github.com/yama6a/prometheus-engine/pull/1

HThorsell:
> Nice! I had some readability comments... ended up doing a PR to fix this, might be easier. I left the functionality untouched really -- great job!
> 
> My recommendations are here [yama6a#1](https://github.com/yama6a/prometheus-engine/pull/1)

Thank you! Your suggestions have been applied, ready for another round.

bwplotka (approved):
LGTM, thanks!

HThorsell:
Ive squashed the commits to comply with the presubmit check for commit message format

bwplotka:
Thanks! 💪🏽

PR #1175, opened 2024-10-18
feat: add /api/v1/status/buildinfo endpoint for Grafana compatibility

yama6a:
## Purpose 

This PR will make the GMP frontend compatible with Grafana for alerting and recording rules. It will also enable the rule-evaluator to be used as a datasource, albeit only for alerting and recording rules. 

Currently, Grafana won't attempt to import any rules without the `/api/v1/status/buildinfo` endpoint available. This PR adds it.

Adding the rule-evaluator as a Prometheus Datasource works as well with this feature. It produces an error when adding the datasource, due to it not implementing the `/api/v1/query` endpoint and its siblings. It does however succeed in importing the rules into Grafana. The GMP frontend will work without said error message.

## Details

The `/api/v1/status/buildinfo` endpoint is used by Grafana in order to determine the "flavor" of a Prometheus Datasource (Mimir / Cortex / Vanilla Prometheus).

Under certain circumstances, Grafana also probes the endpoint `/api/v1/rules/test/test` and expects the content to contain the words "page not found", in order to determine whether the Prometheus Datasource implements the ruler-API, which would allow editing rules via the Grafana UI (which is not supported in vanilla Prometheus or GMP)

## References
- https://github.com/grafana/grafana/blob/65ce173d3fd4a98f7763200cdddb80009a23d59b/public/app/features/alerting/unified/api/buildInfo.ts#L150
- https://github.com/grafana/grafana/blob/65ce173d3fd4a98f7763200cdddb80009a23d59b/public/app/features/alerting/unified/api/buildInfo.ts#L196
- https://github.com/grafana/grafana/blob/65ce173d3fd4a98f7763200cdddb80009a23d59b/public/app/features/alerting/unified/api/ruler.ts#L142

### DIFF
--- a/cmd/frontend/main.go
+++ b/cmd/frontend/main.go
@@ -35,6 +35,8 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/prometheus-engine/cmd/frontend/internal/rule"
+	"github.com/GoogleCloudPlatform/prometheus-engine/internal/promapi"
+	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/export"
 	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/ui"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -112,6 +114,12 @@ func main() {
 		ruleEndpointURLs = append(ruleEndpointURLs, *ruleEndpointURL)
 	}
 
+	version, err := export.Version()
+	if err != nil {
+		_ = level.Error(logger).Log("msg", "Unable to fetch module version", "err", err)
+		os.Exit(1)
+	}
+
 	var g run.Group
 	{
 		term := make(chan os.Signal, 1)
@@ -154,8 +162,11 @@ func main() {
 		)
 
 		server := &http.Server{Addr: *listenAddress}
+		buildInfoHandler := http.HandlerFunc(promapi.BuildinfoHandlerFunc(log.With(logger, "component", "buildinfo-handler"), "frontend", version))
+		http.Handle("/api/v1/status/buildinfo", buildInfoHandler)
 		http.Handle("/metrics", promhttp.HandlerFor(metrics, promhttp.HandlerOpts{Registry: metrics}))
 		http.Handle("/api/v1/rules", http.HandlerFunc(ruleProxy.RuleGroups))
+		http.Handle("/api/v1/rules/", http.NotFoundHandler())
 		http.Handle("/api/v1/alerts", http.HandlerFunc(ruleProxy.Alerts))
 		http.Handle("/api/", authenticate(forward(logger, targetURL, transport)))
 
--- a/cmd/rule-evaluator/main.go
+++ b/cmd/rule-evaluator/main.go
@@ -35,6 +35,7 @@ import (
 
 	"cloud.google.com/go/compute/metadata"
 	"github.com/GoogleCloudPlatform/prometheus-engine/cmd/rule-evaluator/internal"
+	"github.com/GoogleCloudPlatform/prometheus-engine/internal/promapi"
 	"github.com/GoogleCloudPlatform/prometheus-engine/pkg/export"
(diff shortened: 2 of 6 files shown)


cmd/frontend/main.go
```diff
@@ -154,7 +162,10 @@ func main() {
 		)
 
 		server := &http.Server{Addr: *listenAddress}
+		buildInfoHandler := http.HandlerFunc(promapi.BuildinfoHandlerFunc(log.With(logger, "component", "buildinfo-handler"), "frontend", version))
+		http.Handle("/api/v1/status/buildinfo", buildInfoHandler)
 		http.Handle("/metrics", promhttp.HandlerFor(metrics, promhttp.HandlerOpts{Registry: metrics}))
+		http.Handle("/api/v1/rules/test/test", http.NotFoundHandler()) // Magical endpoint used by Grafana to check whether a datasource has the ruler-api enabled. Should contain the words "page not found" in the response.
```
bwplotka, line 168: Nice! Should we instead ensure that all unknown paths respond with `NotFoundHandler`? 🤔 Sounds like a required part of rule API compatibility here?

cmd/rule-evaluator/main.go
```diff
@@ -392,9 +393,16 @@ func main() {
 			}
 		})
 
+		// https://prometheus.io/docs/prometheus/latest/querying/api/#build-information
+		buildInfoHandler := promapi.BuildinfoHandlerFunc(log.With(logger, "handler", "buildinfo"), "rule-evaluator", version)
```
bwplotka, line 397: We might need to stick to Prometheus versions here to mimic this API - it essentially tells what version of Prometheus we are compatible with. But if that works as well, we can iterate later (:

yama6a:
@bwplotka Can we merge this?

internal/promapi/buildinfo.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	promapiv1 "github.com/prometheus/prometheus/web/api/v1"
+)
+
+const (
+	timestampFormat              = "20060102-15:04:05"
+	buildinfoPath                = "/api/v1/status/buildinfo"
+	premoetheusDepencencyVersion = "1.8.2" // current version of the prometheus/prometheus dependency.
```
bwplotka, line 32:
```suggestion
	exposedVersion = "2.55.0" // We support APIs similar to the last pre 3.0 Prometheus version.
```

bwplotka (approved):
Thanks! I don't get how we got the 1.8 version as we depend on newer Prometheus/Prometheus code, but nevertheless what matters is what API the forwarded server supports. For now let's say what GMP supports and it's the latest Prometheus version, at least that's the aim (: at some point this could be a flag, but ok for now!

LGTM after that fix, sorry for lag!

yama6a:
@bwplotka 
> Thanks! I don't get how we got the 1.8 version as we depend on newer Prometheus/Prometheus code, but nevertheless what matters is what API the forwarded server supports. For now let's say what GMP supports and it's the latest Prometheus version, at least that's the aim (: at some point this could be a flag, but ok for now!
> 
> LGTM after that fix, sorry for lag!

No worries; I was referring to this:
https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/go.mod#L27

But I might have misinterpreted the dependency. But what you say makes sense. We should reference the API version that's compatible, so looks good to me, I committed your suggested change. :)

bwplotka:
Yes, it confuses everyone, it's because Prometheus does not follow proper Go Module semver ... which is difficult to follow in retrospective due to https://www.reddit.com/r/golang/comments/ipwea6/go_modules_have_a_v2_problem/ 🙈

yama6a:
@bwplotka yeah, my bad, I forgot about conform. I rewrote branch history, all checks seem to pass now.

bwplotka (approved):
Thanks!

PR #1176, opened 2024-10-23
feat: add log level flag to rules-evaluator & frontend

hsmatulisgoogle:
Adding a log level flags to both `rules-evaluator` & `frontend` binaries. The flag is named `--log.level`


pintohutch (approved):
LGTM but we should get @bwplotka's input as I believe OSS Prometheus has a way of doing this - maybe we could get inspiration? Or maybe this is fine

bwplotka (approved):
Thx!

bwplotka (commented):
Hm, it's true there might be some helper somewhere but we can fix later, thanks!

PR #1190, opened 2024-11-19
chore: fix cross-compile issues

bernot-dev:
Consistently use dynamically linked BoringCrypto with cross-compiling.


pintohutch (commented):
What's the motivation to use dynamic linking over static?

pintohutch (approved):

PR #1207, opened 2025-01-07
Update crypto and net dependencies

hsmatulisgoogle:
See b/387650500


bwplotka (approved):
Thanks, will try to release that.

Conform asks to have a certain commit name, but I can ignore this for now. Also please don't use `release-0.XY` branch prefix, it might collide with some rules for release branches.  (:

bwplotka:
Fixing commit name

PR #1209, opened 2025-01-07
feat: expand VPA to all GMP components

### LINKED ISSUE #975: Enable VPA feature for rule-evaluator
Similar to https://github.com/GoogleCloudPlatform/prometheus-engine/pull/943 but for the rule-evaluator.

bernot-dev:
Prior to this change, when the VPA feature is enabled, it is only used to autoscale the collector pods.

This change expands the VPA feature to autoscale the alertmanager, rule-evaluator, and operator.

Fixes #975


bwplotka (approved):
Looks amazing, let's just make sure to add more documentation on the type `vpa` OperatorConfig field to make this explicit.

Thanks!

PR #1211, opened 2025-01-08
fix: tolerate gke-managed-component nodes

bernot-dev:
Make GMP comply with gke-managed-component standard.

https://cloud.google.com/kubernetes-engine/docs/how-to/isolate-workloads-dedicated-nodes#recommendations-best-practices


pintohutch (approved):
Seems reasonable - and shouldn't hurt anything

PR #1265, opened 2025-01-15
chore: unfork controller-gen

bernot-dev:
This PR unforks `controller-gen`.

In the process, it adds namespace to `SecretKeySelector` objects for namespace-scoped parent objects (e.g. PodMonitorings), thus allowing a PodMonitoring to reference secrets in a different namespace. There is no particular reason this functionality should not be supported.


TheSpiritXIII:
> There is no particular reason this functionality should not be supported.

The reason why is that we should be using the namespace of the `PodMonitoring` itself. Secrets are _not_ cluster-scoped resources, so we violate the multi-tenancy model by treating them as such.

Originally, I played with generics and had something like a `TSecret` object with two versions: A cluster-scoped version and a namespace-scoped version which ClusterPodMonitoring and PodMonitoring used respectively. Before forking controller-gen, I decided this was too much effort to refactor across our codebase, and simply always had namespace but disabled it via a webhook for PodMonitoring.

I would consider trying one of those options unless we decide we want to support namespace for sure!

TheSpiritXIII (approved):

PR #1267, opened 2025-01-17
refactor: default podmonitorings with CRDs

bernot-dev:
Replace defaulting logic in webhook with CRD-based defaults.


bwplotka (approved):
Thanks! One day we might want to setup e2e for those cases, but good for now 👍🏽

PR #1310, opened 2025-01-21
chore: upgrade controller-runtime

bernot-dev:


bwplotka (approved):
Thanks! What's with this defaulting to empty. Kind of obscuring docs a bit, but I assume it's the aftermath of the upgrade?

PR #1333, opened 2025-01-31
feat: Add support for fetching config from env

abhisek:
I am following the official documentation to setup Grafana with datasource-syncer for Managed Prometheus:
https://cloud.google.com/stackdriver/docs/managed-prometheus/query#use-kubernetes

Our standard approach for providing config & secrets are through environment. Adding environment as a fallback source for config when not provided through command line

### DIFF
--- a/cmd/datasource-syncer/main.go
+++ b/cmd/datasource-syncer/main.go
@@ -63,10 +63,15 @@ func main() {
 	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
 	logger = log.With(logger, "caller", log.DefaultCaller)
 
-	if len(*datasourceUIDList) == 0 {
-		//nolint:errcheck
-		level.Error(logger).Log("msg", "--datasource-uid must be set")
-		os.Exit(1)
+	if *datasourceUIDList == "" {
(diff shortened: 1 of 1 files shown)


abhisek:
> This seems like a reasonable feature to support.
> 

Thanks. This will help us setup the `datasource-syncer` within our k8s while maintaining single source of truth for config (i.e. our env providers).

> I'm not sure we would just want to support providing just a subset of the flags though. Any particular rationale for those three?
> 

Specifically for our use, the CA/Key file flags are not required from environment because we can make them available as volume mounts with any file path. A k8s secret is mounted as a file and the file name is a local concern (i.e. within the job yaml) hence the actual path does not matter and can hardcoded as a command line argument. 

However I have added env fallback for `ca_file`, `cert_file`, `key_file` for consistency in user experience. I have consciously decided not to add a fallback for `gcmEndpointOverride` because I feel the defaults are appropriate. In rare cases where it needs to be overridden I think we should force its explicit usage through command line .

> Also, the commit message lister will require a lowercase "A".
> 

Fixed.

> Thanks for the PR!

PR #1337, opened 2025-02-05
feat: workload labels for 0.14

bwplotka:
Backported from 0.15 https://github.com/GoogleCloudPlatform/prometheus-engine/commit/f197f444a3a8646c54d5be1097d7c55d45b94eb4 plus presubmit fixes from https://github.com/GoogleCloudPlatform/prometheus-engine/commit/aae0d4a48ff39cbe167c8961ece851b17221a158

Not changing Go version in docker images, although new minor releases exists: https://go.dev/doc/devel/release#go1.23.minor

Context: b/394290278


pintohutch (changes requested):
Actually - it looks like there's some GH Action failures - could you address those first?

bwplotka:
ofc, just noticed them. 

It's not so simple. The failures are due to using new things from Go 1.23 and bumping the go.mod to require 1.23 and then building some (not operator) images with Go 1.22... I will try some hacks, but likely we need to bump all images to latest 1.23

pintohutch:
Oy - yea. Given the headache, what about just backporting 0.15 as-is instead of trying to patch a previous release?

bwplotka:
Not feasible due to go/gmp:014rollout - but we could challenge it.

I managed to make it green, good to go?

pintohutch (approved):

PR #1339, opened 2025-02-06
chore: add script to bump golang images

bernot-dev:
Convenience script for devs to bump all Go images to latest. Requires [crane](https://github.com/google/go-containerregistry/tree/main/cmd/crane).


bwplotka (approved):
Thanks!

PR #1343, opened 2025-02-07
0.14: Cherry-pick image changes, given Louhi scripts were updated; upgraded to the latest Go 1.23.6

bwplotka:
b/394290278

PR #1344, opened 2025-02-07
fix: rule-eval docker build for 0.14

bwplotka:
Fix & testing https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1343


bwplotka:
Failures are due to non existing (in prod) Prom image.

PR #1381, opened 2025-02-12
docs: add Tensorflow Serving and TorchServe sample yamls

yqlu:
Convention from past docs we are hewing to here:
- if the file is a valid yaml, it ends with .yaml
- if the file is a diff with '+' to illustrate how you should be modifying your existing setup, it ends with .snippet so users don't blindly apply it.


yqlu:
Timing of this needs to be coordinated with the corresponding docs cl/725344646

PR #1384, opened 2025-02-13
Bump release/0.12 dependencies

bernot-dev:
Bump x/crypto and x/net for release/0.12.


bwplotka (approved):
Thanks, but why patching so old releases? 🤔

PR #1487, opened 2025-03-14
chore: simplify dependabot config

bernot-dev:
Use [directories instead of directory](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--) to simplify Dependabot configuration.

Add `release/0.16` branch to scanning list.

Scan `datasource-syncer` and `go-synthetic`.


bwplotka (approved):
Thanks, amazing!

PR #1526, opened 2025-03-14
test: fix flaky crd test

bernot-dev:
The CRD test for a large podmonitoring introduced in #1437 would periodically fail because it could create short regex expressions that would match protected labels. This fix generates regex expressions that will never match protected labels.


bwplotka (approved):
Thanks!

PR #1538, opened 2025-03-20
fix: podmonitoring defaulting

bernot-dev:
Fix bug with PodMonitoring/ClusterPodMonitoring improperly defaulting `Spec.TargetLabels.Metadata`.

TargetLabels are split into two similar structs - one for ClusterPodMonitoring, one for PodMonitoring.

Relevant test `TestCRDDefaulting` is added to verify the defaulting behaves correctly.

This change also reintroduces of `k8s.io/code-generator`, which seems to have been removed by mistake. This is necessary for the generated DeepCopy functions on the new type.

Relevant documentation is updated accordingly.


e2e/crd_validation_test.go
```diff
@@ -96,6 +97,241 @@ func cleanupKindCluster(t *testing.T, clusterName string) func() {
...
+func TestCRDDefaulting(t *testing.T) {
+	t.Parallel()
+
+	c := createKindCluster(t)
+
+	type cpmTest struct {
+		obj  *monitoringv1.ClusterPodMonitoring
+		want *monitoringv1.ClusterPodMonitoring
+	}
+	runCPM := func(t *testing.T, tests map[string]cpmTest) {
```
pintohutch, line 109: question(non-blocking): this seems to deviate a bit from vanilla [table-driven testing](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go), which is much more common in the codebase.

Is there a reason for not going with a pattern like [this](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/bc30e6a2200ee3c531ba8c4a208256483c2e1a06/cmd/rule-evaluator/main_test.go#L78)? I find it a bit more conventional and easier to read.

pkg/operator/apis/monitoring/v1/pod_types.go
```diff
@@ -245,17 +243,34 @@ type ScrapeEndpoint struct {
...
+	// +kubebuilder:default=container;pod;top_level_controller_name;top_level_controller_type
+	Metadata *[]string `json:"metadata,omitempty"`
+	// Labels to transfer from the Kubernetes Pod to Prometheus target labels.
+	// Mappings are applied in order.
+	// +kubebuilder:validation:MaxItems=100
+	FromPod []LabelMapping `json:"fromPod,omitempty"`
+}
+
+// ClusterTargetLabels configures labels for the discovered Prometheus targets.
+type ClusterTargetLabels struct {
```
pintohutch, line 263: question: is there any risk of backward compatibility by changing the go struct in the type here?

charts/operator/crds/monitoring.googleapis.com_clusterpodmonitorings.yaml
```diff
@@ -695,23 +688,26 @@ spec:
                     maxItems: 100
                     type: array
                   metadata:
+                    default:
```
pintohutch, line 691: question: is this the crux of the change? i.e moving the defaulting down a layer into the `.metadata` field?

pintohutch (commented):
Mostly questions - I don't think I have any major concerns. I just want to ensure I understand the change before approving better.

e2e/crd_validation_test.go
```diff
@@ -96,6 +97,241 @@ func cleanupKindCluster(t *testing.T, clusterName string) func() {
 	}
 }
 
+func TestCRDDefaulting(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
pintohutch: > This specific test creates a kind cluster that is not running the operator to isolate the validation/defaulting behavior built into the schema, without webhooks interfering.

Is that desired? That sounds more like a unit test over an E2E test. We'll never have a real life scenario where the operator isn't running in the cluster, right?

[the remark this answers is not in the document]
pintohutch: Gotcha makes sense - thanks for clarifying.

e2e/crd_validation_test.go
pintohutch: Ok - I think it's ultimately a style thing. My preference is for more bite-sized test methods (e.g. break out CPM tests into a separate function altogether). I generally find that more reviewable/readable. But it's subjective, so I won't block.

Thanks for the links!

hack/presubmit.sh
```diff
@@ -37,6 +37,27 @@ warn() {
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 CRD_DIR=${REPO_ROOT}/charts/operator/crds
 
+codegen_diff() {
```
pintohutch, line 40: nit: do you know when this was removed? I'm surprised it didn't break anything when it happened...

pkg/operator/apis/monitoring/v1/pod_types.go
pintohutch: SGTM

charts/operator/crds/monitoring.googleapis.com_clusterpodmonitorings.yaml
pintohutch: Ack thanks

pintohutch (approved):
LGTM % nit question

PR #1539, opened 2025-03-20
fix: default podmonitorings correctly

bernot-dev:
Split PodMonitoring/ClusterPodmonitoring structs to separate defaulting logic for each set of Metadata on the TargetLabels.


pintohutch (approved):

pkg/operator/apis/monitoring/v1/pod_types.go
```diff
@@ -245,17 +243,34 @@ type ScrapeEndpoint struct {
 // TargetLabels configures labels for the discovered Prometheus targets.
 type TargetLabels struct {
```
bwplotka: cc

PR #1552, opened 2025-03-26
fix: add secret in the ClusterPodMonitoring of Inference Gateway

JeffLuoo:


pintohutch:
Hey @bernot-dev could you PTAL?

bwplotka:
We probably forgot to change public docs after this merge - got pinged about a broken link (I attempted to fix those and cc-ed you @JeffLuoo) - next time we probably want to time those.

bwplotka:
BTW: I wonder if PodMonitoring would not work instead if we put the secret to the correct namespace 🤔

PR #1553, opened 2025-03-26
chore: bump golangci-lint to v2

bernot-dev:
Bump golangci-lint to v2.


hsmatulisgoogle:
This is a big change, is there a way we could split it up? Perhaps the change in crypto libraries could be a separate PR to split the vendoring changes into another PR. But otherwise it looks good to me

hsmatulisgoogle (approved):

PR #1555, opened 2025-03-28
chore: bump prometheus

bernot-dev:
Bump prometheus/prometheus to 2.53.4 (current LTS).

Duplicated `Config` and `ScrapeConfig` types in `pkg/operator/apis/monitoring/v1/scrape_config.go` to achieve different YAML marshaling behavior. This ensures backwards compatibility with 2.45 until we upgrade GoogleCloudPlatform/prometheus to 2.53.

Other than that, mostly chasing upstream with moving packages and changing function signatures.


pkg/operator/apis/monitoring/v1/scrape_config.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/alecthomas/units"
+	"github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+	promconfig "github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/discovery"
+	"github.com/prometheus/prometheus/model/relabel"
+)
+
+type Config struct {
```
bwplotka, line 28: Some sustainability questions:

1. We desperately need commentary on why we are vendoring this.
2. How to track what fields we removed? What's the process on when we are upgrading Prometheus and more fields needs to be removed or added?

bwplotka: I think Prometheus would not accept omitzero even in the latest version, because we never envisioned our structs to be Marshalled (only unmarshalled). Thanks!

bwplotka (approved):
Thanks! LGTM, thanks for documenting the vendored scrape config code 💪🏽

bwplotka:
I think I have long term solution for our yaml issues, we could try https://github.com/goccy/go-yaml?tab=readme-ov-file#features going forward.

bwplotka:
Added issue for this https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1629

bwplotka (changes requested):
I think there is a quick win to avoid vendoring config, discussed in https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1629#issuecomment-2796484805

Should we try it first? (:

bwplotka:
Done in https://github.com/GoogleCloudPlatform/prometheus-engine/commit/ff932eec20dde93ccdee120c34fc3ae3ec93a8b8

Issue #1550, opened 2025-03-25
Managed alertmanager no longer sending slack alerts


gravelg:
Hello, 

We have noticed that since March 18th, our managed alertmanager is no longer sending alerts. We hve not made any configuration changes to the alertmanager config contained in the `alertmanager` secret, and we can see that the configuration is loaded correctly.

Alertmanager logs are full of errors now however:
```
{"caller":"main.go:192","level":"info","msg":"Starting Alertmanager","ts":"2025-03-22T08:25:10.555Z","version":"(version=0.27.0-gmp.1, branch=, revision=0ddd406d04a5076cb73567d5a11972bfefc7e833)"}
{"build_context":"(go=go1.22.7 X:boringcrypto, platform=linux/amd64, user=, date=2024-09-20T12:17:58+00:00, tags=boring)","caller":"main.go:193","level":"info","ts":"2025-03-22T08:25:10.555Z"}
{"caller":"cluster.go:683","component":"cluster","interval":"2s","level":"info","msg":"Waiting for gossip to settle...","ts":"2025-03-22T08:25:10.556Z"}
{"caller":"coordinator.go:113","component":"configuration","file":"/alertmanager/config_out/config.yaml","level":"info","msg":"Loading configuration file","ts":"2025-03-22T08:25:10.593Z"}
{"caller":"coordinator.go:126","component":"configuration","file":"/alertmanager/config_out/config.yaml","level":"info","msg":"Completed loading of configuration file","ts":"2025-03-22T08:25:10.594Z"}
{"address":"[::]:9093","caller":"tls_config.go:313","level":"info","msg":"Listening on","ts":"2025-03-22T08:25:10.597Z"}
{"address":"[::]:9093","caller":"tls_config.go:316","http2":false,"level":"info","msg":"TLS is disabled.","ts":"2025-03-22T08:25:10.597Z"}
{"caller":"coordinator.go:113","component":"configuration","file":"/alertmanager/config_out/config.yaml","level":"info","msg":"Loading configuration file","ts":"2025-03-22T08:25:11.659Z"}
{"caller":"coordinator.go:126","component":"configuration","file":"/alertmanager/config_out/config.yaml","level":"info","msg":"Completed loading of configuration file","ts":"2025-03-22T08:25:11.661Z"}
[...]

PR #1558, opened 2025-03-28
fix: secrets being obfuscated accidentally #1550

bwplotka:
Reproduces: https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1550


bwplotka:
With this we need a quick fix. We need to play with marshalling to customize it. E.g. one way is to only unmarshal google cloud section and remove it manually somehow from the source, then append on top

pkg/operator/operator_config.go
```diff
@@ -485,10 +484,10 @@ func (r *operatorConfigReconciler) ensureAlertmanagerConfigSecret(ctx context.Co
 }
 
 type alertmanagerConfig struct {
```
hsmatulisgoogle, line 486: Could we take advantage of [this](https://github.com/prometheus/alertmanager/blob/9b35e7696449df7e99c97a6721daeb3f500f8242/config/config.go#L42-L45) to allow umarshalling of secrets when required instead of cloning the upstream config? 

```
// MarshalSecretValue if set to true will expose Secret type
// through the marshal interfaces. Useful for outside projects
// that load and marshal the Alertmanager config.
var MarshalSecretValue bool = commoncfg.MarshalSecretValue
```
IE.  we can set `promcommonconfig.MarshalSecretValue = true` when needed

hsmatulisgoogle (changes requested):

pkg/operator/operator_config.go
```diff
@@ -485,10 +484,10 @@ func (r *operatorConfigReconciler) ensureAlertmanagerConfigSecret(ctx context.Co
...
-	alertmanagerconfig.Config `yaml:",inline"`
+	Config AlertmanagerConfig `yaml:",inline"`
 
 	// Google Cloud configuration. Matches our fork's configuration.
-	GoogleCloud googleCloudAlertmanagerConfig `yaml:"google_cloud,omitempty"`
+	GoogleCloud googleCloudAlertmanagerConfig `yaml:"-"`
```
bwplotka, line 490: Is `-` sensible here?

bwplotka (commented):
This is great, generally it should work, but I would love to challenge this. We have opportunity to not put defaults in marshalled content and not introduce forks and many vendor changes to a critical bug fix.

I spoke to @bernot-dev today and we want to try a simpler fix with `map[string]string` re-encoding just for `google_cloud.external_url` field switch. This should be trivial to write and test and we will try to find time in between conference items.

bwplotka:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1559

PR #1598, opened 2025-04-02
chore: prepare for 0.15.3-rc.0 release

hsmatulisgoogle:
This PR prepares the release candidate v0.15.3-rc.0


bwplotka:
Note: We could do 0.15.2 as well but it's ok to do 0.15.3 as well. Since we started 0.15.3 with this merge, let's go for it!

PR #1602, opened 2025-04-07
fix: remove validation on frompod source label

bernot-dev:
The source label name does not need to be valid for the scrape config, just the destination label.


bwplotka (approved):
Thanks! Let's backport to 0.16 after this!

Issue #1629, opened 2025-04-09
Remove our yaml hacks by switching to goccy/go-yaml lib.


bwplotka:
Recently we have trouble customizing marshalling of 3rd party structs like:

* Prometheus omitzero problem that cause us to copy structs which is hard to sustain https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1555#issue-2955046159
* Alertmanager secret redaction hacks https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1562

I think we coud replace those frictions by replacing gopkg.in/yaml with https://github.com/goccy/go-yaml?tab=readme-ov-file#features

Another reason for this work is that [`gopkg.in/yaml.*` project seems to be deprecated](https://github.com/go-yaml/yaml?tab=readme-ov-file#this-project-is-unmaintained).

bwplotka:
We might want to try https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1629 long term

bwplotka:
Thanks for a quick look!

> [goccy/go-yaml doesn't seem to support omitzero](https://go.dev/play/p/7Dc7hOcAOGj), either.

Note: I think you mean `omitempty` not `omitzero`? [AFAIK omitempty is what we want and it's supported.](https://pkg.go.dev/github.com/goccy/go-yaml#section-readme:~:text=omitempty%20%20%20%20Only%20include%20the%20field%20if%20it%27s%20not%20set%20to%20the%20zero). Maybe in other libs it's called omitzero.

Anyway, I understand this problem it's not trivial.

However, the unmasking problem is one thing, but we still generate "garbage" for all empty fields, which impacts debugging, we hit scalability limits for k8s resource sizes sooner. Plus we still have challenge [with Prometheus struct compatibility](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1555#discussion_r2033623811) -- one could  say it's tmp, but next month we will do another upgrade and hit the same issue.

As a result I think it's worth to properly fix all our yaml pain once and for all, even if it involves getting our hands dirty in the world of yaml unmarshalling.

I tried today something and I think we could do a bit of magic like below, thanks to marshal customizations:

```
package yamlutil

import (
	"github.com/goccy/go-yaml"
)

func MarshalOmitEmpty(v any) ([]byte, error) {
	return yaml.MarshalWithOptions(v, yaml.CustomMarshaler(customOmitEmptyMarshaler()))
}

[...]

PR #1660, opened 2025-05-19
feat: prometheus 2.53.4 support

pintohutch:
We do this to prepare a release minor that supports the latest Prometheus 2.53
release.

We had to fix some merge conflicts due to upstream refactors; much of that work
was captured by cherry-picking in commit https://github.com/GoogleCloudPlatform/prometheus-engine/commit/c3d31f9c419b376844c61635a259f1b15c7b1c12.

Additionally we needed to update the Prometheus config tests to check for the
existence of some new fields that were introduced (e.g. http_headers).

We also remove the replace directive in the go.mod file for the older prometheus
and x/exp packages and pin them to prometheus 2.53.4.


pintohutch:
The e2e tests are failing because the Prometheus image is older than 2.53 and is emitting errors trying to load the config with unknown attributes. This is expected until we publish our forked 2.53 after merging https://github.com/GoogleCloudPlatform/prometheus/pull/229.

pintohutch:
The prometheus image bump looks like it worked and the tests are all passing now.

e2e/ruler_test.go
```diff
@@ -254,6 +254,7 @@ alerting:
     alertmanagers:
         - follow_redirects: true
           enable_http2: true
+          http_headers: null
```
bwplotka, line 257: I wished we could switch to https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1629 otherwise we have this small rolling restart risk (operator upgrades, but ruler/collector not yet). 

Since you want to put it in 0.15 looks like it won't happen and we have to live with that risk.

bwplotka (approved):
Thanks LGTM!

My only suggestion would be to do the (should be relatively easy)  switch for  https://github.com/GoogleCloudPlatform/prometheus-engine/issues/1629 to avoid extra unnecessary new empty field encoding in our YAML file (causing potential rolling upgrade compatibility issues). Given this is for 0.15 backport, we might want to skip that and keep the risk. 

Is there a reason to rush with 0.15 for this?

Also, can we now make 2.53.4 a default?

image

bwplotka:
Superseding with https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1696

PR #1662, opened 2025-05-21
feat: Add .gitattributes to minimize diffs

hsmatulis:
In theory this should help by minimizing diffs view for tagged files by default in the GitHub UI . See [the docs here](https://github.com/github-linguist/linguist/blob/main/docs/overrides.md)

The following files/directories are now marked as generated to reduce diffs:
- `vendor/**`
- `go.sum`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+vendor/** linguist-generated
+go.sum linguist-generated


hsmatulis:
> I think vendor is fine, go.sum is probably fine. But we definitely want to track changes to the other 3 manifests closely.

Makes sense, removed

PR #1663, opened 2025-05-22
Patch 0.12

bernot-dev:
Patch dependencies on release/0.12.


hsmatulis (changes requested):
Any way we can keep the linter check? If not, could we document why we have to remove the check?

hsmatulis (approved):

PR #1672, opened 2025-06-03
Fix: Exit operator if metadata fetching fails

bernot-dev:
When the operator starts on GCE, it attempts to fetch metadata such as project ID, cluster name, and location. Previously, if these calls failed, the errors were logged, but the operator would continue running with potentially empty default values for these critical parameters. This could lead to undefined behavior or errors later on.

This commit modifies the startup sequence in `cmd/operator/main.go`. If any of the metadata calls fail, the operator will now log the error and then exit with a status code of 1. This ensures that the operator does not attempt to run without essential configuration information.

Error messages have also been improved by wrapping the original errors with more context using `fmt.Errorf`.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -19,6 +19,7 @@ import (
 	"crypto/fips140"
 	"errors"
 	"flag"
+	"fmt"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


bwplotka (changes requested):
Thanks!

Metadata fills the default values, which can be overridden by flags. Failing on errors for default values would prohibit using operator on custom GCE/GKE with odd metadata servers where users might be able to override the flags to make things work.

I assume (would need to check) we don't use explicit flag, so for managed we might want this behaviour though - should we fail the binary on those errors ONLY if all 3 related flags are not specified? 🤔

PR #1674, opened 2025-06-09
fix:  GoogleCloud.Export collector config to ignore empty string

leonzz:
Compression and CredentialsFile could be empty string, in these cases they should not be populated into collector configmap.


pkg/operator/collection.go
```diff
@@ -276,18 +276,20 @@ func (r *collectionReconciler) ensureCollectorConfig(ctx context.Context, spec *
...
-	if spec.Credentials != nil {
-		credentialsFile = path.Join(secretsDir, pathForSelector(r.opts.PublicNamespace, &monitoringv1.SecretOrConfigMap{Secret: spec.Credentials}))
-	}
-
 	cfg.GoogleCloud = &GoogleCloudConfig{
 		Export: &GoogleCloudExportConfig{
-			Compression:     ptr.To(string(compression)),
-			CredentialsFile: ptr.To(credentialsFile),
+			Compression:     nil,
+			CredentialsFile: nil,
```
bwplotka, line 282: We could remove those lines as well:


```suggestion
```

bwplotka (approved):
It would epic to have a unit test for this, but otherwise LGTM!

pkg/operator/collection.go
leonzz: Done.

leonzz:
> It would epic to have a unit test for this, but otherwise LGTM!

Added a test by re-using one of the test cases in `TestCollectionReconcile` and verify the `google_cloud.export` section is empty.

It can catch the issue before the `collection.go`  change:  

```
--- FAIL: TestCollectionReconcile (0.02s)
    --- FAIL: TestCollectionReconcile/clusternodemonitoring:_update_status:_empty_endpoint (0.00s)
        collection.go:162: "level"=0 "msg"="reconciling collection"
        collection.go:495: "msg"="generating scrape config failed for ClusterNodeMonitoring endpoint" "error"="invalid definition for endpoint with index 0: invalid scrape interval: empty duration string" "namespace"="gmp-test" "name"="prom-example"
        collection_test.go:482: unexpected collector configmap (-want, +got):   (
              	"""
              	global: {}
              	google_cloud:
            - 	    export: {}
            + 	    export:
            + 	        compression: ""
            + 	        credentials: ""
              	"""
              )
FAIL
FAIL	github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator	67.158s
```

I didn't add the verification to all the test cases because some of them has long text in the `scrape_configs` section so there will be huge chunks of test data unrelated to the thing we're verifying.  Let me know if it makes sense-)

PR #1685, opened 2025-06-27
feat: add vulnupdatelist hack script

bwplotka:
Auxiliary Go script tailored for on-demand Go vulnerability requirements for Go upgrades. This is to speed up vuln fixing in the environment of old modules with a lot of breaking changes (we feasibly can't upgrade all deps to the latest, we have to minimize the upgrades to only those modules with the vulns.)

NOTE: I purposely make the script small e.g. not do the upgrades (do one thing and do thing well). The output can be feeded to a script later on that attempts to upgrades those in the CI systems.

NOTE2: Added special Go module for hack scripts so those don't pollute the main prometheus-engine module.


bwplotka:
Good question, it's because:

* We need this on demand for local use (e.g. scripting, local dev) to reduce the toil and increase the dev velocity.
* We are formatting the result into something easier to consume (potentially later used in scripts)
* This GH action does not check the severity.

bwplotka:
See one example use https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1699

bwplotka:
Will create a separate PR with a clean state.

PR #1698, opened 2025-07-04
fix: update test after GCM improved PromQL compatibility

bwplotka:


bwplotka:
Change was approved on main, merging.

PR #1699, opened 2025-07-04
feat: add hack script for fixing vulns on release branches

bwplotka:
A manual, idempotent script that updates Go vuln on our release branches.

We can hook it to Louhi or GH action at some point; the plan is to extend this to Docker/Go and npm one day.

Dependabot is doing similar, but IMO the eventual dependabot approach does not scale here (too many PRs, no way to bulk approve it, too long to wait for things to rebase/pass test); we need on-demand or at least atomic automation. Also we want portability to Louhi.

Depends on https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1685

### DIFF
--- a/hack/release-lib.sh
+++ b/hack/release-lib.sh
@@ -1,0 +1,227 @@
+#!/usr/bin/env bash
+# Copyright 2025 Google LLC
+#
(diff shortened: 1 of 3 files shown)


bwplotka:
Will create a separate PR with a clean state.

PR #1700, opened 2025-07-07
feat: add hack script for cutting RCs

bwplotka:
Also refactored hack/cmd/prepare_rc for reliability.

Depends on https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1699

### DIFF
--- a/.github/workflows/release-bot.yml
+++ b/.github/workflows/release-bot.yml
@@ -135,7 +135,8 @@ jobs:
         id: prepare
         working-directory: ./main_branch
         run: |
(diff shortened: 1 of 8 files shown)


bwplotka:
What's trigger_rc?

There are set of script I based on, that's why basing on those.

Let's sync internally, especially I wonder if the prepare_rc is complex enough for Go code 🤔 also at this point it does not prepare rc really (e.g. it does not regenerate manifests), so technically it's "update tags" or something.

PR #1739, opened 2025-08-22
chore: use 2.53.5-gmp.0-gke.6 (rc.6) version

bwplotka:
To test 2.53.5 with PE before 0.17.0 release.


bwplotka:
Ok, I forget it does not make sense to do that, because this image is only in staging...

PR #1741, opened 2025-08-26
fix: fix example delete metric desc script

bwplotka:
Seems like [ListMetricDescriptors no longer allow empty Filter ](https://github.com/GoogleCloudPlatform/prometheus-engine/actions/runs/17233161113/job/48891743848?pr=1740). Use this opportunity for cheaper server side calls and use regex filtering feature.


bwplotka:
Added internal ticket about this.

PR #1750, opened 2025-08-28
fix: revert to boringcrypto

bernot-dev:
Go Cryptographic Module is not yet NIST validated. Returning to BoringCrypto.

### DIFF
--- a/cmd/config-reloader/Dockerfile
+++ b/cmd/config-reloader/Dockerfile
@@ -23,16 +23,29 @@ COPY go.sum go.sum
 COPY vendor* vendor
 COPY cmd cmd
 
-ENV GOEXPERIMENT=noboringcrypto
-ENV CGO_ENABLED=0
-ENV GOFIPS140=latest
+ENV GOEXPERIMENT=boringcrypto
+ENV CGO_ENABLED=1
+ENV GOFIPS140=off
 ENV GOOS=${TARGETOS}
 ENV GOARCH=${TARGETARCH}
-RUN go build \
-  -ldflags="-X github.com/prometheus/common/version.Version=$(cat charts/values.global.yaml | yq '.version' ) \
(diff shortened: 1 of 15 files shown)


cmd/config-reloader/Dockerfile
```diff
@@ -23,16 +23,29 @@ COPY go.sum go.sum
...
+ENV GOFIPS140=off
 ENV GOOS=${TARGETOS}
 ENV GOARCH=${TARGETARCH}
-RUN go build \
-  -ldflags="-X github.com/prometheus/common/version.Version=$(cat charts/values.global.yaml | yq '.version' ) \
-  -X github.com/prometheus/common/version.BuildDate=$(date --iso-8601=seconds)" \
-  -o config-reloader \
-  cmd/config-reloader/*.go
+RUN if [ "${TARGETARCH}" = "arm64" ] && [ "${BUILDARCH}" != "arm64" ]; then \
+      apt install -y --no-install-recommends \
```
pintohutch, line 32: question: what is this doing effectively?

We should already be supporting multi-arch builds, no?

cmd/config-reloader/Dockerfile
```diff
@@ -23,16 +23,29 @@ COPY go.sum go.sum
...
+        gcc-aarch64-linux-gnu libc6-dev-arm64-cross; \
+      CC=aarch64-linux-gnu-gcc; \
+    fi && \
+    GOOS=${TARGETOS} GOARCH=${TARGETARCH} CC=${CC} \
+      go build \
+        -ldflags="-linkmode=external -extldflags=-static -X github.com/prometheus/common/version.Version=$(cat charts/values.global.yaml | yq '.version' ) \
```
pintohutch, line 38: Actually @bernot-dev I actually don't think we want these statically linked

PR #1765, opened 2025-09-12
refactor: reorder containers in rule-evaluator

bernot-dev:
Due to Kustomize/JSON Patch limitations, it solves problems for us if the rule-evaluator container is first in the manifest. This allows us to reuse the same patch for P4SA.


bwplotka:
Let's make sure to port it to 0.15 branch to have it for p4sa!

PR #1789, opened 2025-10-27
fix: update out of sync Prometheus image

hsmatulis:
The prometheus image in the example directory should be kept in sync with the one used by the 0.17 branch

### DIFF
--- a/examples/prometheus.yaml
+++ b/examples/prometheus.yaml
@@ -108,7 +108,7 @@ spec:
           privileged: false
       containers:
       - name: prometheus
-        image: gke.gcr.io/prometheus-engine/prometheus:v2.53.5-gmp.0-gke.13
+        image: gke.gcr.io/prometheus-engine/prometheus:v2.53.5-gmp.1-gke.2
         args:
         - --config.file=/prometheus/config_out/config.yaml
         - --storage.tsdb.path=/prometheus/data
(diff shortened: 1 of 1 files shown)


examples/prometheus.yaml
bwplotka: this is ideally bumped too, as mentioned in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1786/files#r2461450478

bwplotka (commented):
Thanks, added some comments. Also not sure if docs should link to unreleased 0.17 branch/release at the moment...

Also please update commit msg, that's why conform is complaining.

FYI, have you seen https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1786? 

Nevertheless let's converge to something that will work for docs. I'd say .6 image with config-reloader bump and doing this on 0.15 release might be best for now (let's discuss in gchat), but I wonder if there's a room for special branch for docs or at least minimal templating for basics like config-reloader version.

examples/prometheus.yaml
bwplotka: Can we quickly hook it with /charts while we are here?

examples/prometheus.yaml
```diff
@@ -108,7 +108,7 @@ spec:
           privileged: false
       containers:
       - name: prometheus
-        image: gke.gcr.io/prometheus-engine/prometheus:v2.45.3-gmp.7-gke.0
+        image: gke.gcr.io/prometheus-engine/prometheus:v2.53.5-gmp.0-gke.8
```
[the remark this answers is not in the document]
hsmatulis: Bumped

PR #1801, opened 2025-11-10
Split the config maps based on size while being created. So that many config maps are created with max size of 950KB and rule evaluator loads all configmaps to file for rules generator

balakumarpg:
**Why:**
Since the k8s configmaps have hard limit of 1MB, when there is a need of creating more GlobalRules, ClusterRules or Rules, all are going into one ConfigMap by rule-generator and from there it is being mounted as volume to rule-evaluator. Which brings the hard limit of alert definitions in one GKE cluster which is enabled the GMP can not go beyond 1 MB size.

This change will overcome that hard limit.

### DIFF
--- a/charts/rule-evaluator/templates/deployment.yaml
+++ b/charts/rule-evaluator/templates/deployment.yaml
@@ -111,8 +111,21 @@ spec:
       - name: config-out
         emptyDir: {}
       - name: rules
-        configMap:
-          name: rules
+        # Mount exactly one ConfigMap per rule type (3 total)
+        # - rules: namespace-scoped Rules
+        # - clusterrules: cluster-scoped ClusterRules
+        # - globalrules: GlobalRules
+        projected:
+          sources:
+          - configMap:
+              name: rules
+              optional: true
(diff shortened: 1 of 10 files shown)


balakumarpg:
> Thanks!
> 
> So this makes sense on operator side, but how rule-eval should consume this? rule-eval currently only reads one config map and there's no easy way (or is there?) to tell rule-val pod to load dynamic number of those files? 🤔
> 
> We can't change deployment dynamically in practice within the security constraints we need to work with for managed GMP solution at the moment. That would be the only solution, right?

Thanks for the valuable input. Please check now, I have made some changes with respect to your comments.

balakumarpg:
> Also before we add some complexity, have you tried compression option? [`main`/doc/api.md#monitoring.googleapis.com/v1.ConfigSpec](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/doc/api.md#monitoring.googleapis.com/v1.ConfigSpec)

Yes, after compression it is 1.3 MB, only the GlobalRules.

PR #1815, opened 2025-12-02
refactor: bump golangci-lint to 2.6.2

bernot-dev:
Bump golangci-lint to 2.6.2.

The removals of `omitempty` in json struct tags have been verified by [modernize](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_omitzero) to be ineffective. `omitzero` was not added to avoid behavior changes. Corresponding fields have been explicitly marked optional. This results in some updates to the documentation, but no changes to the CRDs themselves.

Instances of `interface{}` have been changed to `any`.

String formatting functions have been replaced when they do not use actual formatting placeholders.

[`TestMain` implicitly calls `os.Exit` since 1.15](https://github.com/golang/go/issues/34129).

The pattern of splitting and then iterating over the parts has been replaced with `SplitSeq`, which is more efficient.


pkg/operator/apis/monitoring/v1/common_types.go
```diff
@@ -104,6 +104,8 @@ func relabelingsForSelector(selector metav1.LabelSelector, crd interface{}) ([]*
 				SourceLabels: prommodel.LabelNames{objectLabelPresent + sanitizeLabelName(exp.Key)},
 				Regex:        relabel.MustNewRegexp("true"),
 			})
+		default:
```
bwplotka: TODO for myself (or anyone else): Double check if this is for sure all cases in op.Operator.

pkg/operator/apis/monitoring/v1/monitoring_types.go
```diff
@@ -44,10 +44,10 @@ type MonitoringCondition struct {
 	Status corev1.ConditionStatus `json:"status"`
 	// The last time this condition was updated.
 	// +optional
-	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
+	LastUpdateTime metav1.Time `json:"lastUpdateTime"`
```
bwplotka: Hm, I think those are not safe to remove. Modernize is doing this assuming we use std JSON encoder/decoder but I wonder if that's true (we use yaml.v3, same k8s).

bwplotka (changes requested):
Thanks! Majority changes make sense, just few suggestions.

> The removals of omitempty in json struct tags have been verified by [modernize](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_omitzero) to be ineffective. omitzero was not added to avoid behavior changes. Corresponding fields have been explicitly marked optional. This results in some updates to the documentation, but no changes to the CRDs themselves.

Hm, are we sure about the modernize check? This is true for std json, but we don't use std json decoding but that yaml.v3 which does have a unique omitempty-almost-like-omitzero behaviour - https://pkg.go.dev/gopkg.in/yaml.v3#section-readme:~:text=Only%20include%20the%20field%20if%20it%27s%20not%20set%20to%20the%20zero

Notably zero value is omitted in our yaml parsing (also what k8s is using) (!), so I think this can potentially break things. Can we ignore that modernize rule just to be safe (unless we have time to test this thoroughly)?

PR #1821, opened 2025-12-09
fix: add builtinassets to frontend

bernot-dev:
The builtinassets were accidentally removed in a previous commit. This change reintroduces these files so the frontend image will work.

Also fixes build issues stemming from relying on an old version of Node.js and a build image that's no longer supported. Optimized build with more selective file copying. It still takes some time in no small part because there's something like 50K files copied for `prometheus_ui`.

### DIFF
--- a/cmd/frontend/Dockerfile
+++ b/cmd/frontend/Dockerfile
@@ -15,13 +15,18 @@
 FROM --platform=$BUILDPLATFORM google-go.pkg.dev/golang:1.25.4@sha256:3f90cabe0630565f75de1f7cdb0898bf9e39ff4b52df1d1e21aff75b87e68fde AS buildbase
 
 # Compile the UI assets.
-FROM --platform=$BUILDPLATFORM gcr.io/google-appengine/nodejs AS assets
+FROM --platform=$BUILDPLATFORM us-central1-docker.pkg.dev/serverless-runtimes/google-24-full/runtimes/nodejs24@sha256:12f9b32a1afdd87221c0a6c1a907c74978f7de75902264badf2db08ddb21203c AS assets
+WORKDIR /app
 # To build the UI we need a recent node version and the go toolchain.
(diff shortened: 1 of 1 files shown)


bwplotka:
Hm, not so different if we consider:

* e2e dir (expects image)
* we do [similar on fork](https://github.com/GoogleCloudPlatform/prometheus/blob/158c6e40871e575112780a449e2972059992e678/.github/workflows/presubmit.yml#L72). You can even use e2e https://github.com/efficientgo/e2e 

I can try to write a quick test tomorrow morning to help out 🤗

cmd/frontend/Dockerfile
```diff
@@ -15,13 +15,18 @@
...
+WORKDIR /app
 # To build the UI we need a recent node version and the go toolchain.
-RUN install_node v17.9.0
+COPY --from=buildbase /usr/bin/make /usr/bin/
 COPY --from=buildbase /usr/local/go /usr/local/
 ENV PATH="/usr/local/go/bin:${PATH}"
-COPY . /app
-RUN pkg/ui/build.sh
+COPY --chown=33:33 ./third_party/prometheus_ui ./third_party/prometheus_ui
+COPY --chown=33:33 go.mod go.sum .
```
bwplotka: Why 33? Can we add comment why?

Would USER root help? https://github.com/GoogleCloudPlatform/prometheus/blob/158c6e40871e575112780a449e2972059992e678/Dockerfile.google#L21

It's an asset builder so can be root?

bwplotka (approved):
Thanks!

We need to have test otherwise we will risk having those incidents again. 

We need to also port it to release/0.17

cmd/frontend/Dockerfile
bernot-dev: 33 is the default user for the runtime image. Either way.

PR #1826, opened 2025-12-11
fix: fix release/0.12 vulnerabilities

bwplotka:
Done via script:

```
./ops/gmpctl.sh vulnfix -b release/0.12
```


hsmatulis (approved):
Thanks! Got it

PR #1829, opened 2025-12-12
build: relax linter config for maintenance mode in release/0.12

hsmatulis:
Switch golangci-lint from `enable-all: true` strategy to an allowlist one. The release/0.12 branch is strictly for dependency updates & vulnerability fixes.

The previous configuration caused unrelated build failures whenever golangci-lint added new linters (e.g., `usetesting`) or updated style rules (e.g., `revive`).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,66 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


bwplotka (approved):
Agree, LGTM

PR #1834, opened 2025-12-16
fix(configs,operator): allow regex in ScrapeEndpoint.Port

bwplotka:
Ensuring we have https://github.com/GoogleCloudPlatform/prometheus-engine/commit/e70bf7053604ce174caccfe340f3b3eec8939115 fix on main.


bernot-dev (approved):
LGTM

PR #1838, opened 2026-01-13
fix: prevent panic of rule-evaluator on shutdown

bernot-dev:
The rule-evaluator contained a bug where it would attempt to restart the rule evalutor manager during shutdown, and would reuse a done channel that had already been closed, resulting in a panic. This change signals the manager when an interrupt is received, and does not attempt to restart it.

### DIFF
--- a/cmd/rule-evaluator/main_test.go
+++ b/cmd/rule-evaluator/main_test.go
@@ -18,12 +18,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
+	"sync"
 	"testing"
 	"time"
 
+	"github.com/go-kit/log"
 	"github.com/google/go-cmp/cmp"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/version"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/promql/parser"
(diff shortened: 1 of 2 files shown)


manifests/rule-evaluator.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
```
bwplotka: Just learned that we shouldn't change this. 🙈  I was doing this all wrong too.

go/copyright#the-year

cmd/rule-evaluator/main_test.go
```diff
@@ -179,3 +190,51 @@ func TestSelect(t *testing.T) {
...
+	defer func() {
+		if r := recover(); r != nil {
+			t.Error(r)
+		}
+	}()
+	reg := prometheus.NewRegistry()
+	rulesMetrics := rules.NewGroupMetrics(reg)
+	re, err := newRuleEvaluator(t.Context(), log.NewNopLogger(), &evaluatorOptions{TargetURL: &url.URL{}}, version.Version, nopAppendable{}, notifier.NewManager(&notifier.Options{}, log.NewNopLogger()), rulesMetrics)
+	if err != nil {
+		t.Error(err)
```
bwplotka: Let's fatal, otherwise unrelated test failures are added on `newRuleEvaluator` errors.

```suggestion
		t.Fatal(err)
```

manifests/rule-evaluator.yaml
bernot-dev: Resolved #1839.

PR #1843, opened 2026-01-16
fix: gmpctl vulncheck friction

bernot-dev:
Fix `hack` directory to `ops`.

Allow `sed` instead of `gsed`, when `gsed` is not available.

Indent got thrashed a bit. I'm using `shfmt`. Is there something different that makes sense? Does it matter?


bwplotka (approved):
Thank you, LGTM!

PR #1849, opened 2026-01-22
fix: defaulting for PodMonitoring and ClusterPodMonitoring target labels metadata

bernot-dev:
Target label metadata was not being defaulted properly when a PodMonitoring/ClusterPodMonitoring did not specify at least an empty TargetLabels object. This PR fixes the bug and adds tests to cover this case.


bwplotka:
Thanks for a quick fix!

PR #1921, opened 2026-05-15
chore(deps): bump golang.org/x/net to v0.53.0

brtkwr:
Resolves the `govulncheck` failure currently affecting every PR against `main`:

> Vulnerability #1: GO-2026-4918
>     Infinite loop in HTTP/2 transport when given bad SETTINGS_MAX_FRAME_SIZE in
>     net/http/internal/http2 in golang.org/x/net
>   Module: golang.org/x/net
>     Found in: golang.org/x/net@v0.48.0
>     Fixed in: golang.org/x/net@v0.53.0
>     Example traces found:
>       #1: e2e/kube/pod.go:44:28: kube.PodLogs calls rest.Request.Stream, which eventually calls http2.Transport.NewClientConn

Bumping the indirect `golang.org/x/net` dependency to v0.53.0. `go mod tidy` after the bump also rolls forward the sibling `golang.org/x/sys`, `golang.org/x/term`, and `golang.org/x/text` modules to the versions pulled in transitively by x/net v0.53.

## Test plan

- `go build ./...` clean
- `go test ./pkg/operator/...` passes (unit tests)
- e2e tests not runnable locally without a kind cluster, but the diff is dependency-only — no behavior changes
- `govulncheck ./...` should no longer report GO-2026-4918 after this (couldn't verify locally without the tool but the version table is unambiguous)

cc @bwplotka @lyanco @bernot-dev @pintohutch — flagged from #1920 where `govulncheck` blocks unrelated work; rebasing that PR on top of this one as soon as it merges.


bwplotka (approved):
Thanks

We need to revive dependabot/renovate, got stuck it seems.

PR #1922, opened 2026-05-16
feat: shard rule-ConfigMaps to avoid etcd size limit

yama6a:
## Problem

  Rules CRDs (Rules, ClusterRules, GlobalRules) get packed into a single `rules-generated` ConfigMap. Once you have enough rules, it blows past the 1MB etcd limit and the operator starts failing. Gzip helps but only buys time.

  ## Solution

  The operator now bin-packs rule entries across N sharded ConfigMaps (`rules-generated-0`, `rules-generated-1`, ...) based on actual measured size with an 800KB threshold. The number of shards is fully dynamic, nothing hardcoded.

  On the config-reloader side, instead of mounting a single ConfigMap as a volume, a new syncer goroutine discovers shards via a label selector (`monitoring.googleapis.com/rules-shard=true`) through the K8s API and writes them to the existing emptyDir volume at `/etc/rules`. It hashes content so downstream reloads only  fire when something actually changed.

  Stale shards and the legacy `rules-generated` ConfigMap get cleaned up automatically on each reconciliation.

  ## Things worth noting

- ** E2E Tests**: I was not able to run them locally, I ran into issues running a Kind cluster on my machine. Perhaps we can run them here in CI? Don't have proof that they actually work. Unit tests pass tho.

- **Rule Evaluator**: The rule evaluator just reads yaml files from a directory, it has no idea where they came from, so sharding is invisible to it and it should work as-is without changes, regardless if it's deployed stand-alone or not.

- **Standalone rule-evaluator is untouched.** `charts/rule-evaluator/` and `manifests/rule-evaluator.yaml` don't change at all. If you deploy the rule-evaluator yourself without the operator, everything works as before - you just don't get sharding.

- **Should be a no-op upgrade.** On first reconciliation the operator creates shard-0 with the same content, deletes the old ConfigMap, and the syncer picks it up. Volume name and mount path are unchanged so nothing else needs to care.

- **RBAC got a bit wider.** The operator Role now has unrestricted ConfigMap verbs (no resourceNames) because shard names are dynamic. Feels fine since it's namespace-scoped anyway. The collector gets a new namespace-scoped Role for list/watch - the ClusterRole stays untouched. Not sure if there;s a better way to do this, but I hope this is reasonable.

- **Couldn't run the full `make presubmit`.** The frontend stuff needs amd64 and I'm on an ARM Mac - fought with it for a bit and gave up. Go side compiles and tests pass, and I regenerated `manifests/operator.yaml`, but would appreciate a second pair of eyes on whether anything is missing.

- **Might have gone overboard with tests.** There's unit tests for the sharding logic, the syncer, the predicate, plus e2e for multi-shard with both no-compression and gzip. If any of them feel like noise, happy to trim.


cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,166 @@
...
+	})
+	if err != nil {
+		return false, fmt.Errorf("list configmaps: %w", err)
+	}
+
+	files := make(map[string][]byte)
+	for i := range cmList.Items {
+		cm := &cmList.Items[i]
+		for k, v := range cm.Data {
+			files[cm.Name+"__"+k] = []byte(v)
```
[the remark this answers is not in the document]
yama6a: The ConfigMap names are `rules-generated-0`, `rules-generated-1`, etc. The keys are like `rules__default__my-rules.yaml`. Both are well under 253 chars individually and nowhere near 255 combined. We control both sides so there's no risk of user-supplied names blowing this up.

cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+func (s *ConfigMapSyncer) writeFiles(files map[string][]byte) error {
+	if err := os.MkdirAll(s.outputDir, 0o755); err != nil {
+		return err
+	}
+
+	for name, data := range files {
+		if err := os.WriteFile(filepath.Join(s.outputDir, name), data, 0o644); err != nil {
```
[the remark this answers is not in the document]
yama6a: the keys are entirely generated by the operator (see setConfigMapData in pkg/operator/rules.go) - they're always `{type}__{namespace}__{name}.yaml` or `empty.yaml`. There's no user-controlled input in the key names.

But then again, it doesn't hurt to be defensive. I added a commit to address this.

charts/operator/templates/role.yaml
```diff
@@ -68,7 +68,7 @@ rules:
 - resources:
   - configmaps
   apiGroups: [""]
-  verbs: ["list", "watch", "create"]
+  verbs: ["list", "watch", "create", "get", "patch", "update", "delete"]
```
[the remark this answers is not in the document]
yama6a: Technically redundant, but it acts as a safety net - if the broader rule gets accidentally removed in a future refactor, the operator still has access to the three fixed ConfigMaps. I can remove the redundant named ones if you want, tho. Up to you.

cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,166 @@
...
+		case <-ticker.C:
+			if _, err := s.Sync(ctx); err != nil {
+				//nolint:errcheck
+				level.Warn(s.logger).Log("msg", "configmap sync failed", "err", err)
+			}
+		}
+	}
+}
+
+func (s *ConfigMapSyncer) writeFiles(files map[string][]byte) error {
```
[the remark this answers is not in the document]
yama6a: This is fine in practice - the config-reloader already hashes the directory on each poll and only triggers a reload when the hash changes. If it catches a mid-write state, it just picks it up on the next cycle. Not worth the complexity of an atomic swap imho.

bernot-dev:
Hi @yama6a. Thanks for the PR. This is definitely an interesting feature.

I approved the workflow to run, and it looks like the new e2e test is not passing, so the other ones got cancelled. We also use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) standard for commit message. Running `make conform` will tell you if the commit messages are compliant.

The e2e/Kind tests are a bit tricky, because they've really be built around what's convenient for our small team of core developers. Sorry to hear they weren't working for you. It's somewhere on my list to make that barrier to entry a bit lower. Feel free to push your changes and I'll approve the workflow to run here in CI.

From a release planning perspective, I would rather not delete the old configmap in the same update as the shards are introduced. That way, if we need to do a rollback for any reason, we're a bit less likely to be broken. At first glance, it seems like they may not conflict if the old configmap doesn't have the `monitoring.googleapis.com/rules-shard=true` label. However, if you're actually hitting the limit, that may not be helpful. Thoughts on this?

Thanks again for this contribution. We'll do our best to make sure you get timely reviews.

yama6a:
> Hi @yama6a. Thanks for the PR. This is definitely an interesting feature.
> 
> I approved the workflow to run, and it looks like the new e2e test is not passing, so the other ones got cancelled. We also use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) standard for commit message. Running `make conform` will tell you if the commit messages are compliant.
> 
> The e2e/Kind tests are a bit tricky, because they've really be built around what's convenient for our small team of core developers. Sorry to hear they weren't working for you. It's somewhere on my list to make that barrier to entry a bit lower. Feel free to push your changes and I'll approve the workflow to run here in CI.
> 
> From a release planning perspective, I would rather not delete the old configmap in the same update as the shards are introduced. That way, if we need to do a rollback for any reason, we're a bit less likely to be broken. At first glance, it seems like they may not conflict if the old configmap doesn't have the `monitoring.googleapis.com/rules-shard=true` label. However, if you're actually hitting the limit, that may not be helpful. Thoughts on this?
> 
> Thanks again for this contribution. We'll do our best to make sure you get timely reviews.

@bernot-dev 
Good catch, pushed a change to keep the legacy rules-generated ConfigMap around for this release.

The old rule-evaluator deployment mounts it by name, so deleting it here would briefly crashloop the rule-evaluator on rollback, until the old operator recreates the CM.

Can be dropped (along with the RBAC resourceNames entry) in a follow-up release once this one has been out for a while.

I pushed changes to make it conform as well. Can you enable auto-approval for the e2e-test-runs? I think we did that in some of my previous PRs to this project, so I get a quicker feedback loop to fix the e2e tests.

cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,169 @@
+// Copyright 2024 Google LLC
```
bwplotka: nit: 

```suggestion
// Copyright 2026 Google LLC
```

cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,169 @@
...
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+type ConfigMapSyncer struct {
```
bwplotka, line 45: nit: A light commentary on a public structs would be great

cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,169 @@
...
+		client:    client,
+		namespace: namespace,
+		selector:  selector,
+		outputDir: outputDir,
+		interval:  interval,
+		logger:    logger,
+	}
+}
+
+func (s *ConfigMapSyncer) Sync(ctx context.Context) (bool, error) {
```
bwplotka, line 84: ditto (commentary)

cmd/config-reloader/internal/syncer.go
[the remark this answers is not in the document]
bwplotka: What about side-injections? Somebody creates out-of band recording rules into the namespace, exceeding the rule.

We could add an easy validation here to ensure we ignore such a rule, little harm in adding it.

Low risk I agree, but one could design their self-deploy GMP in this way. WDYT?

cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,169 @@
...
+			files[cm.Name+"__"+k] = v
+		}
+	}
+
+	hash := hashFiles(files)
+	if hash == s.lastHash {
+		return false, nil
+	}
+
+	if err := s.writeFiles(files); err != nil {
```
bwplotka, line 121: Nice, this logic is begging for rule-evaluator sharding one day (: 

Could be extended one day 👍🏽

cmd/config-reloader/internal/syncer.go
```diff
@@ -0,0 +1,169 @@
...
+		return false, err
+	}
+
+	s.lastHash = hash
+	//nolint:errcheck
```
bwplotka: Do you mind adding to https://github.com/GoogleCloudPlatform/prometheus-engine/blob/c2c7e38e14842aaec2110fa95b3042274ae9729e/.golangci.yml#L91 instead?

Not sure why it was not done before

cmd/config-reloader/internal/syncer.go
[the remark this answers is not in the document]
bwplotka: Again, side injection is one low risk case. Thanks for resolving

cmd/config-reloader/internal/syncer_test.go
```diff
@@ -0,0 +1,362 @@
+// Copyright 2024 Google LLC
```
bwplotka: ditto

cmd/config-reloader/internal/syncer.go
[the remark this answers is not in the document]
bwplotka: Hm, so it worked with one file, but now we have many. Are we sure rule-eval will not load inconsistent state of rule groups, especially during 1->3 and 3->1 e.g. causing duplicate groups or key recording rule missing? What am I missing?

One alternative is writing to a big file locally (: Another is integrating K8s API into Thanos config-reloader (we could have our copy of it, it's not a lot of code). Symlink is odd but might work.

cmd/config-reloader/README.md
```diff
@@ -9,6 +9,10 @@ Meant to be run as a sidecar.
 Usage of config-reloader:
   -config-dir string
     	config directory to watch for changes
+  -config-dir-from-configmap-namespace string
+    	namespace to list ConfigMaps from (required when --config-dir-from-configmap-selector is set)
+  -config-dir-from-configmap-selector string
+    	label selector to discover rule ConfigMaps via K8s API (e.g. monitoring.googleapis.com/rules-shard=true). When set, replaces --config-dir for rule file discovery.
```
bwplotka: Except this is config-reloader that is also used in other places e.g. collector config reloading.

We likely need to fix help of this flag to be more generic 🤔

manifests/operator.yaml
```diff
@@ -247,6 +247,18 @@ rules:
   apiGroups: ["monitoring.googleapis.com"]
   verbs: ["get", "update", "list", "watch"]
 ---
+# Source: operator/templates/role.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: collector
```
bwplotka, line 254: We probably should avoid touching collector here and switching recording rule-eval only. Should it be rule-eval?

Sharding Prometheus config would be also be interesting and possible, but switching nodes to use more K8s API is tricky.

manifests/operator.yaml
```diff
@@ -276,6 +288,20 @@ subjects:
 - name: operator
   kind: ServiceAccount
 ---
+# Source: operator/templates/rolebinding.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: collector
```
bwplotka, line 295: ditto, did you mean rule-eval?

bwplotka (commented):
Great job! Some initial thoughts, thanks!

bwplotka:
I changed the workflows so you don't need approve.

cmd/config-reloader/internal/syncer.go
yama6a: done

cmd/config-reloader/internal/syncer.go
yama6a: done

cmd/config-reloader/internal/syncer.go
yama6a: done

cmd/config-reloader/internal/syncer_test.go
yama6a: done

cmd/rule-evaluator/main.go
```diff
@@ -1054,6 +1054,28 @@ func (e *ruleEvaluator) Stop() {
...
+func (e *ruleEvaluator) AlertingRules() []*rules.AlertingRule {
+	e.mtx.Lock()
+	curr := e.rulesManager
+	e.mtx.Unlock()
+	if curr == nil {
+		return nil
+	}
+	return curr.AlertingRules()
+}
+
```
yama6a, line 1078: `NewAPI()` captured the rulesManager pointer at startup, so after ApplyConfig swapped managers the API kept reading the stopped one. Pass *ruleEvaluator and dereference e.rulesManager per call as a fix.

cmd/config-reloader/internal/syncer.go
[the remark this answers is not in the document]
yama6a: Fair enough, Iadded a namePrefix check in the syncer. only ConfigMaps starting with rules-generated- are materialized. main.go wires it up with that prefix, but the syncer itself stays generic (empty prefix disables the check). Also added a test for the rogue-CM case.

cmd/config-reloader/internal/syncer.go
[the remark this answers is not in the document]
yama6a: True, fair point. There's a real window when going between one and multiple shards, where we can briefly see duplicates or gaps, but it self-heals on the next sync within about 10 seconds.

Having one merged file would fix the atomicity thingy, but it adds a global group-name uniqueness constraint that we don't have today, not sure we wanna deal with that.

If a few seconds out-of-sync'ness isn't acceptable, I would propose to implement the same ..data symlink swap that K8s does for configmap mounts. Then no rule-eval changes would be needed then. It seems like the least invasive option to me. What do you think? Can we live with the self-healing?

cmd/config-reloader/internal/syncer.go
yama6a: done

cmd/config-reloader/README.md
yama6a: yup, fixed

manifests/operator.yaml
yama6a: Hm, I see what you mean. But looking at how it works today, rule-eval is seems to be running under the "collector" SA, and it already uses kubernetes_sd_configs for Alertmanager discovery, so its SA isn't just leaning on configmaps from the collector ClusterRole, it also needs the endpoints/services/pods reads. So a clean split would mean a parallel ClusterRole with all of those, and if we miss one, ... not so good 😅

The only thing we'd actually be taking away from the collector SA is namespaced configmap list/watch (it already has cluster-wide configmap get anyway). Feels like a rather risky/invasive change. I can give it a shot if you want, tho. Your call, what do you think?

yama6a:
@bwplotka @bernot-dev 
Could you take a peek at my two remaining comments:

https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1922#discussion_r3274668044

And 

https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1922#discussion_r3274340632

If you let me know how you'd like me to move forward, I can work on the remaining open things next week some time.

yama6a:
added the atomicity fix now. should be good to go.

PR #1937, opened 2026-06-01
[DO NOT MERGE] test

bwplotka:
Ensuring no false positives for b/513253808


bwplotka:
Confirmed we have test against namespace 

image

PR #1938, opened 2026-06-01
fix: gmpctl hardening

bwplotka:
A few fixes.

Includes:
- Only set GPG_TTY if a valid TTY was successfully detected.
- Optimize directory walks with filepath.WalkDir in cmd_vulnfix.go.
- Forward stderr for temp schema run in cmd_vulnfix.go.
- Use idiomatic string comparison for branch == "main".
- Use find -prune to efficiently skip large directories in lib.sh.
- Add CVSSMetricV30 fallback to NVD API severity parsing in nvdapi.go.
- Enable `main` branch vulnfix.
- Dependency, tagging, cache, and semconv sync fixes.
- Fix documentation format (CI was adding +go downloading on a cold cache)
- Pin TOOLCHAIN to Go 1.25 for consistent Go version behavior, while allowing 1.26 (future) on images

### DIFF
--- a/ops/gmpctl/cmd_vulnfix.go
+++ b/ops/gmpctl/cmd_vulnfix.go
@@ -30,7 +30,7 @@ var (
 	vulnfixBranch              = vulnfixFlags.String("b", "", "Release branch to work on; Project is auto-detected from this")
 	vulnfixPRBranch            = vulnfixFlags.String("pr-branch", "", "(default: $USER/BRANCH-vulnfix) Upstream branch to push to (user-confirmed first).")
 	vulnfixSyncDockerfilesFrom = vulnfixFlags.Bool("sync-dockerfiles-from", false, "Optional branch name to sync Dockerfiles from. Useful when things changed.")
-	vulnfixGoVersion           = vulnfixFlags.String("go-version", "", "Go minor version to use for toolchain and docker images.")
+	vulnfixGoVersion           = vulnfixFlags.String("go-version", "", "Go minor version to use for docker images.")
 )
 
 // Attempt a minimal dependency upgrade to solve fixable vulnerabilities.
--- a/ops/gmpctl/cmd_vulnfix.go
+++ b/ops/gmpctl/cmd_vulnfix.go
@@ -19,6 +19,10 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
 )
 
 var (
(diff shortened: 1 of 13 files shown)


ops/gmpctl/lib.sh
```diff
@@ -442,17 +440,22 @@ release-lib::manifests_regen() {
...
-	# TODO(bwplotka): Manage deps better. It's getting confusing what bins we should use (worktree bingo? script bingo?).
-	go install helm.sh/helm/v3/cmd/helm@latest
-	go install github.com/google/addlicense@latest
-	go install github.com/mikefarah/yq/v4@latest
-
 	# Hack: Do the bingo variable swap. This allows injecting our own.
 	# This is faster than running requiring bingo and running bingo get.
-	cp "${dir}/.bingo/variables.env" "${dir}/.bingo/variables.env.bak"
-	echo "#!/bin/bash" >"${dir}/.bingo/variables.env" # Clean the file.
+	# NOTE: Only needed before 0.19.
```
bwplotka, line 445: cc

ops/gmpctl/cmd_vulnfix.go
```diff
@@ -74,21 +79,46 @@ func vulnfix() error {
...
+		// This makes it simpler to maintain dependencies across old versions, forks and tools (e.g. code gen).
+		// This follows what e.g. Prometheus is doing https://github.com/prometheus/prometheus/pull/18938#issue-4676291443
+		fmt.Sprintf("GOTOOLCHAIN=go1.25.0"),
 	}
 	if *vulnfixSyncDockerfilesFrom {
 		opts = append(opts, "SYNC_DOCKERFILES_FROM=true")
 	}
+	// Update go version in go.mod to what toolchain is set to if it was updated by accident
+	// otherwise it won't work with our toolchain.
+	if _, err := runCommand(&cmdOpts{Dir: dir, Envs: opts}, "go", "mod", "edit", "-go=1.25.0"); err != nil {
```
bernot-dev, line 106: This is incompatible with `GOTOOLCHAIN=local`, which is applied to ensure we're actually building with the toolchain from the build image.

Also, we do not want to lock in to a specific patch version. For 1.25, latest is 1.25.11, and we do not want to introduce new vulnerabilities from the toolchain because we're locked into `.0`.

bernot-dev (changes requested):

ops/gmpctl/cmd_vulnfix.go
```diff
@@ -74,21 +79,46 @@ func vulnfix() error {
...
+
 	opts := []string{
 		fmt.Sprintf("DIR=%v", dir),
 		fmt.Sprintf("BRANCH=%v", branch),
 		fmt.Sprintf("PROJECT=%v", proj.Name),
+		fmt.Sprintf("GO_VERSION=%v", goVersion),
+		// We are hardcoding toolchain everywhere for now, until we have deps that require higher version.
+		// This makes it simpler to maintain dependencies across old versions, forks and tools (e.g. code gen).
+		// This follows what e.g. Prometheus is doing https://github.com/prometheus/prometheus/pull/18938#issue-4676291443
+		fmt.Sprintf("GOTOOLCHAIN=go1.25.0"),
```
bernot-dev, line 99: `GOTOOLCHAIN` feels like the wrong tool to prevent dependency updates.

ops/gmpctl/cmd_vulnfix.go
bwplotka: Unfortunately that also means it depends what Go version you have installed etc.

bwplotka:
Newer version of the same, without toolchain https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1971

PR #1944, opened 2026-06-03
fix: fix main vulnerabilities

bwplotka:
```
./gmpctl.sh vulnfix
```


bwplotka:
govulncheck seems to fail because Go action has an old Go version (actions/setup-go@v5.0.0)

bernot-dev (approved):

PR #1955, opened 2026-06-09
Remove duplicated ksm metrics when self-deployed ksm clusterpodmonitor has duplicated entries as the gke managed one

CatherineF-dev:
Remove duplicated ksm metrics when self-deployed ksm clusterpodmonitor has duplicated entries as the gke managed one

### DIFF
--- a/pkg/operator/collection.go
+++ b/pkg/operator/collection.go
@@ -409,6 +409,8 @@ func (r *collectionReconciler) makeCollectorConfig(ctx context.Context, spec *mo
 		return nil, nil, fmt.Errorf("failed to list ClusterPodMonitorings: %w", err)
 	}
 
+	clusterPodMons.Items = deduplicateKSM(clusterPodMons.Items)
+
 	// Mark status updates in batch with single timestamp.
 	for _, cmon := range clusterPodMons.Items {
 		cond := &monitoringv1.MonitoringCondition{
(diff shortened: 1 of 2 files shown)


pkg/operator/collection.go
```diff
@@ -505,3 +507,114 @@ func makeRemoteWriteConfig(exports []monitoringv1.ExportSpec) ([]*promconfig.Rem
...
+					if r.Action == "keep" && len(r.SourceLabels) == 1 && r.SourceLabels[0] == "__name__" {
+						gkeKeepRules = append(gkeKeepRules, r)
+					}
+				}
+
+				if len(gkeKeepRules) > 0 {
+					// GKE has allowlist. Drop these metrics from Customer KSM.
+					epCopy := custEp.DeepCopy()
+					for _, kr := range gkeKeepRules {
+						epCopy.MetricRelabeling = append(epCopy.MetricRelabeling, monitoringv1.RelabelingRule{
```
bwplotka, line 569: We can't modify user configs at the moment

pkg/operator/collection_test.go
```diff
@@ -548,3 +548,333 @@ func TestSetConfigMapData(t *testing.T) {
...
 	}
 }
+
+func TestDeduplicateKSM(t *testing.T) {
+	systemLabels := map[string]string{
+		"app.kubernetes.io/part-of": "google-cloud-managed-prometheus",
+		"app.kubernetes.io/name":    "gke-managed-kube-state-metrics",
+	}
+	customerLabels := map[string]string{
+		"app.kubernetes.io/name": "self-deployed-kube-state-metrics",
```
bwplotka, line 558: I guess the assumption is that user would somehow set this label?

PR #1963, opened 2026-06-18
chore(gmpctl): remove install of bingo deps

bernot-dev:
#1861 removed Bingo. This removes the dependency on Bingo from gmpctl.

### DIFF
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -442,17 +442,7 @@ release-lib::manifests_regen() {
 		return 1
 	fi
 
-	# TODO(bwplotka): Manage deps better. It's getting confusing what bins we should use (worktree bingo? script bingo?).
-	go install helm.sh/helm/v3/cmd/helm@latest
-	go install github.com/google/addlicense@latest
-	go install github.com/mikefarah/yq/v4@latest
-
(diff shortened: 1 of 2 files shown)


ops/gmpctl/lib.sh
```diff
@@ -442,17 +442,7 @@ release-lib::manifests_regen() {
...
-	go install github.com/google/addlicense@latest
-	go install github.com/mikefarah/yq/v4@latest
-
-	# Hack: Do the bingo variable swap. This allows injecting our own.
-	# This is faster than running requiring bingo and running bingo get.
-	cp "${dir}/.bingo/variables.env" "${dir}/.bingo/variables.env.bak"
-	echo "#!/bin/bash" >"${dir}/.bingo/variables.env" # Clean the file.
-	YQ="$(which yq)" HELM="$(which helm)" ADDLICENSE="$(which addlicense)" bash "${dir}/hack/presubmit.sh" manifests
-	cp "${dir}/.bingo/variables.env.bak" "${dir}/.bingo/variables.env"
+	bash "${dir}/hack/presubmit.sh" manifests
```
bwplotka, line 445: Unfortunately this won't work across all versions - it only makes it work for 0.19/main.

This could make it work for now: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1938/changes#r3441808186

PR #1967, opened 2026-06-23
fix: fix main vulnerabilities

bwplotka:
```
./gmpctl.sh vulnfix
```

### DIFF
--- a/.github/workflows/check-images.yml
+++ b/.github/workflows/check-images.yml
@@ -30,5 +30,7 @@ jobs:
       with:
         persist-credentials: false
     - uses: actions/setup-go@v6
+      with:
+        go-version: '1.25'
     - name: Check manifest images
       run: make check-images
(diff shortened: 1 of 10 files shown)


.github/workflows/check-images.yml
```diff
@@ -30,5 +30,7 @@ jobs:
       with:
         persist-credentials: false
     - uses: actions/setup-go@v6
+      with:
+        go-version: '1.25'
```
bernot-dev: Why would we want to do this?

bwplotka: https://github.com/GoogleCloudPlatform/prometheus-engine/actions/runs/28031355288/job/82972613566 - there might be better way to do this

bernot-dev: I think `go-version: stable` should use the latest stable release.

https://github.com/actions/setup-go#supported-version-syntax

PR #1971, opened 2026-06-26
gmpctl hardening

bwplotka:
See commits for the logical change and rationales.

### DIFF
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -167,7 +167,7 @@ release-lib::gomod_vulnfix() {
 		fi
 
 		echo "🔄 Updating module '${mod_path}' to version '${desired_version}'..."
-		${SED} -i "s|\(	${mod_path} \).*|\1${desired_version}|" "${dir}/go.mod"
+		${SED} -i "s|\(${mod_path}[ 	][ 	]*\)v[^ 	]*|\1${desired_version}|g" "${dir}/go.mod"
 	done <"${vuln_file}"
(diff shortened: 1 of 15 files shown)


bwplotka:
It's an AI mess at this point, we have to be careful: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1999 -- to clean up

ops/gmpctl/lib.sh
```diff
@@ -442,17 +459,22 @@ release-lib::manifests_regen() {
...
-	# TODO(bwplotka): Manage deps better. It's getting confusing what bins we should use (worktree bingo? script bingo?).
-	go install helm.sh/helm/v3/cmd/helm@latest
-	go install github.com/google/addlicense@latest
-	go install github.com/mikefarah/yq/v4@latest
-
 	# Hack: Do the bingo variable swap. This allows injecting our own.
 	# This is faster than running requiring bingo and running bingo get.
-	cp "${dir}/.bingo/variables.env" "${dir}/.bingo/variables.env.bak"
-	echo "#!/bin/bash" >"${dir}/.bingo/variables.env" # Clean the file.
+	# NOTE: Only needed before 0.19.
```
bwplotka, line 464: This is what's needed on main gmpctl so it does not pollute and actually install tools.

cc @quentinmit

bwplotka: I can pull it in in alone change, but there's bit more to fix (distroless is now giving up -linux suffixes, etc)

bwplotka:
Splitting things that makes sense to smaller PRs.

PR #1972, opened 2026-06-26
ci: use stable alias for setup-go action

bernot-dev:
Use stable alias for setup-go instead of pinned/cached version.


cmd/rule-evaluator/internal/api.go
```diff
@@ -137,15 +137,22 @@ func alertsToAPIAlerts(alerts []*rules.Alert) []*apiv1.Alert {
 		})
 	}
 	// Sort for testability.
-	sort.Slice(apiAlerts, func(i, j int) bool {
-		a, b := apiAlerts[i].Labels.Hash(), apiAlerts[j].Labels.Hash()
-		if a == b {
-			a, b = apiAlerts[i].Annotations.Hash(), apiAlerts[j].Annotations.Hash()
+	slices.SortFunc(apiAlerts, func(a, b *apiv1.Alert) int {
```
bwplotka, line 140: Thanks - next time, let's just annotate in the commit/PR why we change other files than CI (otherwise it looks like incidental)

PR #1973, opened 2026-06-26
Update versions on all binaries

bernot-dev:
- Bumps the release version to `0.19.0` and the corresponding component
  image tags to `v0.19.0-gke.0` in `charts/values.global.yaml`.
- Regenerates all Kubernetes manifests (`manifests/operator.yaml`,
  `manifests/rule-evaluator.yaml`, `cmd/datasource-syncer/datasource-syncer.yaml`)
  to propagate the new version.
- Updates `ops/gmpctl/prep-rc.sh` to automatically align the component
  image tags (`configReloader`, `operator`, `ruleEvaluator`, `datasourceSyncer`)
  with the release version in `values.global.yaml`.

### DIFF
--- a/charts/values.global.yaml
+++ b/charts/values.global.yaml
@@ -16,7 +16,7 @@ commonLabels: false
 namespace:
   public: gmp-public
   system: gmp-system
-version: 0.17.3
+version: 0.19.0
 images:
   # NOTE: All tags have to be quoted otherwise they might be treated as a number.
   bash:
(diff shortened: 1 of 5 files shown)


charts/values.global.yaml
```diff
@@ -16,7 +16,7 @@ commonLabels: false
 namespace:
   public: gmp-public
   system: gmp-system
-version: 0.17.3
+version: 0.19.0
```
bwplotka, line 19: We don't have 0.19 yet, do we? (we can't forward pull images, unfortunately)

bernot-dev:
This bump will happen later, ideally after v3.13 fork release. Closing this for now.

PR #1974, opened 2026-06-26
PO to GMP Migration Tool: PodMonitor Converter with Metadata/Namespace Conversion

karthunni:
This PR implements the initial metadata and namespace migration logic for Prometheus Operator `PodMonitor` resources to GMP resources.

## Key Changes
* Implemented the `ResourceConverter` interface for `PodMonitor`.
* Evaluate `namespaceSelector` to route resources to their correct GMP equivalents (Cluster-Scoped, Multi-Namespace Split, Local Scoping)
* Changed successful migrations to log at the `Info` level using a `migration_status: "success"` attribute to accommodate for production environments where `Debug` logging is disabled
* Added reusable helpers for building `ObjectMeta` and `TypeMeta`.
* Added `ParseAndCleanNamespaces` for trimming and deduplicating namespace lists.
* Added corresponding test suite


pkg/migrate/podmonitor.go
```diff
@@ -0,0 +1,148 @@
...
+	return []*unstructured.Unstructured{u}, nil
+}
+
+func (c *PodMonitorConverter) convertToPodMonitoring(pm *monitoringv1.PodMonitor, logger *slog.Logger) (*unstructured.Unstructured, error) {
+	gmpPM := &gmpv1.PodMonitoring{
+		TypeMeta:   BuildTypeMeta(KindPodMonitoring),
+		ObjectMeta: BuildObjectMeta(pm.Name, pm.Namespace),
+		Spec: gmpv1.PodMonitoringSpec{
+			Selector: pm.Spec.Selector,
+		},
```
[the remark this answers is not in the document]
karthunni: TODO added. Endpoint logic to be added in follow up PR

pkg/migrate/migrate.go
```diff
@@ -348,7 +348,7 @@ func (m *Migrator) convertResources() []*unstructured.Unstructured {
 
 			allOutputs = append(allOutputs, outputs...)
 
-			resourceLogger.Debug("Converted successfully")
+			resourceLogger.Info("Converted successfully", slog.String("migration_status", "success"))
```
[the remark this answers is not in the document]
karthunni: I think it'd be better to throw an error in this case

pkg/migrate/podmonitor.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonitor) (*unstructured.Unstructured, error) {
+	gmpPM := &monitoringv1.PodMonitoring{
+		TypeMeta:   BuildTypeMeta(KindPodMonitoring),
+		ObjectMeta: CopyObjectMeta(pm.ObjectMeta, pm.Namespace),
+		Spec: monitoringv1.PodMonitoringSpec{
+			Selector: pm.Spec.Selector,
+			// TODO: Migrate pm.Spec.PodMetricsEndpoints to Spec.Endpoints in subsequent step.
+		},
```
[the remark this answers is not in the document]
karthunni: jobLabel and PodTargetLabels conversion will be added in upcoming PRs

pkg/migrate/podmonitor.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.PodMonitor) (*unstructured.Unstructured, error) {
+	gmpCPM := &monitoringv1.ClusterPodMonitoring{
+		TypeMeta:   BuildTypeMeta(KindClusterPodMonitoring),
+		ObjectMeta: CopyObjectMeta(pm.ObjectMeta, ""), // Cluster-scoped, namespace is omitted
+		Spec: monitoringv1.ClusterPodMonitoringSpec{
+			Selector: pm.Spec.Selector,
+			// TODO: Migrate pm.Spec.PodMetricsEndpoints to Spec.Endpoints in subsequent step.
+		},
```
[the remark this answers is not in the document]
karthunni: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1974#discussion_r3493015303

pkg/migrate/podmonitor.go
```diff
@@ -0,0 +1,147 @@
...
+	return []*unstructured.Unstructured{u}, nil
+}
+
+func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonitor) (*unstructured.Unstructured, error) {
+	gmpPM := &monitoringv1.PodMonitoring{
+		TypeMeta:   BuildTypeMeta(KindPodMonitoring),
+		ObjectMeta: CopyObjectMeta(pm.ObjectMeta, pm.Namespace),
+		Spec: monitoringv1.PodMonitoringSpec{
+			Selector: pm.Spec.Selector,
+			// TODO: Migrate pm.Spec.PodMetricsEndpoints to Spec.Endpoints in subsequent step.
```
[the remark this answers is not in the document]
karthunni: Tool is currently noted as experimental & WIP so missing functionality is expected

pkg/migrate/podmonitor.go
```diff
@@ -0,0 +1,147 @@
...
+	u.SetAPIVersion(GMPAPIVersion)
+	u.SetKind(KindPodMonitoring)
+
+	return u, nil
+}
+
+func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.PodMonitor) (*unstructured.Unstructured, error) {
+	gmpCPM := &monitoringv1.ClusterPodMonitoring{
+		TypeMeta:   BuildTypeMeta(KindClusterPodMonitoring),
+		ObjectMeta: CopyObjectMeta(pm.ObjectMeta, ""), // Cluster-scoped, namespace is omitted
```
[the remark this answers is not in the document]
karthunni: Added: Tracking generated files to throw errors when collision detected

pkg/migrate/migrate.go
```diff
@@ -346,10 +349,42 @@ func (m *Migrator) convertResources() []*unstructured.Unstructured {
...
+				srcKey, originalSrc, key)
+		}
+		verifiedKeys = append(verifiedKeys, key)
+	}
+	// Add verified keys
+	for _, key := range verifiedKeys {
+		seen[key] = srcKey
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
karthunni: This case wouldn't occur. One input resource could not output identical resources itself as we deduplicate on namespace parsing, we only need to check against resources generated by other inputs

pkg/migrate/helpers.go
```diff
@@ -0,0 +1,74 @@
...
+	dst := metav1.ObjectMeta{
+		Name:        src.Name,
+		Namespace:   targetNamespace,
+		Labels:      maps.Clone(src.Labels),
+		Annotations: maps.Clone(src.Annotations),
+	}
+
+	// Strip cluster deployment history
+	delete(dst.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+	delete(dst.Annotations, "meta.helm.sh/release-name")
```
bernot-dev: I'm not sure about removing annotations. And why just these ones?

True, Helm annotations are probably not going to remain relevant. But what about other management tools (Kustomize, Terraform, Argo, etc.)?

pkg/migrate/podmonitor_test.go
```diff
@@ -0,0 +1,297 @@
...
+	"testing"
+
+	monitoringv1 "github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator/apis/monitoring/v1"
+	pomonitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func TestPodMonitorNamespaceHarness(t *testing.T) {
```
bernot-dev: Use table driven test to make this more concise.

bernot-dev (changes requested):

pkg/migrate/helpers.go
karthunni: Keeping the delete on `kubectl.kubernetes.io/last-applied-configuration` as `kubectl apply` blindly accepts it for a new resource, and the stale config can cause issues on subsequent applies

& comment added to consider stripping deployment annotations in the future (removed it for now)

pkg/migrate/podmonitor_test.go
```diff
@@ -17,258 +17,279 @@ package migrate
...
+			}
 
-		var gmpPM monitoringv1.PodMonitoring
-		fromUnstructured(t, out, &gmpPM)
-		if gmpPM.Spec.Selector.MatchLabels["app"] != "local-app" {
-			t.Errorf("expected selector app='local-app', got %v", gmpPM.Spec.Selector.MatchLabels)
-		}
-	})
+			for i := range actual {
+				expectedVal := reflect.ValueOf(tc.expected[i])
```
bernot-dev: Can we use a type switch instead of reflection here? It will be a more readable, faster, and have better type safety.

go.mod
```diff
@@ -46,6 +46,7 @@ require (
 
 require (
 	github.com/efficientgo/e2e v0.14.1-0.20230710114240-c316eb95ae5b
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.75.0
```
bwplotka, line 49: Should we add comment or even lint that ensures we only depend on "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1" package?

Otherwise we will get into depedency hell (different versions of Prometheus etc). There's a https://golangci-lint.run/docs/linters/configuration/#depguard we could configure

bwplotka: Actually you import separate go module with reduced deps (only k8s, so perfect).

pkg/migrate/migrate.go
```diff
@@ -346,10 +349,42 @@ func (m *Migrator) convertResources() []*unstructured.Unstructured {
...
 		}
 	}
 	return allOutputs
 }
+
+// checkCollisions verifies that none of the generated outputs conflict with previously generated outputs.
+func (m *Migrator) checkCollisions(outputs []*unstructured.Unstructured, seen map[string]string, srcKind, srcNamespace, srcName string) error {
+	srcKey := fmt.Sprintf("%s/%s/%s", srcKind, srcNamespace, srcName)
+
+	// Store verified keys
```
bwplotka:
```suggestion
	// Store verified keys.
```

We need to change AGENTS.md or/and add linter to not forget about this (:

pkg/migrate/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+		Name:        src.Name,
+		Namespace:   targetNamespace,
+		Labels:      maps.Clone(src.Labels),
+		Annotations: maps.Clone(src.Annotations),
+	}
+
+	// Strip this annotation because it will contain the old PO schema.
+	delete(dst.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+
+	// Future consideration: Delete deployment specific annotations/labels (helm, ARGO, Kustomize etc.)
```
bwplotka: Should we really preserve those? Or should we reset for now? 

Another options is to add prefix.

Anyway let's document the current behaviour

pkg/migrate/migrate.go
bwplotka: Would like to help @karthunni ?

pkg/migrate/helpers.go
bwplotka: Things like below could be perhaps harmful to keep? Misleading? I'm leaning to empty, reset I think, then configurability.

```
app.kubernetes.io/instance: xyz-backend
app.kubernetes.io/managed-by: Helm
app.kubernetes.io/name: api-gateway
app.kubernetes.io/version: 0.1.0
helm.sh/chart: api-gateway-0.1.0
```

pkg/migrate/podmonitor.go
```diff
@@ -0,0 +1,147 @@
...
+	return KindPodMonitor
+}
+
+// Convert translates a Prometheus Operator PodMonitor into GMP resources.
+func (c *PodMonitorConverter) Convert(_ context.Context, logger *slog.Logger, unstruct *unstructured.Unstructured, _ *ResourceCache) ([]*unstructured.Unstructured, error) {
+	if unstruct == nil || unstruct.Object == nil {
+		return nil, errors.New("cannot convert nil or uninitialized unstructured resource")
+	}
+
+	// 1. Unmarshal unstructured to typed PodMonitor
```
bwplotka:
```suggestion
	// 1. Unmarshal unstructured to typed PodMonitor.
```
etc

bwplotka (approved):
Generally makes sense, thanks!

pkg/migrate/migrate.go
karthunni: PR created to address this:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1981

pkg/migrate/helpers.go
karthunni: I agree, seems like it's more likely to be misleading than useful. Modified this to now strip all labels and annotations

bernot-dev (approved):

PR #1980, opened 2026-07-06
fix: fix release/0.18 vulnerabilities

bwplotka:
```
./gmpctl.sh vulnfix
```


bwplotka:
Nothing to fix, ~so why our image looks vuln?~ answer: we never released 0.18.1-r.1

PR #1989, opened 2026-07-20
PO to GMP Migration Tool: Podmonitor Limits, ScrapeClass, Metadata. FilterRunning Migration

karthunni:
This change completes the missing configuration fields in the PodMonitor migration path to GKE Managed Prometheus (PodMonitoring and ClusterPodMonitoring).

Key changes:
* **Node Metadata Mapping:** Maps `attachMetadata.node` to `targetLabels.metadata: ["node"]`.
* **Limits Mapping:** Maps `sampleLimit`, `labelLimit`, `labelNameLengthLimit`, and `labelValueLengthLimit` to GMP's `ScrapeLimits`.
* **FilterRunning aggregation:** Aggregates endpoint-level `filterRunning` configurations to the resource level.
* **ScrapeClass TODO & Warnings:** Emits warning logs when a `scrapeClassName` is configured, noting that inherited settings will be lost. Added a TODO code comments to resolve and merge `ScrapeClass` settings from the Prometheus CR once that migration pipeline is implemented.

### DIFF
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -43,10 +43,8 @@ const (
 
 // Constants representing the supported ScrapeProtocol enum values defined in upstream Prometheus Operator.
 const (
-	scrapeProtocolOpenMetricsText001 = pomonitoringv1.ScrapeProtocol("OpenMetricsText0.0.1")
 	scrapeProtocolOpenMetricsText100 = pomonitoringv1.ScrapeProtocol("OpenMetricsText1.0.0")
 	scrapeProtocolPrometheusProto    = pomonitoringv1.ScrapeProtocol("PrometheusProto")
-	scrapeProtocolPrometheusText004  = pomonitoringv1.ScrapeProtocol("PrometheusText0.0.4")
 )
 
 var (
@@ -958,10 +956,6 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
 	}
 	if labelValueLengthLimit != nil {
 		limits.LabelValueLength = *labelValueLengthLimit
-	}
-	// Return nil if all limit fields are zero to avoid emitting empty limits objects in YAML.
-	if *limits == (monitoringv1.ScrapeLimits{}) {
-		return nil
 	}
 	return limits
 }
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -69,6 +69,7 @@ var (
 		"__meta_kubernetes_pod_name":            labelPod,
 		"__meta_kubernetes_pod_container_name":  labelContainer,
 		"__meta_kubernetes_pod_node_name":       labelNode,
+		"__meta_kubernetes_node_name":           labelNode,
 		"__meta_kubernetes_namespace":           export.KeyNamespace,
 		"__meta_kubernetes_pod_controller_name": labelTopLevelControllerName,
 		"__meta_kubernetes_pod_controller_kind": labelTopLevelControllerType,
@@ -759,6 +760,10 @@ func shouldSkipRelabelConfig(logger *slog.Logger, config pomonitoringv1.RelabelC
 			logger.Warn(fmt.Sprintf("Relabeling rule referencing pod annotation %q is unsupported in GMP. The rule has been dropped.", string(sl)))
 			return true
 		}
+		if strings.HasPrefix(string(sl), "__meta_kubernetes_node_label_") {
+			logger.Warn(fmt.Sprintf("Relabeling rule referencing node label %q is unsupported in GMP (only node name is supported). The rule has been dropped.", string(sl)))
+			return true
+		}
 	}
 	return false
 }
@@ -920,6 +925,19 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
 	logger.Info(fmt.Sprintf("Complex relabeling rule (target: %q) promoted from pre-scrape 'relabelings' to post-scrape 'metricRelabeling'.", data.targetLabel))
 }
 
+// warnUnsupportedSpecFields logs warnings for spec-level fields that GMP does not support or need.
+func warnUnsupportedSpecFields(logger *slog.Logger, spec pomonitoringv1.PodMonitorSpec) {
+	if spec.TargetLimit != nil {
+		logger.Warn("Field 'targetLimit' is unnecessary in GMP Managed Collection and has been dropped. Target discovery and scaling are managed automatically by GKE.")
+	}
+	if spec.KeepDroppedTargets != nil {
+		logger.Warn("Field 'keepDroppedTargets' is unnecessary in GMP Managed Collection and has been dropped.")
+	}
+	if spec.BodySizeLimit != nil {
+		logger.Warn("Field 'bodySizeLimit' is unsupported by GMP Managed Collection and has been dropped. Scrape response buffer limits are managed automatically by GMP.")
+	}
+}
+
 // convertLimits maps PodMonitor limit settings to GMP ScrapeLimits.
 func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLengthLimit *uint64) *monitoringv1.ScrapeLimits {
 	if sampleLimit == nil && labelLimit == nil && labelNameLengthLimit == nil && labelValueLengthLimit == nil {
@@ -937,6 +955,10 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
 	}
 	if labelValueLengthLimit != nil {
 		limits.LabelValueLength = *labelValueLengthLimit
+	}
+	// Return nil if all limit fields are zero to avoid emitting empty limits objects in YAML.
+	if *limits == (monitoringv1.ScrapeLimits{}) {
+		return nil
 	}
 	return limits
 }
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -758,8 +758,8 @@ func shouldSkipRelabelConfig(logger *slog.Logger, config pomonitoringv1.RelabelC
 			logger.Warn(fmt.Sprintf("Relabeling rule referencing pod annotation %q is unsupported in GMP. The rule has been dropped.", string(sl)))
 			return true
 		}
-		if strings.HasPrefix(string(sl), "__meta_kubernetes_node_label_") {
-			logger.Warn(fmt.Sprintf("Relabeling rule referencing node label %q is unsupported in GMP (only node name is supported). The rule has been dropped.", string(sl)))
+		if strings.HasPrefix(string(sl), "__meta_kubernetes_node_") && string(sl) != "__meta_kubernetes_node_name" {
+			logger.Warn(fmt.Sprintf("Relabeling rule referencing node metadata %q is unsupported in GMP (only node name is supported). The rule has been dropped.", string(sl)))
 			return true
 		}
 	}
@@ -956,6 +956,10 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
 	}
 	if labelValueLengthLimit != nil {
 		limits.LabelValueLength = *labelValueLengthLimit
+	}
+	// Return nil if all fields in limits remain zero since zero values are stripped by omitempty.
+	if *limits == (monitoringv1.ScrapeLimits{}) {
+		return nil
 	}
 	return limits
 }
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -41,6 +41,14 @@ const (
 	labelAddress                = "__address__"
 )
 
+// Constants representing the supported ScrapeProtocol enum values defined in upstream Prometheus Operator.
+const (
+	scrapeProtocolOpenMetricsText001 = pomonitoringv1.ScrapeProtocol("OpenMetricsText0.0.1")
+	scrapeProtocolOpenMetricsText100 = pomonitoringv1.ScrapeProtocol("OpenMetricsText1.0.0")
+	scrapeProtocolPrometheusProto    = pomonitoringv1.ScrapeProtocol("PrometheusProto")
+	scrapeProtocolPrometheusText004  = pomonitoringv1.ScrapeProtocol("PrometheusText0.0.4")
+)
+
 var (
 	// protectedLabels contains the list of labels that are protected by GMP and cannot
 	// be overwritten by targetLabels or relabeling rules.
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -926,7 +926,10 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
 }
 
 // warnUnsupportedSpecFields logs warnings for spec-level fields that GMP does not support or need.
-func warnUnsupportedSpecFields(logger *slog.Logger, spec pomonitoringv1.PodMonitorSpec) {
+func warnUnsupportedSpecFields(logger *slog.Logger, spec *pomonitoringv1.PodMonitorSpec) {
+	if spec == nil {
+		return
+	}
 	if spec.TargetLimit != nil {
 		logger.Warn("Field 'targetLimit' is unnecessary in GMP Managed Collection and has been dropped. Target discovery and scaling are managed automatically by GKE.")
 	}
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -911,3 +911,24 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
 	res.PromotedRules = append(res.PromotedRules, promoted)
 	logger.Info(fmt.Sprintf("Complex relabeling rule (target: %q) promoted from pre-scrape 'relabelings' to post-scrape 'metricRelabeling'.", data.targetLabel))
 }
+
+// convertLimits maps PodMonitor limit settings to GMP ScrapeLimits.
+func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLengthLimit *uint64) *monitoringv1.ScrapeLimits {
+	if sampleLimit == nil && labelLimit == nil && labelNameLengthLimit == nil && labelValueLengthLimit == nil {
+		return nil
+	}
+	limits := &monitoringv1.ScrapeLimits{}
+	if sampleLimit != nil {
+		limits.Samples = *sampleLimit
+	}
+	if labelLimit != nil {
+		limits.Labels = *labelLimit
+	}
+	if labelNameLengthLimit != nil {
+		limits.LabelNameLength = *labelNameLengthLimit
+	}
+	if labelValueLengthLimit != nil {
+		limits.LabelValueLength = *labelValueLengthLimit
+	}
+	return limits
+}
--- a/pkg/migrate/podmonitor.go
+++ b/pkg/migrate/podmonitor.go
@@ -249,10 +249,10 @@ func (c *PodMonitorConverter) convertEndpoints(
 
 		// 5. Warnings for Unsupported Fields in Endpoint.
 		if ep.FollowRedirects != nil && !*ep.FollowRedirects {
-			convCtx.logger.Warn("Field 'followRedirects: false' is unsupported by GMP Managed Collection and has been dropped. The collector will always follow redirects.")
+			convCtx.logger.Warn(fmt.Sprintf("endpoint [%d]: field 'followRedirects: false' is unsupported by GMP Managed Collection and has been dropped. The collector will always follow redirects.", i))
 		}
 		if ep.EnableHttp2 != nil && !*ep.EnableHttp2 {
-			convCtx.logger.Warn("Field 'enableHttp2: false' is unsupported by GMP Managed Collection and has been dropped. The collector will always negotiate HTTP/2 for TLS connections.")
+			convCtx.logger.Warn(fmt.Sprintf("endpoint [%d]: field 'enableHttp2: false' is unsupported by GMP Managed Collection and has been dropped. The collector will always negotiate HTTP/2 for TLS connections.", i))
 		}
 
 		if ep.HonorLabels {
@@ -296,6 +296,7 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 	}
 
 	// Spec-level warnings for unsupported fields.
+	warnUnsupportedSpecFields(logger, pm.Spec)
 	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
 	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
 		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
@@ -341,6 +342,7 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 			hasTrue = true
 		}
 	}
+	// A nil filterRunning defaults to true in the GMP operator.
 	var filterRunning *bool
 	if hasFalse {
 		falseVal := false
@@ -404,6 +406,7 @@ func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.P
 	}
 
 	// Spec-level warnings for unsupported fields.
+	warnUnsupportedSpecFields(logger, pm.Spec)
 	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
 	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
 		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
@@ -442,6 +445,7 @@ func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.P
 			hasTrue = true
 		}
 	}
+	// A nil filterRunning defaults to true in the GMP operator.
 	var filterRunning *bool
 	if hasFalse {
 		falseVal := false
--- a/pkg/migrate/podmonitor.go
+++ b/pkg/migrate/podmonitor.go
@@ -323,13 +323,20 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		}
 	}
 
-	var filterRunning *bool
+	var hasFalse, hasTrue bool
 	for _, ep := range pm.Spec.PodMetricsEndpoints {
 		if ep.FilterRunning != nil && !*ep.FilterRunning {
-			falseVal := false
-			filterRunning = &falseVal
-			logger.Warn("Endpoint-level configuration conflict detected: at least one endpoint is configured with 'filterRunning: false', but GMP only supports 'filterRunning' at the resource level. Setting 'filterRunning: false' globally on the PodMonitoring resource.")
-			break
+			hasFalse = true
+		} else {
+			hasTrue = true
+		}
+	}
+	var filterRunning *bool
+	if hasFalse {
+		falseVal := false
+		filterRunning = &falseVal
+		if hasTrue {
+			logger.Warn("Endpoint-level configuration conflict detected: some endpoints are configured with 'filterRunning: false' and others with 'true' (or default), but GMP only supports 'filterRunning' at the resource level. Setting 'filterRunning: false' globally on the PodMonitoring resource.")
 		}
 	}
 
@@ -404,13 +411,20 @@ func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.P
 		filteredMetadata = &union
 	}
 
-	var filterRunning *bool
+	var hasFalse, hasTrue bool
 	for _, ep := range pm.Spec.PodMetricsEndpoints {
 		if ep.FilterRunning != nil && !*ep.FilterRunning {
-			falseVal := false
-			filterRunning = &falseVal
-			logger.Warn("Endpoint-level configuration conflict detected: at least one endpoint is configured with 'filterRunning: false', but GMP only supports 'filterRunning' at the resource level. Setting 'filterRunning: false' globally on the ClusterPodMonitoring resource.")
-			break
+			hasFalse = true
+		} else {
+			hasTrue = true
+		}
+	}
+	var filterRunning *bool
+	if hasFalse {
+		falseVal := false
+		filterRunning = &falseVal
+		if hasTrue {
+			logger.Warn("Endpoint-level configuration conflict detected: some endpoints are configured with 'filterRunning: false' and others with 'true' (or default), but GMP only supports 'filterRunning' at the resource level. Setting 'filterRunning: false' globally on the ClusterPodMonitoring resource.")
 		}
 	}
 
--- a/pkg/migrate/podmonitor.go
+++ b/pkg/migrate/podmonitor.go
@@ -322,12 +322,15 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		}
 	}
 
+	// In GMP, Metadata: nil on a PodMonitoring defaults to emitting namespaced defaults (container, pod, etc.).
+	// When setting Metadata explicitly for AttachMetadata.Node, we must merge namespacedMetadataDefaults so that default metadata is not dropped.
 	if pm.Spec.AttachMetadata != nil && pm.Spec.AttachMetadata.Node != nil && *pm.Spec.AttachMetadata.Node {
 		if filteredMetadata == nil {
-			filteredMetadata = &[]string{labelNode}
-		} else if !slices.Contains(*filteredMetadata, labelNode) {
-			metadataCopy := append(slices.Clone(*filteredMetadata), labelNode)
-			filteredMetadata = &metadataCopy
+			union := unionMetadata([]string{labelNode}, namespacedMetadataDefaults)
+			filteredMetadata = &union
+		} else {
+			union := unionMetadata([]string{labelNode}, *filteredMetadata)
+			filteredMetadata = &union
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


pkg/migrate/helpers.go
```diff
@@ -811,3 +811,24 @@ func convertTargetLabels(logger *slog.Logger, sourceLabels []string, jobLabel st
...
+		limits.Labels = *labelLimit
+	}
+	if labelNameLengthLimit != nil {
+		limits.LabelNameLength = *labelNameLengthLimit
+	}
+	if labelValueLengthLimit != nil {
+		limits.LabelValueLength = *labelValueLengthLimit
+	}
+	return limits
+}
```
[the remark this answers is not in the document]
karthunni: Performing refactors in a follow up PR, as to not mix the changes

https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1990

pkg/migrate/podmonitor.go
```diff
@@ -347,12 +396,52 @@ func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.P
...
+
 	var filteredMetadata *[]string
 	if rules.ResourceCombined.Metadata != nil {
 		union := unionMetadata(*rules.ResourceCombined.Metadata, clusterMetadataDefaults)
 		filteredMetadata = &union
 	}
 
+	if pm.Spec.AttachMetadata != nil && pm.Spec.AttachMetadata.Node != nil && *pm.Spec.AttachMetadata.Node {
+		if filteredMetadata == nil {
+			filteredMetadata = &[]string{"node"}
```
dashpole: ClusterPodMonitoring: `attachMetadata.node: true` silently drops the `namespace` label.

In `convertToClusterPodMonitoring`, `filteredMetadata` is `nil` whenever no relabeling rule references metadata. Overwriting `nil` with `&[]string{"node"}` removes `namespace`, because `Metadata: nil` in GMP is interpreted as `[namespace]`, and `namespace` is emitted exclusively via `relabelingsForMetadata`. Seed from `clusterMetadataDefaults` instead:
```suggestion
		if filteredMetadata == nil {
			metadataCopy := append(slices.Clone(clusterMetadataDefaults), "node")
			filteredMetadata = &metadataCopy
```

pkg/migrate/podmonitor.go
```diff
@@ -285,6 +292,18 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		logger.Warn("Resulting PodMonitoring selector is empty. It will select and scrape all pods in this namespace. Verify if this is intended.")
 	}
 
+	// Spec-level warnings for unsupported fields.
+	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
+	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
+		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
+	}
+	for _, sp := range pm.Spec.ScrapeProtocols {
+		if strings.Contains(strings.ToLower(string(sp)), "protobuf") {
```
dashpole: The `scrapeProtocols` protobuf check can never fire.

`ScrapeProtocol` is an enum (`PrometheusProto`, `OpenMetricsText0.0.1`, `OpenMetricsText1.0.0`, `PrometheusText0.0.4`). None contains the substring `protobuf`—`strings.ToLower("PrometheusProto")` evaluates to `"prometheusproto"`. Compare against the enum value directly or check for `"PrometheusProto"` / `"proto"`.

pkg/migrate/podmonitor.go
```diff
@@ -347,12 +396,52 @@ func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.P
 		logger.Warn("Resulting ClusterPodMonitoring selector is empty. It will select and scrape all pods across all namespaces. Verify if this is intended.")
 	}
 
+	// Spec-level warnings for unsupported fields.
+	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
+	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
+		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the pre-scanned inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
+	}
+	for _, sp := range pm.Spec.ScrapeProtocols {
+		if strings.Contains(strings.ToLower(string(sp)), "protobuf") {
```
dashpole: Same in `convertToClusterPodMonitoring`: `strings.Contains(strings.ToLower(string(sp)), "protobuf")` can never fire for `PrometheusProto`.

pkg/migrate/podmonitor_test.go
```diff
@@ -1389,6 +1389,154 @@ func TestPodMonitorConversion(t *testing.T) {
...
+							LabelNameLength:  100,
+							LabelValueLength: 200,
+						},
+						FilterRunning: ptrTo(false),
+					},
+				},
+			},
+		},
+		{
+			name: "ScrapeProtocols protobuf warning",
```
dashpole, line 1872: No test asserts any of the new warnings, and the `scrapeProtocols` test uses invalid inputs.

The test harness logs to a testing writer and never inspects output, so warnings have zero effective coverage. Furthermore, this test passed only because it fed `"application/vnd.google.protobuf"` (an HTTP media type string) rather than the CRD enum value `"PrometheusProto"`. Consider checking warnings against a buffer and using valid enum inputs.

pkg/migrate/podmonitor_test.go
```diff
@@ -1389,6 +1389,154 @@ func TestPodMonitorConversion(t *testing.T) {
 				},
```
dashpole, line 1723: No `ClusterPodMonitoring` coverage for any new field.

All three new test cases omit `namespaceSelector`, so only `convertToPodMonitoring` executes. The duplicated cluster block—where the metadata namespace bug lives—is never exercised. Consider adding a `namespaceSelector.any: true` case covering limits, node metadata, and `filterRunning`.

pkg/migrate/podmonitor.go
```diff
@@ -285,6 +292,18 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		logger.Warn("Resulting PodMonitoring selector is empty. It will select and scrape all pods in this namespace. Verify if this is intended.")
 	}
 
+	// Spec-level warnings for unsupported fields.
```
dashpole, line 298: The duplication between `convertToPodMonitoring` and `convertToClusterPodMonitoring` is load-bearing.

The `attachMetadata.node` metadata bug exists because these ~35-line blocks were copied and only one needed different defaults. Extracting the metadata/filterRunning/limits processing into a shared helper parameterized by defaults would eliminate this class of bug.

pkg/migrate/podmonitor.go
```diff
@@ -347,12 +396,52 @@ func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.P
 		logger.Warn("Resulting ClusterPodMonitoring selector is empty. It will select and scrape all pods across all namespaces. Verify if this is intended.")
 	}
 
+	// Spec-level warnings for unsupported fields.
+	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
+	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
+		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the pre-scanned inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
```
dashpole: `scrapeClassName` warning text drifted between the copies.

Line 298 says `"not found in the inputs"`, while line 402 says `"not found in the pre-scanned inputs"`. Standardizing the phrasing across both places will keep log output consistent.

pkg/migrate/podmonitor.go
```diff
@@ -285,6 +292,18 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		logger.Warn("Resulting PodMonitoring selector is empty. It will select and scrape all pods in this namespace. Verify if this is intended.")
 	}
 
+	// Spec-level warnings for unsupported fields.
+	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
+	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
+		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
```
dashpole, line 302: The warning asserts a lookup that never happens.

`ScrapeClass %q was not found in the inputs`—but nothing consults `cache` for a `ScrapeClass`. Given `TODO(M2)`, phrase it as unsupported/unresolved rather than not-found.

pkg/migrate/podmonitor.go
```diff
@@ -248,6 +248,13 @@ func (c *PodMonitorConverter) convertEndpoints(
 		}
 
 		// 5. Warnings for Unsupported Fields in Endpoint.
```
dashpole, line 250: `targetLimit`, `keepDroppedTargets`, and `bodySizeLimit` are dropped in silence.

These are limit fields inside the PR's stated scope. GMP's `ScrapeLimits` has no equivalent, and every other unsupported field gets a warning. Consider adding explicit warning logs when these are set.

pkg/migrate/podmonitor.go
```diff
@@ -301,6 +320,34 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		}
 	}
 
+	if pm.Spec.AttachMetadata != nil && pm.Spec.AttachMetadata.Node != nil && *pm.Spec.AttachMetadata.Node {
```
dashpole, line 327: `attachMetadata.node` -> `metadata: ["node"]` is an approximation worth noting.

In prometheus-operator, `attachMetadata.node: true` attaches node discovery metadata (`__meta_kubernetes_node_label_*`, `__meta_kubernetes_node_name`) for relabeling without emitting an output label. GMP's `metadata: ["node"]` emits a `node` label from `__meta_kubernetes_pod_node_name`. It's a sensible best effort, but users relabeling from node labels lose that data without notice.

pkg/migrate/podmonitor.go
```diff
@@ -301,6 +320,34 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
...
+		if filteredMetadata == nil {
+			filteredMetadata = &[]string{"node"}
```
dashpole: `"node"` is hardcoded here and in several places. `labelNode` already exists at `helpers.go:36` and is used at `helpers.go:64`. Consider using `labelNode`.

pkg/migrate/helpers.go
```diff
@@ -892,3 +892,24 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
 	res.PromotedRules = append(res.PromotedRules, promoted)
 	logger.Info(fmt.Sprintf("Complex relabeling rule (target: %q) promoted from pre-scrape 'relabelings' to post-scrape 'metricRelabeling'.", data.targetLabel))
 }
+
+// convertLimits maps PodMonitor limit settings to GMP ScrapeLimits.
+func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLengthLimit *uint64) *monitoringv1.ScrapeLimits {
```
dashpole, line 943: `convertLimits` returns a non-nil `&ScrapeLimits{}` when all four inputs are explicit `0`, emitting `limits: {}`. Harmless as Prometheus reads 0 as unlimited, but avoidable noise in output YAML.

pkg/migrate/podmonitor.go
```diff
@@ -301,6 +320,34 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
...
+
+	var hasFalse, hasTrue bool
+	for _, ep := range pm.Spec.PodMetricsEndpoints {
+		if ep.FilterRunning != nil && !*ep.FilterRunning {
+			hasFalse = true
+		} else {
+			hasTrue = true
+		}
+	}
+	var filterRunning *bool
```
dashpole, line 346: nit: `filterRunning` stays `nil` when all endpoints are `true`, which is correct (`nil` default filters running pods in operator `pod_config.go:277`), but reads like an oversight. A quick comment explaining this default would clarify intent.

pkg/migrate/podmonitor.go
```diff
@@ -248,6 +248,13 @@ func (c *PodMonitorConverter) convertEndpoints(
 		}
 
 		// 5. Warnings for Unsupported Fields in Endpoint.
+		if ep.FollowRedirects != nil && !*ep.FollowRedirects {
```
dashpole, line 251: The two new endpoint warnings (`followRedirects`, `enableHttp2`) sit inside the per-endpoint loop and repeat verbatim per endpoint without index. Adding `endpoint` index or port name will make them actionable in multi-endpoint resources.

pkg/migrate/podmonitor.go
```diff
@@ -248,6 +248,13 @@ func (c *PodMonitorConverter) convertEndpoints(
...
+			convCtx.logger.Warn("Field 'followRedirects: false' is unsupported by GMP Managed Collection and has been dropped. The collector will always follow redirects.")
+		}
+		if ep.EnableHttp2 != nil && !*ep.EnableHttp2 {
+			convCtx.logger.Warn("Field 'enableHttp2: false' is unsupported by GMP Managed Collection and has been dropped. The collector will always negotiate HTTP/2 for TLS connections.")
+		}
```
dashpole, line 256: Nit: the blank line here splits the 'Warnings for Unsupported Fields in Endpoint' block introduced by the comment above it.

pkg/migrate/podmonitor.go
dashpole: If this is just because M2 isn't implemented, you can ignore that

pkg/migrate/podmonitor.go
dashpole: I wonder if there is a way to warn if users try to relabel from node labels...

pkg/migrate/podmonitor.go
karthunni: Addressed in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1990

pkg/migrate/podmonitor.go
karthunni: Yeah that is the case

pkg/migrate/podmonitor.go
karthunni: Added an explicit `__meta_kubernetes_node_label_` check in `shouldSkipRelabelConfig`

pkg/migrate/podmonitor.go
```diff
@@ -304,6 +325,35 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		}
 	}
 
+	if pm.Spec.AttachMetadata != nil && pm.Spec.AttachMetadata.Node != nil && *pm.Spec.AttachMetadata.Node {
+		if filteredMetadata == nil {
```
dashpole, line 328: If `filteredMetadata == nil`, setting `&[]string{labelNode}` drops default metadata labels (`container`, `pod`, etc.) in namespaced `PodMonitoring`. Consider merging defaults like in `convertToClusterPodMonitoring`:
```suggestion
		if filteredMetadata == nil {
			union := unionMetadata([]string{labelNode}, namespacedMetadataDefaults)
			filteredMetadata = &union
		} else {
			union := unionMetadata([]string{labelNode}, *filteredMetadata)
			filteredMetadata = &union
		}
```

pkg/migrate/podmonitor.go
```diff
@@ -288,6 +295,20 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
...
 
+	// Spec-level warnings for unsupported fields.
+	warnUnsupportedSpecFields(logger, pm.Spec)
+	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
+	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
+		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
+	}
+	// Check against the PrometheusProto enum value or the substring proto.
+	for _, sp := range pm.Spec.ScrapeProtocols {
+		if sp == scrapeProtocolPrometheusProto || strings.Contains(strings.ToLower(string(sp)), "proto") {
```
dashpole: `sp == scrapeProtocolPrometheusProto` already matches `PrometheusProto`. The `strings.Contains(..., "proto")` fallback is dead code and can be removed.

pkg/migrate/podmonitor.go
```diff
@@ -288,6 +295,20 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		logger.Warn("Resulting PodMonitoring selector is empty. It will select and scrape all pods in this namespace. Verify if this is intended.")
 	}
 
+	// Spec-level warnings for unsupported fields.
+	warnUnsupportedSpecFields(logger, pm.Spec)
```
dashpole: Consider moving the `scrapeClassName` and `scrapeProtocols` warning checks into `warnUnsupportedSpecFields` to avoid duplicating them across `convertToPodMonitoring` and `convertToClusterPodMonitoring`.

pkg/migrate/podmonitor.go
```diff
@@ -288,6 +295,20 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
...
+	// TODO(M2): Resolve and merge ScrapeClass configurations from Prometheus CR if scrapeClassName is specified.
+	if pm.Spec.ScrapeClassName != nil && *pm.Spec.ScrapeClassName != "" {
+		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *pm.Spec.ScrapeClassName))
```
dashpole, line 302: The warning claims `was not found in the inputs`, implying a lookup was attempted. Saying `is not yet supported` or `could not be resolved` would be clearer.

pkg/migrate/helpers.go
```diff
@@ -41,6 +41,14 @@ const (
 	labelAddress                = "__address__"
 )
 
+// Constants representing the supported ScrapeProtocol enum values defined in upstream Prometheus Operator.
+const (
+	scrapeProtocolOpenMetricsText001 = pomonitoringv1.ScrapeProtocol("OpenMetricsText0.0.1")
```
dashpole: Unused constant `scrapeProtocolOpenMetricsText001` (along with `scrapeProtocolPrometheusText004` on line 49). Consider removing them if they aren't needed.

pkg/migrate/helpers.go
```diff
@@ -901,3 +914,41 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
 	res.PromotedRules = append(res.PromotedRules, promoted)
 	logger.Info(fmt.Sprintf("Complex relabeling rule (target: %q) promoted from pre-scrape 'relabelings' to post-scrape 'metricRelabeling'.", data.targetLabel))
 }
+
+// warnUnsupportedSpecFields logs warnings for spec-level fields that GMP does not support or need.
+func warnUnsupportedSpecFields(logger *slog.Logger, spec pomonitoringv1.PodMonitorSpec) {
```
dashpole: Consider passing `spec *pomonitoringv1.PodMonitorSpec` by pointer to avoid copying the 15+ field struct on every call.
```suggestion
func warnUnsupportedSpecFields(logger *slog.Logger, spec *pomonitoringv1.PodMonitorSpec) {
```

pkg/migrate/podmonitor_test.go
```diff
@@ -1435,6 +1884,16 @@ func TestPodMonitorConversion(t *testing.T) {
 				t.Fatalf("Convert failed: %v", err)
```
dashpole, line 2140: `wantWarnings` only asserts positive matches. Adding a `dontWantWarnings []string` field to the test table would allow pinning negative cases (e.g. verifying no conflict warning is emitted when all endpoints set `filterRunning: false`).

pkg/migrate/podmonitor.go
karthunni: Plan is to add support in M2, but I could temporarily change the logs to `is not yet supported``

pkg/migrate/helpers.go
karthunni: Defined them in case they are needed for the future and/or for test cases -- but can remove them

pkg/migrate/podmonitor.go
karthunni: Addressed in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1990
https://github.com/GoogleCloudPlatform/prometheus-engine/blob/e26af0ba5102083f10d1263683ed9538afedab3b/pkg/migrate/podmonitor.go#L229

pkg/migrate/podmonitor.go
karthunni: Refactored in  https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1990

pkg/migrate/podmonitor.go
dashpole: no, thats fine.

pkg/migrate/helpers.go
dashpole: The typical Go convention is to not define anything that is unused.

pkg/migrate/podmonitor.go
```diff
@@ -304,6 +322,35 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		}
 	}
 
+	if pm.Spec.AttachMetadata != nil && pm.Spec.AttachMetadata.Node != nil && *pm.Spec.AttachMetadata.Node {
+		if filteredMetadata == nil {
+			filteredMetadata = &[]string{labelNode}
```
dashpole: When filteredMetadata is nil, setting it to &[]string{labelNode} causes GMP to drop default namespaced metadata (container, pod, etc.). Consider seeding namespacedMetadataDefaults using unionMetadata, similar to convertToClusterPodMonitoring:
```suggestion
			union := unionMetadata([]string{labelNode}, namespacedMetadataDefaults)
			filteredMetadata = &union
```

pkg/migrate/helpers.go
```diff
@@ -901,3 +914,44 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
...
+		limits.Labels = *labelLimit
+	}
+	if labelNameLengthLimit != nil {
+		limits.LabelNameLength = *labelNameLengthLimit
+	}
+	if labelValueLengthLimit != nil {
+		limits.LabelValueLength = *labelValueLengthLimit
+	}
+	// Return nil if all limit fields are zero to avoid emitting empty limits objects in YAML.
+	if *limits == (monitoringv1.ScrapeLimits{}) {
```
dashpole, line 961: If a caller explicitly passes a non-nil limit pointer set to 0, *limits == (monitoringv1.ScrapeLimits{}) evaluates to true and returns nil, dropping the explicit zero configuration. We can safely remove this check since the initial check on line 936 already handles all-nil inputs:
```suggestion
	return limits
```

pkg/migrate/podmonitor.go
```diff
@@ -304,6 +322,38 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
 		}
 	}
 
+	// In GMP, Metadata: nil on a PodMonitoring defaults to emitting namespaced defaults (container, pod, etc.).
+	// When setting Metadata explicitly for AttachMetadata.Node, we must merge namespacedMetadataDefaults so that default metadata is not dropped.
+	if pm.Spec.AttachMetadata != nil && pm.Spec.AttachMetadata.Node != nil && *pm.Spec.AttachMetadata.Node {
+		if filteredMetadata == nil {
```
dashpole, line 328: In existing test cases where `attachMetadata.node` is true, `filteredMetadata` is always nil when reaching this check. Consider adding a test case where a relabeling rule populates `filteredMetadata` beforehand so the `else` merging branch is also tested.

pkg/migrate/podmonitor_test.go
```diff
@@ -677,6 +681,61 @@ func TestPodMonitorConversion(t *testing.T) {
...
+									SourceLabels: []pomonitoringv1.LabelName{"__meta_kubernetes_node_label_kubernetes_io_hostname"},
+									TargetLabel:  "hostname",
+									Action:       "replace",
+								},
+							},
+						},
+					},
+				},
+			},
+			wantWarnings: []string{
```
dashpole, line 723: Since this test case provides two separate node label relabeling rules, checking only a generic warning substring could pass even if only one rule emitted a warning. Consider asserting on both specific label names (`__meta_kubernetes_node_label_topology_kubernetes_io_zone` and `__meta_kubernetes_node_label_kubernetes_io_hostname`) in `wantWarnings`.

pkg/migrate/helpers.go
```diff
@@ -751,6 +758,10 @@ func shouldSkipRelabelConfig(logger *slog.Logger, config pomonitoringv1.RelabelC
 			logger.Warn(fmt.Sprintf("Relabeling rule referencing pod annotation %q is unsupported in GMP. The rule has been dropped.", string(sl)))
 			return true
 		}
+		if strings.HasPrefix(string(sl), "__meta_kubernetes_node_label_") {
```
dashpole: To catch all unsupported node metadata (like node annotations, addresses, or presence labels) rather than just node labels, consider checking for the general `__meta_kubernetes_node_` prefix instead:
```suggestion
		if strings.HasPrefix(string(sl), "__meta_kubernetes_node_") && string(sl) != "__meta_kubernetes_node_name" {
```

pkg/migrate/helpers.go
```diff
@@ -911,3 +922,40 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
...
+	if labelLimit != nil {
+		limits.Labels = *labelLimit
+	}
+	if labelNameLengthLimit != nil {
+		limits.LabelNameLength = *labelNameLengthLimit
+	}
+	if labelValueLengthLimit != nil {
+		limits.LabelValueLength = *labelValueLengthLimit
+	}
+	return limits
```
dashpole, line 964: Since every field in `monitoringv1.ScrapeLimits` has a `json:"...,omitempty"` struct tag, zero values like `Samples: 0` are stripped during serialization and emit an empty object `limits: {}` in the manifest. Consider returning `nil` if all fields in `limits` remain zero:
```suggestion
	if *limits == (monitoringv1.ScrapeLimits{}) {
		return nil
	}
	return limits
```

pkg/migrate/podmonitor.go
```diff
@@ -304,6 +322,38 @@ func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonit
...
+			hasTrue = true
+		}
+	}
+	// A nil filterRunning defaults to true in the GMP operator.
+	var filterRunning *bool
+	if hasFalse {
+		falseVal := false
+		filterRunning = &falseVal
+		if hasTrue {
+			logger.Warn("Endpoint-level configuration conflict detected: some endpoints are configured with 'filterRunning: false' and others with 'true' (or default), but GMP only supports 'filterRunning' at the resource level. Setting 'filterRunning: false' globally on the PodMonitoring resource.")
```
dashpole, line 351: Promoting `filterRunning: false` globally from a single specialized endpoint causes standard app endpoints in this monitor to attempt scraping terminated, failed, or pending pods, filling collection logs with timeouts.

A clean architectural solution would be Automatic Resource Sub-Division (Endpoint Splitting): when endpoints have incompatible `filterRunning` requirements, automatically split the input `PodMonitor` into distinct GMP `PodMonitoring` resources (e.g., `<name>-running` and `<name>-all-states`).

Given the complexity of dynamic resource splitting, please feel free to just add a TODO comment here to track this as a future improvement!

dashpole (commented):
### 🏗️ Architectural Proposal: The "action-required Filter" Pattern for Partial Migration

As we expand the PO-to-GMP migration tool to cover complex real-world Prometheus Operator (PO) manifests, we frequently encounter settings that are either structurally unsupported by GKE Managed Prometheus (GMP) CRDs, security-blocked by validation webhooks, or unresolvable in an offline CLI tool.

Currently, our conversion logic handles these friction points in one of two ways:
1. **Hard Failures (`return nil, error`)**: Aborting conversion entirely when encountering schema incompatibilities (e.g., missing ports or basic-auth in `proxyUrl`).
2. **Lossy Drops with Warning Logs**: Dropping rules and letting the resource deploy in an active state (e.g., dropping annotation target filters or unresolved `ScrapeClasses`).

Both approaches create friction: hard failures break batch automated migration pipelines, while active lossy drops can cause continuous scrape failure logs, paging alerts, or massive telemetry billing spikes.

To solve this across the broader migration architecture, we should adopt the **"Inert Draft / No-Op Placeholder Filter"** pattern as a core structural design principle.

---

#### ⚙️ How the Mechanism Works (3-Step Pipeline)

When the migration tool encounters an unresolvable required setting, an unsupported authentication mechanism, or a dropped filtering rule that poses high operational risk, instead of throwing a fatal error or exporting a broken/unbounded active scraper, it performs three steps:

1. **Sanitize the Incompatible Field**: Replace illegal or unresolvable values with valid schema defaults so that `kubectl apply` succeeds without Kubernetes API webhook rejection.
2. **Inject an No-Op action-required Pod Selector**: Override or append to `selector.matchLabels` an intentionally impossible placeholder label requirement:
   ```yaml
   selector:
     matchLabels:
       gmp.migration.todo/action-required: "configure-proxy-authentication"
3. **Attach an Actionable Remediation Annotation**: Document the exact modification and manual operator follow-up steps directly in `metadata.annotations`:
   ```yaml
   metadata:
     annotations:
       gmp.migration.google.com/action-required: |
         Endpoint [0] proxyUrl contained plaintext credentials which were stripped for GMP API compatibility. 
         Configure egress network/proxy authentication, then delete the 'gmp.migration.todo' label from matchLabels to activate live scraping.
   ```

---

#### 🌟 Key Benefits of the Pattern

* **Zero Operational Noise**: Because no pods in the cluster will ever match the placeholder label `gmp.migration.todo/action-required: ...`, Prometheus discovers **0 targets and performs 0 scrapes**.
* **Prevents Scrape Scope & Billing Explosions**: Eliminates the risk of dropped pre-scrape filtering rules accidentally causing an unbounded wildcard scrape across an entire namespace or cluster, safeguarding teams against unexpected billing surges.
* **Eliminates Hard Migration Roadblocks**: Converts static validation errors (which normally halt automated batch migration scripts) into cleanly exported, 99%-complete draft manifests sitting safely in git repositories ready for operator inspection.

---

#### 📍List of Application Areas Across the Migration Tool

We should institutionalize this pattern across the following seven scenarios:

1. **Credentials Embedded in `proxyUrl` (Basic-Auth in URLs)**
   * *Current Behavior*: Fatal error; completely halts resource conversion because GMP API validation rejects plaintext URL credentials.
   * *Placeholder Application*: Strip out `user:pass@`, export the sanitized URL `http://proxy-server:8080`, and apply the no-op selector directing operators to configure egress proxy auth at the network or service mesh layer.

2. **Dropped Annotation & Node-Label Target Filters on Empty Selectors**
   * *Current Behavior*: Drops rules referencing pod annotations (`__meta_kubernetes_pod_annotation_*`) or node labels (`__meta_kubernetes_node_label_*`) with a warning log, causing an empty selector to passively scrape *every* pod in the namespace or cluster.
   * *Placeholder Application*: When dropping a target filter leaves an unconstrained or empty selector, inject the no-op placeholder selector to prevent an accidental wildcard scrape and force operator verification.

3. **Dropped `KeepEqual`, `DropEqual`, `LabelKeep`, and `LabelDrop` Actions**
   * *Current Behavior*: Dropped with a log warning; removes ingestion deduplication and high-availability active/standby filters.
   * *Placeholder Application*: When deduplication rules cannot be promoted to post-scrape `metricRelabelings`, render the monitor inert with a placeholder selector to prevent double-ingesting high-volume metrics.

4. **Missing or Omitted `port` and `targetPort` in Endpoints**
   * *Current Behavior*: Fatal error; conversion aborts because GMP schemas strictly require a port definition.
   * *Placeholder Application*: Emit a placeholder `port: "GMP_MIGRATION_TODO_PORT"`, apply the no-op selector, and add a remediation annotation prompting the operator to supply the target container port name.

5. **Conflicting Pre-Scrape `keep` Rules on the Same Pod Label**
   * *Current Behavior*: Fatal error; selector translation fails when multiple regex keep rules target the same Kubernetes pod label.
   * *Placeholder Application*: Avoid aborting migration; generate an inert draft with instructions to manually resolve or promote complex boolean discovery logic into post-scrape rules.

6. **Unsupported Protobuf Scrape Protocols (`PrometheusProto`) & Disabled HTTP/2**
   * *Current Behavior*: Logs a warning that scrapes may fail if the target lacks OpenMetrics/text fallbacks or HTTP/2 negotiation.
   * *Placeholder Application*: Inject the no-op selector and diagnostic annotation to prevent immediate scrape negotiation failures in production until text fallbacks are confirmed on target workloads.

---

(My own writing)

This isn't in scope for this PR (would be a follow-up), and requires discussion with @bernot-dev and @bwplotka. I think we are going to encounter a lot of cases where we can translate most of a PodMonitor/ServiceMonitor, but get stuck on a few fields we can't handle. If there is some mechanism (i'm sure there alternatives to the above) to give the user what we are able to figure out, and clearly signal what is left to be migrated, I think that would help make this less brittle.

dashpole (approved):

pkg/migrate/podmonitor.go
```diff
@@ -353,12 +405,55 @@ func (c *PodMonitorConverter) convertToClusterPodMonitoring(pm *pomonitoringv1.P
 		logger.Warn("Resulting ClusterPodMonitoring selector is empty. It will select and scrape all pods across all namespaces. Verify if this is intended.")
```
karthunni, line 405: I don't think namespace selector supports `matchExpressions`:
https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.NamespaceSelector

pkg/migrate/podmonitor.go
karthunni: In the design, we decided on this current functionality (set to false). 

Adding a comment to document this potential TODO in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1990 instead (since its refactored into a helper function)

PR #1990, opened 2026-07-21
PO to GMP Migration Tool: Refactor Common Conversion Functions

karthunni:
Refactoring conversion functions to reduce code duplication and simplify implementation for impending ServiceMonitor migration logic.

### DIFF
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -154,13 +154,26 @@ func determineNamespaceScoping(nsSel pomonitoringv1.NamespaceSelector, defaultNS
 	return []string{defaultNS}, false, nil
 }
 
+// commonMonitorSpec holds common fields extracted from Prometheus Operator monitor specs for building GMP resources.
+type commonMonitorSpec struct {
+	endpoints        []monitoringv1.ScrapeEndpoint
+	mergedFromPod    []monitoringv1.LabelMapping
+	mergedSelector   metav1.LabelSelector
+	metadata         *[]string
+	filterRunning    *bool
+	limits           *monitoringv1.ScrapeLimits
+	generatedSecrets []*unstructured.Unstructured
+}
+
 // conversionContext groups common parameters passed down to conversion helper functions.
 type conversionContext struct {
 	logger *slog.Logger
 	// cache provides access to dependent resources.
 	cache *ResourceCache
-	// namespace is the source namespace of the primary resource.
-	namespace string
+	// sourceNamespace is the original namespace where inputs (Secrets/ConfigMaps) live in the cache.
+	sourceNamespace string
+	// targetNamespace is the destination namespace for generated resources.
+	targetNamespace string
 	// generatedSecrets accumulates created Secrets when migrating ConfigMaps, keyed by Secret name.
 	generatedSecrets map[string]*unstructured.Unstructured
 }
@@ -380,7 +393,7 @@ func (c *conversionContext) extractResourceKey(kind, name, key string) (string, 
 		return "", fmt.Errorf("%s reference has an empty key for name %q", kindUpper, name)
 	}
 
-	obj, ok := c.cache.Get(kind, c.namespace, name)
+	obj, ok := c.cache.Get(kind, c.sourceNamespace, name)
 	if !ok {
 		c.logger.Warn("Resource not found in cache. Cannot extract key. Hardcoding placeholder.",
 			slog.String("referenced_kind", kind),
@@ -469,7 +482,7 @@ func (c *conversionContext) convertConfigMapToSecretSelector(sel *corev1.ConfigM
 	}
 
 	if _, exists := c.generatedSecrets[secretName]; !exists {
-		obj, ok := c.cache.Get(KindConfigMap, c.namespace, sel.Name)
+		obj, ok := c.cache.Get(KindConfigMap, c.sourceNamespace, sel.Name)
 		if !ok {
 			c.logger.Warn("TLS ConfigMap reference was not found in the inputs. Updated reference to GMP Secret, but you must manually convert your ConfigMap to a Secret with this name in GMP.",
 				slog.String("configmap", sel.Name),
@@ -483,7 +496,7 @@ func (c *conversionContext) convertConfigMapToSecretSelector(sel *corev1.ConfigM
 			newSecret.SetAPIVersion("v1")
 			newSecret.SetKind(KindSecret)
 			newSecret.SetName(secretName)
-			newSecret.SetNamespace(c.namespace)
+			newSecret.SetNamespace(c.targetNamespace)
 
 			data, found, _ := unstructured.NestedMap(obj.Object, "data")
 			if found {
@@ -497,7 +510,7 @@ func (c *conversionContext) convertConfigMapToSecretSelector(sel *corev1.ConfigM
 		}
 	}
 
-	secretRef := &monitoringv1.SecretKeySelector{Name: secretName, Key: secretKey, Namespace: c.namespace}
+	secretRef := &monitoringv1.SecretKeySelector{Name: secretName, Key: secretKey, Namespace: c.targetNamespace}
 	return &monitoringv1.SecretSelector{Secret: secretRef}, nil
 }
 
@@ -531,7 +544,7 @@ func (c *conversionContext) convertSecretSelector(sel *corev1.SecretKeySelector)
 		c.logger.Warn("Secret reference had 'optional: true'. GMP does not support optional secrets. The reference is now mandatory.",
 			slog.String("secret", sel.Name))
 	}
-	secretRef := &monitoringv1.SecretKeySelector{Name: sel.Name, Key: sel.Key, Namespace: c.namespace}
+	secretRef := &monitoringv1.SecretKeySelector{Name: sel.Name, Key: sel.Key, Namespace: c.targetNamespace}
 	return &monitoringv1.SecretSelector{Secret: secretRef}, nil
 }
 
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -16,6 +16,7 @@ package migrate
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -30,7 +31,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
@@ -125,8 +125,8 @@ func CopyObjectMeta(src metav1.ObjectMeta, targetNamespace string, logger *slog.
 	return dst
 }
 
-// ParseAndCleanNamespaces trims whitespace, filters out empty strings, and deduplicates namespaces.
-func ParseAndCleanNamespaces(namespaces []string) []string {
+// parseAndCleanNamespaces trims whitespace, filters out empty strings, and deduplicates namespaces.
+func parseAndCleanNamespaces(namespaces []string) []string {
 	unique := make(map[string]bool)
 	var cleaned []string
 	for _, ns := range namespaces {
@@ -137,6 +137,21 @@ func ParseAndCleanNamespaces(namespaces []string) []string {
 		}
 	}
 	return cleaned
+}
+
+// determineNamespaceScoping resolves the target namespaces from a NamespaceSelector.
+func determineNamespaceScoping(nsSel pomonitoringv1.NamespaceSelector, defaultNS string) ([]string, bool, error) {
+	if nsSel.Any {
+		return nil, true, nil
+	}
+	if len(nsSel.MatchNames) > 0 {
+		targetNamespaces := parseAndCleanNamespaces(nsSel.MatchNames)
+		if len(targetNamespaces) == 0 {
+			return nil, false, errors.New("namespaceSelector.matchNames contains only empty or invalid values")
+		}
+		return targetNamespaces, false, nil
+	}
+	return []string{defaultNS}, false, nil
 }
 
 // conversionContext groups common parameters passed down to conversion helper functions.
@@ -1023,6 +1038,21 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
 	return limits
 }
 
+// toStrictUnstructured converts a struct to a strictly JSON-compatible unstructured map.
+// uses JSON to silently convert unsupported Go primitives (like uint64) into safe float64 numbers.
+// ensures the resulting map will not panic on DeepCopy.
+func toStrictUnstructured(obj interface{}) (map[string]interface{}, error) {
+	b, err := json.Marshal(obj)
+	if err != nil {
+		return nil, err
+	}
+	var u map[string]interface{}
+	if err := json.Unmarshal(b, &u); err != nil {
+		return nil, err
+	}
+	return u, nil
+}
+
 // buildPodMonitoring constructs a GMP PodMonitoring resource from common spec.
 func buildPodMonitoring(
 	srcMeta metav1.ObjectMeta,
@@ -1048,7 +1078,7 @@ func buildPodMonitoring(
 		},
 	}
 
-	unstructuredMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(gmpPM)
+	unstructuredMap, err := toStrictUnstructured(gmpPM)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal PodMonitoring: %w", err)
 	}
@@ -1085,7 +1115,7 @@ func buildClusterPodMonitoring(
 		},
 	}
 
-	unstructuredMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(gmpCPM)
+	unstructuredMap, err := toStrictUnstructured(gmpCPM)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal ClusterPodMonitoring: %w", err)
 	}
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -1031,3 +1031,64 @@ func buildClusterPodMonitoring(
 
 	return u, nil
 }
+
+// resolveScrapeClass handles ScrapeClass resolution.
+// Logs a warning that ScrapeClass settings will be lost.
+// TODO(M2): Lookup ScrapeClass from Prometheus CR and merge its settings.
+func resolveScrapeClass(name *string, logger *slog.Logger) {
+	if name != nil && *name != "" {
+		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *name))
+	}
+}
+
+// validateScrapeProtocols logs a warning if any scrape protocol requires Protobuf.
+func validateScrapeProtocols(protocols []pomonitoringv1.ScrapeProtocol, logger *slog.Logger) {
+	for _, sp := range protocols {
+		if sp == scrapeProtocolPrometheusProto || strings.Contains(strings.ToLower(string(sp)), "proto") {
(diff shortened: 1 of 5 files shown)


pkg/migrate/helpers.go
```diff
@@ -832,3 +841,123 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+	}
+	if hasFalse {
+		falseVal := false
+		if hasTrue {
+			logger.Warn("Endpoint-level configuration conflict detected: some endpoints are configured with 'filterRunning: false' and others with 'true' (or default), but GMP only supports 'filterRunning' at the resource level. Setting 'filterRunning: false' globally.")
+		}
+		return &falseVal
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
karthunni: GMP operator defaults `p.Spec.FilterRunning == nil ` to `true`, but we can be explicit if preferred

pkg/migrate/helpers.go
```diff
@@ -832,3 +918,225 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+// convertProxyURL verifies proxy URL credentials.
+func convertProxyURL(proxyURL *string) (string, error) {
+	if proxyURL == nil {
+		return "", nil
+	}
+	if strings.Contains(*proxyURL, "@") {
+		return "", errors.New("proxyUrl contains credentials (matches '@'), which is blocked by GMP API validation")
+	}
+	return *proxyURL, nil
+}
```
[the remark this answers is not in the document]
karthunni: According to the design, we want to return an error in this case as scraping would fail with a placeholder

pkg/migrate/helpers.go
```diff
@@ -136,6 +139,21 @@ func ParseAndCleanNamespaces(namespaces []string) []string {
...
+	}
+	if len(nsSel.MatchNames) > 0 {
+		targetNamespaces := parseAndCleanNamespaces(nsSel.MatchNames)
+		if len(targetNamespaces) == 0 {
+			return nil, false, errors.New("namespaceSelector.matchNames contains only empty or invalid values")
+		}
+		return targetNamespaces, false, nil
+	}
+	return []string{defaultNS}, false, nil
+}
```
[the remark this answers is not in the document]
karthunni: Not sure where this "general rule" is coming from, but anything malformed configuration should lead to errors (i.e if migrating would lead to failed scrapes).

pkg/migrate/podmonitor.go
```diff
@@ -147,166 +127,104 @@ func (c *PodMonitorConverter) convertEndpoints(
...
 	}
 
 	mergedFromPod := mergeFromPod(logger, convertTargetLabels(logger, pm.Spec.PodTargetLabels, pm.Spec.JobLabel, "Pod"), rules.ResourceCombined.FromPod)
 	mergedSelector, err := mergeLabelSelector(pm.Spec.Selector, rules.ResourceCombined.MatchLabels, rules.ResourceCombined.MatchExpressions)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 	if len(mergedSelector.MatchLabels) == 0 && len(mergedSelector.MatchExpressions) == 0 {
 		logger.Warn("Resulting PodMonitoring selector is empty. It will select and scrape all pods in this namespace. Verify if this is intended.")
```
dashpole: The empty selector check was moved into `convertToMonitorSpec`, so converting a cluster-scoped monitor now logs `Resulting PodMonitoring selector is empty... in this namespace` instead of `ClusterPodMonitoring` across all namespaces. Consider passing the resource kind down or moving the warning to the kind-specific builder.

pkg/migrate/helpers.go
```diff
@@ -913,3 +995,225 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
 	}
 	return limits
 }
+
+// toStrictUnstructured converts a struct to a strictly JSON-compatible unstructured map.
```
dashpole: The comment states JSON is used for unsupported Go primitives like `uint64`, but `ToUnstructured` does support `uint64` (the real issue is `unstructured.DeepCopy` panicking on `uint64`). Also, unmarshaling numbers into `float64` loses integer precision above 2^53 for large `ScrapeLimits`. Consider normalizing limits to `int64` or using `json.NewDecoder` with `UseNumber()`.

pkg/migrate/helpers.go
```diff
@@ -913,3 +995,225 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+	for _, sp := range protocols {
+		if strings.Contains(strings.ToLower(string(sp)), "protobuf") {
+			logger.Warn("Scrape protocol settings (scrapeProtocols) requiring Protobuf are unsupported. Scrapes may fail if target lacks text fallback.")
+			break
+		}
+	}
+}
+
+// resolveAttachMetadata appends "node" to metadata if attachMetadata.node is enabled.
+func resolveAttachMetadata(attachMetadata *pomonitoringv1.AttachMetadata, baseMetadata *[]string) *[]string {
```
dashpole: nit: Because `resolveAttachMetadata` runs before `unionMetadata`, setting `attachMetadata.node: true` returns a non-nil slice (`node`), which causes `unionMetadata` to pull in all default metadata labels (`container`, `pod`, etc.). It would be helpful to add a comment explaining this tradeoff.

pkg/migrate/podmonitor.go
```diff
@@ -147,166 +127,104 @@ func (c *PodMonitorConverter) convertEndpoints(
...
+		mergedFromPod:    mergedFromPod,
+		mergedSelector:   mergedSelector,
+		metadata:         metadata,
+		filterRunning:    filterRunning,
+		limits:           limits,
+		generatedSecrets: convCtx.getGeneratedSecrets(),
+	}, nil
+}
+
+func (c *PodMonitorConverter) convertToPodMonitoring(pm *pomonitoringv1.PodMonitor, logger *slog.Logger, cache *ResourceCache) (*unstructured.Unstructured, []*unstructured.Unstructured, error) {
```
dashpole: `convertToPodMonitoring` and `convertToClusterPodMonitoring` are structurally identical except for their defaults slice and builder call. Consider collapsing them into a single parameterized helper.

pkg/migrate/helpers.go
```diff
@@ -612,6 +634,66 @@ func (c *conversionContext) convertAuthorization(auth *pomonitoringv1.SafeAuthor
 	}, nil
 }
 
+// applyAuthAndTLS converts credentials and TLS settings for a generic endpoint.
+func (c *conversionContext) applyAuthAndTLS(
```
dashpole, line 665: `applyAuthAndTLS` accepts `i int` to format `endpoint [%d]`, but the caller at `podmonitor.go:155` wraps the error again, producing redundant error context. Consider dropping `i` from `applyAuthAndTLS` and wrapping errors at the call site.

pkg/migrate/helpers.go
```diff
@@ -77,21 +81,21 @@ type relabelingData struct {
 	rewrittenSources []string
```
dashpole, line 86: `PreScrapeRelabelingResult` and `ExtractedPreScrapeRules` were exported while `parseAndCleanNamespaces` was un-exported. Since neither struct is used outside `pkg/migrate`, consider keeping them un-exported.

pkg/migrate/helpers.go
```diff
@@ -913,3 +995,225 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+			break
+		}
+	}
+}
+
+// resolveAttachMetadata appends "node" to metadata if attachMetadata.node is enabled.
+func resolveAttachMetadata(attachMetadata *pomonitoringv1.AttachMetadata, baseMetadata *[]string) *[]string {
+	if attachMetadata != nil && attachMetadata.Node != nil && *attachMetadata.Node {
+		if baseMetadata == nil {
+			return &[]string{"node"}
```
dashpole: Quick nit: `resolveAttachMetadata` hardcodes `"node"` directly instead of using the `labelNode` constant defined earlier in the file.

pkg/migrate/helpers.go
```diff
@@ -913,3 +995,225 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+	if honorTimestamps != nil && *honorTimestamps {
+		logger.Warn("Field 'honorTimestamps: true' is unsupported and dropped. GMP always uses the scrape ingestion timestamp. Target metric timestamps will be ignored.")
+	}
+	if trackTimestampsStaleness != nil {
+		logger.Warn("Field 'trackTimestampsStaleness' is unsupported in GMP and has been dropped.")
+	}
+}
+
+// combineAndConvertRelabelings combines promoted pre-scrape rules and converts metricRelabelings.
+func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.RelabelingRule, configs []pomonitoringv1.RelabelConfig) ([]monitoringv1.RelabelingRule, error) {
```
dashpole: `combineAndConvertRelabelings` returns an error, but `convertMetricRelabelings` always returns `nil` error. Consider dropping the error return from both signatures.

pkg/migrate/helpers.go
```diff
@@ -130,6 +134,21 @@ func ParseAndCleanNamespaces(namespaces []string) []string {
 	return cleaned
 }
 
+// determineNamespaceScoping resolves the target namespaces from a NamespaceSelector.
+func determineNamespaceScoping(nsSel pomonitoringv1.NamespaceSelector, defaultNS string) ([]string, bool, error) {
```
dashpole, line 143: The newly extracted helpers (`determineNamespaceScoping`, `resolveScrapeIntervalAndTimeout`, `convertProxyURL`, `resolveFilterRunning`, `resolveAttachMetadata`, `toStrictUnstructured`) currently lack direct unit tests in `helpers_test.go`. Adding table-driven unit tests would ensure boundary conditions and error cases are thoroughly covered.

pkg/migrate/helpers.go
```diff
@@ -955,3 +1031,242 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+
+// resolveAttachMetadata appends "node" to metadata if attachMetadata.node is enabled.
+func resolveAttachMetadata(attachMetadata *pomonitoringv1.AttachMetadata, baseMetadata *[]string, isCluster bool) *[]string {
+	if attachMetadata != nil && attachMetadata.Node != nil && *attachMetadata.Node {
+		if baseMetadata == nil {
+			if isCluster {
+				union := unionMetadata([]string{labelNode}, clusterMetadataDefaults)
+				return &union
+			}
+			return &[]string{labelNode}
```
dashpole: If baseMetadata is nil for namespaced PodMonitoring, returning only labelNode drops default metadata labels (container, pod, etc). Should we union with namespacedMetadataDefaults here?
```suggestion
			union := unionMetadata([]string{labelNode}, namespacedMetadataDefaults)
			return &union
```

pkg/migrate/helpers.go
```diff
@@ -955,3 +1031,242 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+		logger.Warn(fmt.Sprintf("endpoint [%d]: field 'enableHttp2: false' is unsupported by GMP Managed Collection and has been dropped. The collector will always negotiate HTTP/2 for TLS connections.", i))
+	}
+	if honorLabels {
+		logger.Warn(fmt.Sprintf("endpoint [%d]: field 'honorLabels: true' is unsupported and dropped. GMP always overrides conflicting labels. Clashing metric labels will be renamed with the 'exported_' prefix.", i))
+	}
+	if honorTimestamps != nil && *honorTimestamps {
+		logger.Warn(fmt.Sprintf("endpoint [%d]: field 'honorTimestamps: true' is unsupported and dropped. GMP always uses the scrape ingestion timestamp. Target metric timestamps will be ignored.", i))
+	}
+	if trackTimestampsStaleness != nil {
+		logger.Warn(fmt.Sprintf("endpoint [%d]: fField 'trackTimestampsStaleness' is unsupported in GMP and has been dropped.", i))
```
dashpole: Small typo in warning message:
```suggestion
		logger.Warn(fmt.Sprintf("endpoint [%d]: field 'trackTimestampsStaleness' is unsupported in GMP and has been dropped.", i))
```

pkg/migrate/helpers.go
```diff
@@ -955,3 +1031,242 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+	}
+}
+
+// combineAndConvertRelabelings combines promoted pre-scrape rules and converts metricRelabelings.
+func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.RelabelingRule, configs []pomonitoringv1.RelabelConfig) []monitoringv1.RelabelingRule {
+	totalRules := len(promoted) + len(configs)
+	if totalRules == 0 {
```
dashpole, line 1280: If all configs are dropped during metric relabeling conversion, allRules is an empty non-nil slice, which serializes as metricRelabeling: [] instead of omitting the field. Returning nil when empty keeps the manifest clean:
```suggestion
	if len(allRules) == 0 {
		return nil
	}
	return allRules
```

pkg/migrate/podmonitor.go
```diff
@@ -76,44 +77,23 @@ func (c *PodMonitorConverter) Convert(_ context.Context, logger *slog.Logger, un
...
-	u, generatedSecrets, err := c.convertToPodMonitoring(&podMonitor, logger, cache)
+	baseU, generatedSecrets, err := c.convertToPodMonitoring(&podMonitor, logger, cache)
 	if err != nil {
 		return nil, err
 	}
-	outputs := []*unstructured.Unstructured{u}
+
+	var outputs []*unstructured.Unstructured
+	for _, ns := range targetNamespaces {
+		uClone := baseU.DeepCopy()
```
dashpole: When targetNamespaces contains multiple namespaces, cloning baseU and setting namespace to ns leaves secret references with Namespace set to the source namespace, creating invalid cross-namespace secret references. Consider updating secretRef.Namespace during cloning.

pkg/migrate/helpers.go
```diff
@@ -959,3 +1035,244 @@ func convertLimits(sampleLimit, labelLimit, labelNameLengthLimit, labelValueLeng
...
+		return nil, err
+	}
+	return sanitizeUInt64(u).(map[string]any), nil
+}
+
+// sanitizeUInt64 recursively converts uint64 primitives to int64 for unstructured compatibility.
+func sanitizeUInt64(val any) any {
+	switch v := val.(type) {
+	case uint64:
+		return int64(v)
```
dashpole, line 1070: If `v` is greater than `math.MaxInt64` (for instance, if `math.MaxUint64` was used in config to indicate unlimited), casting directly to `int64(v)` wraps around to a negative number like `-1`, which could fail CRD schema validation. Consider checking if `v > math.MaxInt64` and capping it at `math.MaxInt64` (note: would require importing `math`).

pkg/migrate/podmonitor.go
```diff
@@ -147,338 +126,151 @@ func (c *PodMonitorConverter) convertEndpoints(
...
 	}
-
-	var hasFalse, hasTrue bool
-	for _, ep := range pm.Spec.PodMetricsEndpoints {
-		if ep.FilterRunning != nil && !*ep.FilterRunning {
-			hasFalse = true
-		} else {
-			hasTrue = true
-		}
+	if len(md) > 0 {
```
dashpole: Since `namespacedMetadataDefaults` guarantees non-namespace labels in `union`, `md` will always be non-empty here. We can simplify this by removing the `len(md) > 0` check and fallback and simply returning `&md` directly.

pkg/migrate/helpers.go
karthunni: Seems uncommon/unrealistic but can add it for safety

pkg/migrate/podmonitor.go
```diff
@@ -76,45 +77,23 @@ func (c *PodMonitorConverter) Convert(_ context.Context, logger *slog.Logger, un
...
+			slog.Any("namespaces", targetNamespaces),
+		)
+	}
 
-		// 2.2 Convert to a base namespaced PodMonitoring.
-		baseU, generatedSecrets, err := c.convertToPodMonitoring(&podMonitor, logger, cache)
+	var outputs []*unstructured.Unstructured
+	for _, ns := range targetNamespaces {
+		pmCopy := podMonitor.DeepCopy()
+		pmCopy.Namespace = ns
```
dashpole: Setting pmCopy.Namespace = ns before conversion causes ResourceCache lookups to use the target namespace instead of the source namespace, which will cause lookups for existing Secrets and ConfigMaps to fail. Consider keeping the original namespace in conversionContext for cache lookups while assigning ns to the resulting monitoring resource.

pkg/migrate/types.go
```diff
@@ -37,6 +39,16 @@ const (
 	KindSecret               = "Secret"
 )
 
+type commonMonitorSpec struct {
```
dashpole: Since commonMonitorSpec is an unexported struct used only by internal builders in helpers.go, consider defining it in helpers.go to keep types.go dedicated to public interfaces and models.

pkg/migrate/podmonitor.go
```diff
@@ -147,338 +124,150 @@ func (c *PodMonitorConverter) convertEndpoints(
...
+// filterMetadata applies namespaced or cluster metadata defaults and strips namespace metadata in namespaced resources.
+func filterMetadata(metadata *[]string, isCluster bool, logger *slog.Logger) *[]string {
+	if metadata == nil {
+		return nil
 	}
-	// Check against the PrometheusProto enum value.
-	if slices.Contains(pm.Spec.ScrapeProtocols, scrapeProtocolPrometheusProto) {
-		logger.Warn("Scrape protocol settings (scrapeProtocols) requiring Protobuf are unsupported. Scrapes may fail if target lacks text fallback.")
+	if isCluster {
+		union := unionMetadata(*metadata, clusterMetadataDefaults)
```
dashpole, line 224: Since `unionMetadata` is called directly by shared utilities in `helpers.go` (such as in `resolveAttachMetadata`), consider moving it and the default metadata slices into `helpers.go`. Conversely, PodMonitor-specific functions currently residing in `helpers.go` like `extractPreScrapeRelabelings` and `warnUnsupportedSpecFields` could live here in `podmonitor.go` to cleanly separate shared generic utilities from resource-specific logic.

pkg/migrate/helpers.go
```diff
@@ -630,10 +661,68 @@ func (c *conversionContext) convertAuthorization(auth *pomonitoringv1.SafeAuthor
...
+	if authorization != nil {
+		auth, err := c.convertAuthorization(authorization)
+		if err != nil {
+			return fmt.Errorf("authorization: %w", err)
+		}
+		gmpEp.Authorization = auth
+	}
+
+	// Handle deprecated BearerTokenSecret -> Authorization.
+	if bearerTokenSecret.Name != "" { // nolint:staticcheck // Map deprecated BearerTokenSecret for backwards compatibility.
```
dashpole: If a malformed `bearerTokenSecret` specifies an empty Name but a non-empty Key, checking only `Name != ""` silently skips it without returning a validation error. Consider checking `if bearerTokenSecret.Name != "" || bearerTokenSecret.Key != ""` so malformed references fail appropriately in `convertAuthorization`.

pkg/migrate/helpers_test.go
```diff
@@ -599,3 +601,366 @@ func TestConvertLimits(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestToStrictUnstructured(t *testing.T) {
+	tests := []struct {
+		name      string
+		obj       any
+		expectErr bool
```
dashpole: Quick nit: `expectErr` is defined in the test table struct here, but only one positive test case is exercised. Consider either adding a negative test case (e.g., passing an incompatible struct that fails unstructured conversion) or removing the error check field to simplify the test.

dashpole (approved):

pkg/migrate/podmonitor.go
karthunni: All 3 functions get refactored to be used in `helpers.go`  for both PodMonitor and ServiceMonitor in the following PR:
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2022

PR #1988, opened 2026-07-13
PO to GMP Migration Tool: Podmonitor Relabeling Migration

karthunni:
Adds support for migrating Prometheus Operator pre-scrape relabeling
configurations (`relabelings`) to Managed Service for Prometheus (GMP)
equivalent fields.

* Translates simple 1:1 label copies to `targetLabels.fromPod` and
   `targetLabels.metadata`.
* Translates exact equality and set-based target filtering rules
   to Kubernetes label selectors (`matchLabels` and `matchExpressions`).
* Promotes complex value-changing or regex-based rules to endpoint-level
   post-scrape `MetricRelabelings` while importing required Kubernetes
   source labels via `targetLabels.fromPod`.
* Clean warnings and drops for unsupported actions (like `labelmap`, `lowercase`, `uppercase`) and
  relabelings targeting pod annotations.

PR #2002, opened 2026-07-24
PO to GMP Migration Tool: Service Resolution Functions

karthunni:
This PR implements the core service and port resolution helper functions in `pkg/migrate/helpers.go` required for the upcoming ServiceMonitor migration work.

Key additions:
- `findServicesBySelector`: Traverses the ResourceCache to return Service CRs matching a specified LabelSelector within target namespaces.
- `resolveServicePort`: Inspects a Service's `spec.ports` list to map a target port reference (either a string name or a port number) to the Pod's actual container targetPort value.
- `convertServiceTargetLabels`: Maps Service-level labels (as defined in `spec.targetLabels` of ServiceMonitor) to static metric relabeling rules (action: replace), appending the "exported_" prefix if it clashes with protected Prometheus labels.
- Added comprehensive unit tests in `helpers_test.go` verifying all helper behaviors, error paths, and edge cases.

### DIFF
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -1310,7 +1310,7 @@ func resolveServicePort(logger *slog.Logger, svc *corev1.Service, portStr string
 	}
 
 	for _, p := range svc.Spec.Ports {
-		if p.Port < 1 {
+		if p.Port < 1 || p.Port > 65535 {
 			logger.Warn("Service port entry has an invalid or out-of-range port number. Skipping malformed entry.",
 				slog.String("service", svc.Name),
 				slog.String("port_name", p.Name),
@@ -1336,6 +1336,7 @@ func convertServiceTargetLabels(logger *slog.Logger, svc *corev1.Service, target
 	}
 
 	var rules []monitoringv1.RelabelingRule
+	seenTargets := make(map[string]bool)
 
 	for _, l := range targetLabels {
 		val, ok := svc.Labels[l]
@@ -1353,6 +1354,15 @@ func convertServiceTargetLabels(logger *slog.Logger, svc *corev1.Service, target
 				slog.String("label", l),
 				slog.String("renamed_target", target))
 		}
+
+		if seenTargets[target] {
+			logger.Warn("Service targetLabel mapping collision. Skipping.",
+				slog.String("service", svc.Name),
+				slog.String("source_label", l),
+				slog.String("target_label", target))
+			continue
+		}
+		seenTargets[target] = true
 
 		rules = append(rules, monitoringv1.RelabelingRule{
 			TargetLabel: target,
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -1156,7 +1156,6 @@ func buildClusterPodMonitoring(
 
 // resolveScrapeClass handles ScrapeClass resolution.
 // Logs a warning that ScrapeClass settings will be lost.
-// TODO(M2): Lookup ScrapeClass from Prometheus CR and merge its settings.
 func resolveScrapeClass(name *string, logger *slog.Logger) {
 	if name != nil && *name != "" {
 		logger.Warn(fmt.Sprintf("ScrapeClass %q was not found in the inputs. The 'scrapeClassName' field has been dropped and inherited settings will be lost.", *name))
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"maps"
 	"slices"
 	"strings"
 
@@ -28,10 +27,10 @@ import (
 	prommodel "github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/google/export"
 	"github.com/prometheus/prometheus/model/relabel"
+	"github.com/prometheus/prometheus/util/strutil"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/validation"
@@ -1298,63 +1297,31 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
 	return allRules
 }
 
-// findServicesBySelector finds Services matching the selector in target namespaces (all if empty).
-func (c *ResourceCache) findServicesBySelector(selector metav1.LabelSelector, namespaces []string) ([]*unstructured.Unstructured, error) {
-	sel, err := metav1.LabelSelectorAsSelector(&selector)
-	if err != nil {
-		return nil, fmt.Errorf("invalid selector: %w", err)
-	}
-
-	var matched []*unstructured.Unstructured
-
-	if c == nil || c.resources == nil {
-		return nil, nil
-	}
-
-	services, ok := c.resources[KindService]
-	if !ok {
-		return nil, nil
-	}
-
-	// To make output deterministic, we sort the keys before iterating.
-	keys := slices.AppendSeq(make([]string, 0, len(services)), maps.Keys(services))
-	slices.Sort(keys)
-
-	for _, key := range keys {
-		svc := services[key]
-		svcNS := svc.GetNamespace()
-
-		if len(namespaces) > 0 && !slices.Contains(namespaces, svcNS) {
-			continue
-		}
-
-		svcLabels := svc.GetLabels()
-		if sel.Matches(labels.Set(svcLabels)) {
-			matched = append(matched, svc)
-		}
-	}
-	return matched, nil
-}
-
-// asInt32 coerces various Go numeric types into an int32.
+// asInt32 coerces various Go numeric types into an int32 within the valid port range [0, 65535].
 func asInt32(val any) (int32, bool) {
+	var n int64
 	switch v := val.(type) {
 	case int:
-		return int32(v), true
+		n = int64(v)
 	case int32:
-		return v, true
+		n = int64(v)
 	case int64:
-		return int32(v), true
+		n = v
 	case float32:
-		return int32(v), true
+		n = int64(v)
 	case float64:
-		return int32(v), true
+		n = int64(v)
+	default:
+		return 0, false
 	}
-	return 0, false
+	if n < 0 || n > 65535 {
+		return 0, false
+	}
+	return int32(n), true
 }
 
 // resolveServicePort resolves a Service port to the backing Pod's target port.
-func resolveServicePort(svc *unstructured.Unstructured, portStr string) (intstr.IntOrString, error) {
+func resolveServicePort(logger *slog.Logger, svc *unstructured.Unstructured, portStr string) (intstr.IntOrString, error) {
 	if portStr == "" {
 		return intstr.IntOrString{}, errors.New("port string cannot be empty")
 	}
@@ -1373,27 +1340,36 @@ func resolveServicePort(svc *unstructured.Unstructured, portStr string) (intstr.
 	for _, p := range ports {
 		portMap, ok := p.(map[string]any)
 		if !ok {
+			logger.Warn("Service port entry is not a valid map. Skipping malformed entry.",
+				slog.String("service", svc.GetName()))
 			continue
 		}
 
 		name, _, _ := unstructured.NestedString(portMap, "name")
 		portVal, foundField, err := unstructured.NestedFieldNoCopy(portMap, "port")
-		if err != nil {
-			return intstr.IntOrString{}, fmt.Errorf("failed to read port field: %w", err)
-		}
-		if !foundField {
-			return intstr.IntOrString{}, errors.New("service port spec is missing the port number")
+		if err != nil || !foundField {
+			logger.Warn("Service port spec is missing the port number. Skipping malformed entry.",
+				slog.String("service", svc.GetName()),
+				slog.String("port_name", name))
+			continue
 		}
 		portNum, ok := asInt32(portVal)
-		if !ok {
-			return intstr.IntOrString{}, fmt.Errorf("invalid port number type in Service spec: %T", portVal)
+		if !ok || portNum < 1 {
+			logger.Warn("Service port entry has an invalid or out-of-range port number. Skipping malformed entry.",
+				slog.String("service", svc.GetName()),
+				slog.String("port_name", name),
+				slog.Any("port_value", portVal))
+			continue
 		}
 
 		// Match by name or port number (as string).
 		if name == portStr || fmt.Sprintf("%d", portNum) == portStr {
 			targetPort, found, err := unstructured.NestedFieldNoCopy(portMap, "targetPort")
 			if err != nil {
-				return intstr.IntOrString{}, fmt.Errorf("failed to read targetPort: %w", err)
+				logger.Warn("Failed to read targetPort from Service port spec. Skipping malformed entry.",
+					slog.String("service", svc.GetName()),
+					slog.String("port_name", name))
+				continue
 			}
 			if !found {
 				// If targetPort is omitted, it defaults to the port number.
@@ -1402,6 +1378,9 @@ func resolveServicePort(svc *unstructured.Unstructured, portStr string) (intstr.
 
 			// targetPort can be int, float64, or string.
 			if valStr, ok := targetPort.(string); ok {
+				if valStr == "" {
+					return intstr.FromInt32(portNum), nil
+				}
 				return intstr.FromString(valStr), nil
 			}
 
@@ -1412,7 +1391,11 @@ func resolveServicePort(svc *unstructured.Unstructured, portStr string) (intstr.
 				return intstr.FromInt32(valNum), nil
 			}
 
-			return intstr.IntOrString{}, fmt.Errorf("invalid targetPort type: %T", targetPort)
+			logger.Warn("Service port entry has an invalid targetPort type. Skipping malformed entry.",
+				slog.String("service", svc.GetName()),
+				slog.String("port_name", name),
+				slog.Any("target_port", targetPort))
+			continue
 		}
 	}
 
@@ -1437,9 +1420,9 @@ func convertServiceTargetLabels(logger *slog.Logger, svc *unstructured.Unstructu
 			continue
 		}
 
-		target := l
-		if protectedLabels[l] {
-			target = "exported_" + l
+		target := strutil.SanitizeLabelName(l)
+		if protectedLabels[target] {
+			target = "exported_" + target
 			logger.Warn("Service targetLabel matches protected label. Renamed target.",
 				slog.String("label", l),
 				slog.String("renamed_target", target))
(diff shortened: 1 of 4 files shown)


pkg/migrate/helpers.go
```diff
@@ -1140,3 +1142,153 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+				}
+				return intstr.FromInt32(valNum), nil
+			}
+
+			return intstr.IntOrString{}, fmt.Errorf("invalid targetPort type: %T", targetPort)
+		}
+	}
+
+	return intstr.IntOrString{}, fmt.Errorf("port %q not found in Service spec", portStr)
+}
```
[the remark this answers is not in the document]
karthunni: For Point 2, we want to fail in that case as an incorrect port would lead to failed scrapes

pkg/migrate/helpers.go
```diff
@@ -1140,3 +1142,163 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+		if err != nil {
+			return intstr.IntOrString{}, fmt.Errorf("failed to read port field: %w", err)
+		}
+		if !foundField {
+			return intstr.IntOrString{}, errors.New("service port spec is missing the port number")
+		}
+		portNum, ok := asInt32(portVal)
+		if !ok {
+			return intstr.IntOrString{}, fmt.Errorf("invalid port number type in Service spec: %T", portVal)
+		}
```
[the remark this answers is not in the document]
karthunni: Our design choice was to return an error in this case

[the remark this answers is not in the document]
karthunni: Putting a placeholder would not be ideal as it would lead to a failed scrape. We can continue to fail the migration for a malformed port, or another option is to just drop the port

pkg/migrate/helpers.go
```diff
@@ -1217,3 +1219,163 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+	for _, l := range targetLabels {
+		val, ok := svcLabels[l]
+		if !ok {
+			logger.Warn("Service-level targetLabel was not found on Service. Skipping mapping.",
+				slog.String("label", l),
+				slog.String("service", svc.GetName()))
+			continue
+		}
+
+		target := l
```
dashpole: `convertServiceTargetLabels` uses raw Kubernetes Service label keys (such as `app.kubernetes.io/name`) as `TargetLabel`. Kubernetes label keys often contain dots and slashes, which are invalid in Prometheus label names. Sanitize non-alphanumeric characters to underscores first, then check `protectedLabels` against the sanitized name.

pkg/migrate/helpers.go
```diff
@@ -1217,3 +1219,163 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+			continue
+		}
+
+		name, _, _ := unstructured.NestedString(portMap, "name")
+		portVal, foundField, err := unstructured.NestedFieldNoCopy(portMap, "port")
+		if err != nil {
+			return intstr.IntOrString{}, fmt.Errorf("failed to read port field: %w", err)
+		}
+		if !foundField {
+			return intstr.IntOrString{}, errors.New("service port spec is missing the port number")
```
dashpole: In `resolveServicePort`, encountering a port entry missing `port` or with an invalid type returns a fatal error immediately. If a Service has multiple ports and an earlier entry is malformed, resolving a valid target port later in the slice fails. Consider skipping malformed entries and returning an error only if no matching port is resolved.

pkg/migrate/helpers.go
```diff
@@ -1217,3 +1219,163 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+		svcLabels := svc.GetLabels()
+		if sel.Matches(labels.Set(svcLabels)) {
+			matched = append(matched, svc)
+		}
+	}
+	return matched, nil
+}
+
+// asInt32 coerces various Go numeric types into an int32.
+func asInt32(val any) (int32, bool) {
```
dashpole: `asInt32` converts `int64` and `float64` to `int32` without bounds or range checks. Out-of-range values like `4294967297` truncate to `1` and can match the wrong port. Consider adding range validation for `[1, 65535]`.

pkg/migrate/helpers.go
```diff
@@ -1217,3 +1219,163 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+			if err != nil {
+				return intstr.IntOrString{}, fmt.Errorf("failed to read targetPort: %w", err)
+			}
+			if !found {
+				// If targetPort is omitted, it defaults to the port number.
+				return intstr.FromInt32(portNum), nil
+			}
+
+			// targetPort can be int, float64, or string.
+			if valStr, ok := targetPort.(string); ok {
```
dashpole: `targetPort: ""` returns `intstr.FromString("")` without validation, while numeric zero falls back to `portNum`. Empty string targetPorts should be treated as invalid or fall back to `portNum`.

pkg/migrate/helpers.go
```diff
@@ -1217,3 +1219,163 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+// findServicesBySelector finds Services matching the selector in target namespaces (all if empty).
+func (c *ResourceCache) findServicesBySelector(selector metav1.LabelSelector, namespaces []string) ([]*unstructured.Unstructured, error) {
+	sel, err := metav1.LabelSelectorAsSelector(&selector)
+	if err != nil {
+		return nil, fmt.Errorf("invalid selector: %w", err)
+	}
+
+	var matched []*unstructured.Unstructured
+
+	if c == nil || c.resources == nil {
```
dashpole: Move `if c == nil || c.resources == nil` check to the top of `findServicesBySelector` to fail fast before parsing the label selector.

pkg/migrate/helpers.go
```diff
@@ -1217,3 +1219,163 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
 
 	return allRules, nil
 }
+
+// findServicesBySelector finds Services matching the selector in target namespaces (all if empty).
+func (c *ResourceCache) findServicesBySelector(selector metav1.LabelSelector, namespaces []string) ([]*unstructured.Unstructured, error) {
```
dashpole: Note for future callers: `namespaces == nil` here matches every namespace, whereas Prometheus Operator default for an omitted `namespaceSelector` is the monitor's own namespace. Adding a note in the doc comment will clarify this behavior.

dashpole: `findServicesBySelector` is a method on `*ResourceCache`, which is defined in `types.go`. Consider moving `findServicesBySelector` to `types.go` alongside the rest of `ResourceCache` methods for co-location.

pkg/migrate/helpers_test.go
```diff
@@ -554,3 +555,325 @@ func TestMergeFromPod(t *testing.T) {
...
+			},
+			selector:   metav1.LabelSelector{MatchLabels: map[string]string{"app": "foo"}},
+			namespaces: []string{"default"},
+			expected:   []string{"svc-a"},
+			wantErr:    false,
+		},
+		{
+			name: "Match multiple services",
+			setupCache: func(cache *ResourceCache) error {
+				if err := addServiceToCache(cache, "default", "svc-a", map[string]string{"app": "foo"}); err != nil {
```
dashpole: In `TestFindServicesBySelector` ("Match multiple services"), `svc-a` is inserted before `svc-b`, matching alphabetical order. Inserting `svc-b` first ensures the explicit key sorting logic is verified.

pkg/migrate/helpers_test.go
```diff
@@ -554,3 +555,325 @@ func TestMergeFromPod(t *testing.T) {
...
+				return
+			}
+			if got != tc.expected {
+				t.Errorf("expected %+v, got %+v", tc.expected, got)
+			}
+		})
+	}
+}
+
+func TestConvertServiceTargetLabels(t *testing.T) {
```
dashpole, line 1225: `podmonitor_test.go` validates generated resources by running them through `ScrapeConfigs()`. Feeding `convertServiceTargetLabels` output into a minimal `PodMonitoring` and invoking `ScrapeConfigs()` would catch invalid target label errors end-to-end.

pkg/migrate/helpers_test.go
```diff
@@ -554,3 +555,325 @@ func TestMergeFromPod(t *testing.T) {
...
+	tests := []struct {
+		name         string
+		service      *unstructured.Unstructured
+		targetLabels []string
+		expected     []monitoringv1.RelabelingRule
+	}{
+		{
```
dashpole, line 1232: `TestConvertServiceTargetLabels` currently only tests simple labels (`app`, `env`, `job`). Add test cases covering labels with slashes and dots (e.g. `app.kubernetes.io/name`).

pkg/migrate/helpers_test.go
```diff
@@ -554,3 +555,325 @@ func TestMergeFromPod(t *testing.T) {
...
+func addServiceToCache(cache *ResourceCache, namespace, name string, labels map[string]string) error {
+	svc := &corev1.Service{
+		TypeMeta: metav1.TypeMeta{APIVersion: "v1", Kind: "Service"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels:    labels,
+		},
+	}
+	u, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(svc)
```
dashpole: `addServiceToCache`, `makeTestService`, and `makeTestServiceWithLabels` can be consolidated into a single builder taking labels and ports. Also, `u, _ :=` swallows `ToUnstructured` conversion errors instead of using `t.Fatal` or returning the error.

pkg/migrate/helpers_test.go
```diff
@@ -554,3 +555,325 @@ func TestMergeFromPod(t *testing.T) {
...
+
+	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := convertServiceTargetLabels(logger, tc.service, tc.targetLabels)
+			if len(got) != len(tc.expected) {
+				t.Fatalf("expected %d rules, got %d", len(tc.expected), len(got))
+			}
+			for i, r := range got {
```
dashpole: `TestConvertServiceTargetLabels` manually compares fields one-by-one. Using `cmp.Diff` (as done elsewhere in `helpers_test.go`) improves assertion readability and diff diagnostic quality.

pkg/migrate/helpers.go
dashpole: This isn't the same mistake as https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1988#discussion_r3683427851.  It is saying that we need to sanitize before we validate

pkg/migrate/helpers.go
dashpole: Hmmm... This seems good, but we definitely need to warn about dropped ports

pkg/migrate/helpers.go
```diff
@@ -1276,3 +1278,73 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+	}
+	if svc == nil {
+		return intstr.IntOrString{}, errors.New("cannot resolve port on nil or uninitialized Service")
+	}
+	if len(svc.Spec.Ports) == 0 {
+		return intstr.IntOrString{}, errors.New("service has no ports defined in spec")
+	}
+
+	for _, p := range svc.Spec.Ports {
+		if p.Port < 1 {
```
dashpole: The port check currently only validates the lower bound, so values over 65535 won't trigger the warning or be skipped. Maybe check the upper boundary too?
```suggestion
		if p.Port < 1 || p.Port > 65535 {
```

pkg/migrate/helpers.go
```diff
@@ -1276,3 +1278,73 @@ func combineAndConvertRelabelings(logger *slog.Logger, promoted []monitoringv1.R
...
+
+// convertServiceTargetLabels maps Service labels to static metricRelabeling rules.
+func convertServiceTargetLabels(logger *slog.Logger, svc *corev1.Service, targetLabels []string) []monitoringv1.RelabelingRule {
+	if svc == nil || len(targetLabels) == 0 {
+		return nil
+	}
+
+	var rules []monitoringv1.RelabelingRule
+
+	for _, l := range targetLabels {
```
dashpole, line 1341: If targetLabels contains duplicates or keys that sanitize to the exact same Prometheus label name (like app.kubernetes.io/name and app_kubernetes_io_name), this could generate conflicting relabeling rules for the same target. Consider tracking seen target labels with a map like in convertTargetLabels.

pkg/migrate/helpers_test.go
```diff
@@ -918,3 +926,403 @@ func TestToStrictUnstructured(t *testing.T) {
...
+			name: "Resolve by port number to targetPort int",
+			service: makeTestTypedService("default", "my-svc", nil, []corev1.ServicePort{
+				{Name: "web", Port: 80, TargetPort: intstr.FromInt32(8080)},
+			}),
+			portStr:  "80",
+			expected: intstr.FromInt32(8080),
+			wantErr:  false,
+		},
+		{
+			name: "Resolve by port number",
```
dashpole: This test case appears identical to the one right above it ("Resolve by port number to targetPort int"). Maybe update this one to test resolving a numeric port string to a named targetPort instead?

pkg/migrate/types.go
```diff
@@ -123,3 +128,47 @@ func (c *ResourceCache) Get(kind, namespace, name string) (*unstructured.Unstruc
 	r, ok := nsMap[key]
 	return r, ok
 }
+
+// findServicesBySelector finds Services matching the label selector within the specified namespaces.
+// Note for callers: passing nil or an empty namespaces slice matches Services across every namespace.
+// Callers should use determineNamespaceScoping to resolve default namespace rules before calling this method.
+func (c *ResourceCache) findServicesBySelector(selector metav1.LabelSelector, namespaces []string) ([]*corev1.Service, error) {
```
dashpole, line 207: Since ResourceCache is passed to external implementations of ResourceConverter, consider exporting findServicesBySelector (like Add and Get) so custom converters can also resolve services by label selector.

Or should we not export most of these types?

karthunni: I believe all converters would exist/be written in this package, thus we shouldn't export findServicesBySelector.

ResourceCache needs to stay exported as its a parameter for ResourceConverter.Convert. Having the core Add/Get accessors be exported seems like best practice for this

pkg/migrate/helpers.go
karthunni: I realize there might be a fundamental misunderstanding here. `resolveServicePort` is looking for a specific port on the `Service` I.e the one specified by the `ServiceMonitor`. So if any of the non-targeted ports have an issue, before we errored out, and after the change, we now log that there is an issue and we skip (even though we skip regardless since the port is irrelevant to us, we're basically just logging about an endpoint we don't care about).

Instead we should only check for malformations if the port matches the one we specified. Now the design choice here is whether we want to
- Error -> Fail the migration of this ServiceMonitor immediately.
- Warn and Drop -> Log a warning to skip this specific Service/endpoint, and continue migrating any other healthy Services/endpoints.

pkg/migrate/helpers_test.go
```diff
@@ -554,3 +563,778 @@ func TestMergeFromPod(t *testing.T) {
...
+			expected: intstr.IntOrString{},
+			wantErr:  true,
+		},
+		{
+			name:     "Nil service",
+			service:  nil,
+			portStr:  "web",
+			expected: intstr.IntOrString{},
+			wantErr:  true,
+		},
```
dashpole, line 1149: We test the svc == nil error guard here, but we could also add test cases for the other two early validation guards in resolveServicePort: an empty port string (portStr == "") and a service with no ports defined in its spec (len(svc.Spec.Ports) == 0).

pkg/migrate/helpers.go
```diff
@@ -911,3 +1000,361 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
...
+
+		if p.Name == portStr || fmt.Sprintf("%d", p.Port) == portStr {
+			if p.TargetPort.IntVal == 0 && p.TargetPort.StrVal == "" {
+				return intstr.FromInt32(p.Port), nil
+			}
+			return p.TargetPort, nil
+		}
+	}
+
+	return intstr.IntOrString{}, fmt.Errorf("port %q not found in Service spec", portStr)
```
dashpole, line 1329: Agree with failing fast here when the port cannot be resolved, since scraping without a valid target port will inevitably fail. To keep error handling semantics consistent across the migration tool, we should also align our secret and config resolution (e.g., in `extractResourceKey`) to fail fast when referenced Secrets or ConfigMaps are missing, rather than logging a warning and generating placeholder strings. Scraping without required authentication or TLS credentials will result in failed scrapes just like a missing port.

pkg/migrate/types.go
```diff
@@ -111,3 +128,47 @@ func (c *ResourceCache) Get(kind, namespace, name string) (*unstructured.Unstruc
...
```
dashpole, line 207: You are iterating across unindexed nested maps with explicit sorting (`slices.Sort`) and runtime reflection (`FromUnstructured`). Consider replacing the unindexed map structure in `ResourceCache` with a standard K8s Indexer (`k8s.io/client-go/tools/cache.Indexer`) configured with namespace and label indices. I think this should be cleaner, and type-safe.

dashpole (approved):
LGTM after the last open comment is addressed

pkg/migrate/types.go
karthunni: If i understand correctly, the benefit with cache indexer would be that i could index namespace and kind to get the groups of targeted services in constant time.
 
But if i wanted to remove the `FromUnstructured` call,  I would need to store the objects as typed instead of Unstructured right? Which means I would need a dedicated Service cache?

dashpole: You don't need a dedicated cache per type to eliminate `FromUnstructured`.

1. **Unified Storage (`runtime.Object` / `any`)**: `client-go/tools/cache.Indexer` stores `any` (`runtime.Object`), which accommodates both typed structs (`*corev1.Service`, `*corev1.Secret`) and `*unstructured.Unstructured` in the same indexer.
2. **Ingestion-Time Conversion**: If we convert known kinds (like `Service`) once when `Add()` is called, the cache can store the typed `*corev1.Service`. When `findServicesBySelector` runs, retrieving it is just an $O(1)$ Go type assertion (`obj.(*corev1.Service)`), eliminating all runtime reflection overhead during conversion.
3. **Namespace Indexing**: `cache.Indexer` allows indexing by namespace (`cache.MetaNamespaceIndexFunc`), so we only evaluate label selectors against services in the target namespace rather than looping across every service in the cache.

Here are some wrappers from upstream that should help make this simple:

1. **[`k8s.io/client-go/listers/generic_helpers.go`](https://github.com/kubernetes/client-go/blob/master/listers/generic_helpers.go)** — Shows `ResourceIndexer[T runtime.Object]`, which wraps `cache.Indexer` and returns typed slices (`[]T`) via $O(1)$ type assertion (`m.(T)`).
2. **[`k8s.io/client-go/listers/core/v1/service.go`](https://github.com/kubernetes/client-go/blob/master/listers/core/v1/service.go)** — Shows `NewServiceLister(indexer cache.Indexer)` instantiated directly over the indexer.
3. **[`k8s.io/client-go/tools/cache/listers.go`](https://github.com/kubernetes/client-go/blob/master/tools/cache/listers.go#L61)** — Shows `ListAllByNamespace`, which retrieves objects by namespace index in $O(1)$ before filtering by selector.

PR #2005, opened 2026-07-24
fix: gmpctl ensure correct tools are installed across old and new releases

bwplotka:
This should unblock the `release` command, so it works across branches and repos.

### DIFF
--- a/ops/gmpctl.sh
+++ b/ops/gmpctl.sh
@@ -26,6 +26,15 @@ SCRIPT_DIR="$(
 	pwd -P
 )"
 
+# Install dependencies in the current context. We switch Go toolchain versions so using go tool might not work.
+go install github.com/google/go-containerregistry/cmd/gcrane@latest
+go install github.com/mikefarah/yq/v4@latest
+go install helm.sh/helm/v3/cmd/helm@latest
+go install github.com/google/addlicense@latest
+go install golang.org/x/vuln/cmd/govulncheck@latest
+
+export PATH="$(go env GOPATH)/bin:${PATH}"
+
 # NOTE gmpctl expects the gmpctl directory to be present on execution for
 # local bash scripts and configuration.
 #
--- a/ops/gmpctl.sh
+++ b/ops/gmpctl.sh
@@ -26,14 +26,15 @@ SCRIPT_DIR="$(
 	pwd -P
 )"
 
-# Install dependencies in the current context. We switch Go toolchain versions so using go tool might not work.
+# Install dependencies in a local directory. We switch Go toolchain versions so using go tool might not work.
+export GOBIN="${SCRIPT_DIR}/.bin"
+export PATH="${GOBIN}:${PATH}"
+
 go install github.com/google/go-containerregistry/cmd/gcrane@latest
 go install github.com/mikefarah/yq/v4@latest
 go install helm.sh/helm/v3/cmd/helm@latest
 go install github.com/google/addlicense@latest
 go install golang.org/x/vuln/cmd/govulncheck@latest
-
-export PATH="$(go env GOPATH)/bin:${PATH}"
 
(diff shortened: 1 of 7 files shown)


ops/gmpctl.sh
```diff
@@ -26,6 +26,15 @@ SCRIPT_DIR="$(
...
 )"
 
+# Install dependencies in the current context. We switch Go toolchain versions so using go tool might not work.
+go install github.com/google/go-containerregistry/cmd/gcrane@latest
+go install github.com/mikefarah/yq/v4@latest
+go install helm.sh/helm/v3/cmd/helm@latest
+go install github.com/google/addlicense@latest
+go install golang.org/x/vuln/cmd/govulncheck@latest
+
+export PATH="$(go env GOPATH)/bin:${PATH}"
```
quentinmit: `$GOPATH` is a colon-separated list, so you can't just interpolate it here.

I suggest setting `$GOBIN` to a local directory instead, which will mean you can be sure where the binaries are, and also that you won't clobber any other versions the caller already has in their `$GOPATH`

ops/gmpctl/lib.sh
```diff
@@ -442,17 +440,21 @@ release-lib::manifests_regen() {
...
-	go install github.com/mikefarah/yq/v4@latest
-
 	# Hack: Do the bingo variable swap. This allows injecting our own.
 	# This is faster than running requiring bingo and running bingo get.
-	cp "${dir}/.bingo/variables.env" "${dir}/.bingo/variables.env.bak"
-	echo "#!/bin/bash" >"${dir}/.bingo/variables.env" # Clean the file.
+	# NOTE: Only needed before 0.19.
+	if [[ -f "${dir}/.bingo/variables.env" ]]; then
+		cp "${dir}/.bingo/variables.env" "${dir}/.bingo/variables.env.bak"
+		echo "#!/bin/bash" >"${dir}/.bingo/variables.env" # Clean the file.
```
quentinmit: I think the script committed this broken environment file for me because `presubmit.sh manifests` errored out. Instead I think

```shell
trap 'mv "${dir}/.bingo/variables.env.bak" "${dir}/.bingo/variables.env"; trap - RETURN' RETURN
```

should cause the file to be cleaned up when the function returns, whether or not `presubmit.sh` fails.

bwplotka: Yup, fixed!

ops/gmpctl.sh
bwplotka: done

ops/gmpctl/lib.sh
```diff
@@ -442,17 +440,17 @@ release-lib::manifests_regen() {
...
+  fi
 
-	# Hack: Do the bingo variable swap. This allows injecting our own.
-	# This is faster than running requiring bingo and running bingo get.
-	cp "${dir}/.bingo/variables.env" "${dir}/.bingo/variables.env.bak"
-	echo "#!/bin/bash" >"${dir}/.bingo/variables.env" # Clean the file.
-	YQ="$(which yq)" HELM="$(which helm)" ADDLICENSE="$(which addlicense)" bash "${dir}/hack/presubmit.sh" manifests
-	cp "${dir}/.bingo/variables.env.bak" "${dir}/.bingo/variables.env"
+  echo "🔄 Regenerating manifests..."
+  YQ="$(command -v yq)" HELM="$(command -v helm)" ADDLICENSE="$(command -v addlicense)" bash "${dir}/hack/presubmit.sh" manifests
```
quentinmit, line 453: Do you still need to rely on `$PATH`? Why can't you use `$GOBIN/yq` etc.?

quentinmit (approved):

ops/gmpctl/lib.sh
bwplotka: Let me fix in another one, fair point.

PR #2007, opened 2026-07-24
fix: remove severity check, make gmpctl more reliable

bwplotka:
This PR removes the NVDAPI flow and severity checking in gmpctl to make gmpctl vulnupdatelist more reliable.

Initially I had it so we prioritize HIGH vulns, but in practice we need all..


avilevy18 (approved):
LGTM

quentinmit (approved):

PR #2008, opened 2026-07-24
fix(ops): allow gmpctl to ignore specified modules; loop

bwplotka:
This PR improves `gmpctl vulnfix` reliability by adding module ignoring capabilities, fixing dependency parsing, and adding multi-pass vulnerability resolution.

### Changes Included

1. **Ignored Modules Support**:
   - Added `vuln_ignored_modules` setting in `.gmpctl.default.yaml`.
   - Added `-vuln-ignore-modules` flag to `gmpctl` and `vulnupdatelist` to skip specific Go module upgrades (such as `github.com/prometheus/prometheus`) during vulnerability remediation.

2. **Dependency Update Bug Fix (`lib.sh`)**:
   - Fixed `release-lib::gomod_vulnfix` parsing of `vulnupdatelist` output (using `$1` instead of `$2` to correctly extract `module@version` rather than CVE IDs).
   - Switched from line-matching `sed` to `go get "${mod_path}@${desired_version}"` to reliably update both direct and indirect dependencies in `go.mod`.

3. **Multi-Pass Vulnerability Resolution (`vulnfix.sh`)**:
   - Added an iterative loop (up to 5 attempts) in `vulnfix.sh` to handle chained/transitive vulnerability fixes (e.g., when updating a module reveals a higher patch requirement in subsequent scans).

4. **Tests**:
   - Added unit tests in `vulnupdatelist/vuln_test.go` verifying ignored module logic and update list compilation.


ops/gmpctl/vulnupdatelist/vuln.go
```diff
@@ -160,10 +151,26 @@ func compileUpdateList(jsonData io.Reader, onlyFixed bool) ([]UpdateList, error)
...
-		return updateList[i].Module < updateList[j].Module
+	allUpdates := slices.Collect(maps.Values(updates))
+	sort.Slice(allUpdates, func(i, j int) bool {
+		return allUpdates[i].Module < allUpdates[j].Module
 	})
+
+	var updateList []UpdateList
+	for _, up := range allUpdates {
+		if up.Ignored {
+			slog.Info("Ignoring module upgrade due to configuration", "module", up.Module)
```
dashpole: Since the configuration comment notes that skipping modules should produce a warning, consider logging as `slog.Warn` instead of `slog.Info` and including the ignored CVE ID for security visibility.
```suggestion
			slog.Warn("IMPORTANT: Found Go vulnerability in an ignored module; skipping upgrade...", "module", up.Module, "cve", up.CVEID)
```

ops/gmpctl/vulnupdatelist/main.go
```diff
@@ -52,6 +54,7 @@ type UpdateList struct {
 	Module         string
 	FixedVersion   *semver.Version
 	Version        string
+	Ignored        bool
```
dashpole: Since ignored modules are filtered out in `compileUpdateList` before returning the update list, every returned `UpdateList` struct always has `Ignored` set to false. Consider removing the `Ignored` field from this struct and simply checking if a module should be ignored directly during the loop in `vuln.go`.

ops/gmpctl/vulnfix.sh
```diff
@@ -100,23 +100,27 @@ echo "*" >>"${DIR}/.gmpctl/.gitignore"
...
+done
 
-	# Check if that helped.
-	echo "⚠️  This will fail on older branches with vendoring; in this case, simply go to ${DIR}, run 'go mod vendor' and rerun."
-	release-lib::vulnlist "${DIR}" "${vuln_file}"
-	if [[ "no vulnerabilities" != $(cat "${vuln_file}") ]]; then
-		echo "❌  After go mod update some vulnerabilities are still found; go to ${DIR} and resolve it manually (select not reusing the ./vulnlist.txt file) and rerun."
-		exit 1
-	fi
+release-lib::vulnlist "${DIR}" "${vuln_file}"
```
dashpole: When the loop breaks early after confirming no remaining vulnerabilities, executing `release-lib::vulnlist` down here redundantly re-runs `govulncheck` on an unchanged codebase. Consider wrapping this call in a condition so it only re-runs if vulnerabilities were still present when the loop exited.
```suggestion
if [[ "no vulnerabilities" != $(cat "${vuln_file}") ]]; then
	release-lib::vulnlist "${DIR}" "${vuln_file}"
fi
```

ops/gmpctl/vulnfix.sh
```diff
@@ -100,23 +100,27 @@ echo "*" >>"${DIR}/.gmpctl/.gitignore"
...
+	if [[ "no vulnerabilities" == $(cat "${vuln_file}") ]]; then
+		echo "✅ No more fixable Go vulnerabilities found!"
+		break
+	fi
 
-if [[ "no vulnerabilities" != $(cat "${vuln_file}") ]]; then
-	# Attempt to update + go mod tidy.
+	echo "🔄 Iteration $i/$max_attempts: Applying dependency updates..."
 	release-lib::gomod_vulnfix "${DIR}" "${vuln_file}"
+	pushd "${DIR}"
```
dashpole: Since `pushd` is already called before entering the loop and `release-lib::gomod_vulnfix` leaves the working directory unchanged, calling `pushd` and `popd` around these git commands is redundant and adds extra directory stack output during loop iterations.

ops/gmpctl/lib.sh
```diff
@@ -145,13 +146,14 @@ release-lib::gomod_vulnfix() {
...
 	while IFS= read -r line || [[ -n "$line" ]]; do
 		# Skip any empty lines in the input file.
 		if [ -z "$line" ]; then
 			continue
 		fi
 
-		mod=$(echo "$line" | awk '{print $2}')
+		mod=$(echo "$line" | awk '{print $1}')
 		mod_path=$(echo "${mod}" | cut -d'@' -f1)
 		desired_version=$(echo "${mod}" | cut -d'@' -f2)
```
dashpole: If an input line lacks an `@` delimiter, `cut` without `-s` returns the full string, which bypasses the empty check below and attempts an invalid `go get` command. Using `cut -s` ensures `desired_version` evaluates to empty for lines without an `@` symbol.
```suggestion
		desired_version=$(echo "${mod}" | cut -s -d'@' -f2)
```

ops/gmpctl/lib.sh
```diff
@@ -161,10 +163,9 @@ release-lib::gomod_vulnfix() {
 		fi
 
 		echo "🔄 Updating module '${mod_path}' to version '${desired_version}'..."
-		${SED} -i "s|\(	${mod_path} \).*|\1${desired_version}|" "${dir}/go.mod"
+		go get "${mod_path}@${desired_version}"
```
dashpole: Since `UpdateList.String()` formats unfixable vulnerabilities as `module@???`, executing `go get` directly will fail with a fatal command error if an unfixable entry reaches this point. Consider adding an explicit check before calling `go get` to verify that `desired_version` is a valid version string and not `???`.

ops/gmpctl/vulnupdatelist/vuln_test.go
```diff
@@ -0,0 +1,44 @@
...
+{"osv":{"id":"GO-2024-0001","summary":"Test vuln 1"}}
+{"finding":{"osv":"GO-2024-0001","fixed_version":"1.2.3","trace":[{"module":"github.com/prometheus/prometheus","version":"1.0.0"}]}}
+{"osv":{"id":"GO-2024-0002","summary":"Test vuln 2"}}
+{"finding":{"osv":"GO-2024-0002","fixed_version":"0.1.5","trace":[{"module":"golang.org/x/net","version":"0.1.0"}]}}
+`
+
+	t.Run("without ignored modules", func(t *testing.T) {
+		updates, err := compileUpdateList(strings.NewReader(mockJSON), false, nil)
+		require.NoError(t, err)
+		require.Len(t, updates, 2)
```
dashpole, line 35: In these test cases, checking only the slice length and module name leaves fields like `FixedVersion`, `Version`, and `CVEID` unverified. Consider adding field-level assertions on the returned structs to better guard against parsing regressions.

ops/gmpctl/lib.sh
bwplotka: tool never prints unfixable modules.

ops/gmpctl/lib.sh
```diff
@@ -145,26 +146,26 @@ release-lib::gomod_vulnfix() {
...
+		desired_version=$(echo "${mod}" | cut -s -d'@' -f2)
 
 		if [[ -z "${mod_path}" ]] || [[ -z "${desired_version}" ]]; then
 			echo "⚠️ Skipping malformed line: $line"
 			continue
 		fi
 
 		echo "🔄 Updating module '${mod_path}' to version '${desired_version}'..."
-		${SED} -i "s|\(	${mod_path} \).*|\1${desired_version}|" "${dir}/go.mod"
+		go get "${mod_path}@${desired_version}"
```
bwplotka: same as https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2015/changes#r3728747023 but probably ok

dashpole (approved):

PR #2010, opened 2026-07-24
fix: remove severity check, make gmpctl more reliable 

bwplotka:
Re applying https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2007 that was approved but I merged to a wrong branch 🙃


bwplotka:
Merging as it was approved in https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2007

PR #2011, opened 2026-07-24
fix(ops): ignore -linux-* suffixes when fetching distroless image tags

bwplotka:
Ignore `-linux-*` architecture-specific tag suffixes when searching for latest distroless tags in gmpctl.

### DIFF
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -373,7 +373,7 @@ release-lib::dockerfile_update_image() {
 	# Exclude RC images.
 	all_tags=$(echo "${all_tags}" | grep -v "rc.*")
 	# Ignore -linux-* architecture suffixes (e.g. -linux-arm64).
-	all_tags=$(echo "${all_tags}" | grep -v "\-linux\-")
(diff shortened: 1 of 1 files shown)


avilevy18 (approved):
LGTM!

AI mentioned a minor nit regarding the `.bingo/variables.env` trap cleanup in `release-lib::manifests_regen()`:

Using a `RETURN` trap (`trap '...' RETURN`) in Bash functions automatically executes cleanup whenever the function finishes or returns (on both success or failure). This avoids needing `trap - EXIT` or touching process-wide `EXIT` traps:

```bash
if [[ -f "${dir}/.bingo/variables.env" ]]; then
    cp "${dir}/.bingo/variables.env" "${dir}/.bingo/variables.env.bak"
    trap 'mv "${dir}/.bingo/variables.env.bak" "${dir}/.bingo/variables.env" 2>/dev/null || true' RETURN
    echo "#!/bin/bash" >"${dir}/.bingo/variables.env" # Clean the file.
fi
```

Non-blocking, feel free to keep as-is or adjust.

PR #2015, opened 2026-07-25
fix(ops): group OpenTelemetry module upgrades and sync semconv imports in Go files

bwplotka:
## Description

This PR fixes vulnerability updates for OpenTelemetry modules in `gmpctl`:

1. **Simultaneous Module Upgrades Across Dependency Tree**: Uses `go list -m all` to discover all direct and indirect `go.opentelemetry.io/otel*` modules (such as `otel/trace`, `otel/sdk`, `otel/metric`, etc.) and upgrades them simultaneously to avoid version drift and partial module upgrades.
2. **Dynamic `semconv` Import Synchronization**: Dynamically queries the exact `semconv` package version imported by `go.opentelemetry.io/otel/sdk/resource` and updates hardcoded `semconv` imports in `.go` files. This prevents `conflicting Schema URL` errors during tracer provider initialization (e.g. `failed to install a new tracer provider: error detecting resource: conflicting Schema URL: https://opentelemetry.io/schemas/1.40.0 and https://opentelemetry.io/schemas/1.41.0`).

Signed-off-by: bwplotka <bwplotka@google.com>

### DIFF
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -181,15 +181,6 @@ release-lib::gomod_vulnfix() {
 				go get ${otel_args}
 			fi
 
-			# OpenTelemetry SDK resource detectors (e.g. WithProcessRuntimeDescription, WithTelemetrySDK)
-			# use the semconv version imported by otel/sdk/resource. We update hardcoded semconv imports in Go
-			# files to match the SDK semconv package version to prevent conflicting Schema URL errors during tracer provider
-			# initialization (e.g. "failed to install a new tracer provider: error detecting resource: conflicting Schema URL:...").
-			sdk_semconv=$(go list -e -f '{{ join .Imports "\n" }}' go.opentelemetry.io/otel/sdk/resource 2>/dev/null | grep 'go.opentelemetry.io/otel/semconv/v' | head -n1 || true)
-			if [[ -n "${sdk_semconv}" ]]; then
-				echo "🔄 Updating semconv imports in Go files to match SDK (${sdk_semconv})..."
-				find . -name "*.go" -not -path "*/vendor/*" -exec ${SED} -i -E "s#go\.opentelemetry\.io/otel/semconv/v1\.[0-9]+\.[0-9]+#${sdk_semconv}#g" {} +
-			fi
 		else
 			echo "🔄 Updating module '${mod_path}' to version '${desired_version}'..."
 			go get "${mod_path}@${desired_version}"
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -168,7 +168,7 @@ release-lib::gomod_vulnfix() {
 		if [[ "${mod_path}" == go.opentelemetry.io/otel* ]]; then
 			# OpenTelemetry core API/SDK modules share versions and schema URLs across packages (e.g. otel, otel/sdk, otel/trace, otel/metric).
 			# Upgrade core otel modules present in the module graph together to avoid conflicting schema URL errors.
-			otel_mods=$(go list -m all 2>/dev/null | awk '{print $1}' | grep -E '^go\.opentelemetry\.io/otel($|/sdk$|/trace$|/metric$|/sdk/metric$|/log$|/sdk/log$)' || true)
+			otel_mods=$(go list -m all 2>/dev/null | awk '/^go\.opentelemetry\.io\/otel($|\/)/ && !/^go\.opentelemetry\.io\/otel\/(contrib|semconv)/ {print $1}')
 			all_otel=$(echo "${mod_path} ${otel_mods}" | tr ' ' '\n' | sort -u)
 			otel_args=""
 			for m in $(echo "${all_otel}" | tr ' ' '\n'); do
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -176,8 +176,10 @@ release-lib::gomod_vulnfix() {
 					otel_args="${otel_args} ${m}@${desired_version}"
 				fi
 			done
-			echo "🔄 Updating OpenTelemetry modules simultaneously:${otel_args}..."
-			go get ${otel_args}
+			if [[ -n "${otel_args// /}" ]]; then
+				echo "🔄 Updating OpenTelemetry modules simultaneously:${otel_args}..."
+				go get ${otel_args}
+			fi
 
 			# OpenTelemetry SDK resource detectors (e.g. WithProcessRuntimeDescription, WithTelemetrySDK)
 			# use the semconv version imported by otel/sdk/resource. We update hardcoded semconv imports in Go
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -158,7 +158,7 @@ release-lib::gomod_vulnfix() {
 
 		mod=$(echo "$line" | awk '{print $1}')
 		mod_path=$(echo "${mod}" | cut -d'@' -f1)
-		desired_version=$(echo "${mod}" | cut -d'@' -f2)
+		desired_version=$(echo "${mod}" | cut -s -d'@' -f2)
 
 		if [[ -z "${mod_path}" ]] || [[ -z "${desired_version}" ]]; then
 			echo "⚠️ Skipping malformed line: $line"
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -143,6 +143,10 @@ release-lib::gomod_vulnfix() {
 		return 1
 	fi
 
+	if [[ "${vuln_file}" != /* ]]; then
+		vuln_file="$(pwd)/${vuln_file}"
+	fi
+
 	# Read the vulnerability file line by line.
 	# The `|| [[ -n "$line" ]]` part handles the case where the last line doesn't have a newline.
 	pushd "${dir}"
--- a/ops/gmpctl/lib.sh
+++ b/ops/gmpctl/lib.sh
@@ -145,13 +145,14 @@ release-lib::gomod_vulnfix() {
 
 	# Read the vulnerability file line by line.
 	# The `|| [[ -n "$line" ]]` part handles the case where the last line doesn't have a newline.
+	pushd "${dir}"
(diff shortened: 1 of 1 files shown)


ops/gmpctl/lib.sh
```diff
@@ -160,11 +161,33 @@ release-lib::gomod_vulnfix() {
 			continue
 		fi
 
-		echo "🔄 Updating module '${mod_path}' to version '${desired_version}'..."
-		${SED} -i "s|\(	${mod_path} \).*|\1${desired_version}|" "${dir}/go.mod"
+		if [[ "${mod_path}" == go.opentelemetry.io/otel* ]]; then
+			# OpenTelemetry modules share versions and schema URLs across packages (e.g. otel, otel/sdk, otel/trace, otel/metric).
+			# Upgrade all otel modules present in go.mod together to avoid conflicting schema URL errors.
```
bwplotka: # As a best practice we want to upgrade all otel modules to the same version.

ops/gmpctl/lib.sh
```diff
@@ -160,11 +161,33 @@ release-lib::gomod_vulnfix() {
...
+			otel_mods=$(grep -E '^\s*go\.opentelemetry\.io/otel' go.mod | awk '{print $1}')
+			all_otel=$(echo "${mod_path} ${otel_mods}" | tr ' ' '\n' | sort -u)
+			otel_args=""
+			for m in ${all_otel}; do
+				otel_args="${otel_args} ${m}@${desired_version}"
+			done
+			echo "🔄 Updating OpenTelemetry modules simultaneously:${otel_args}..."
+			go get ${otel_args}
+
+			# OpenTelemetry SDK resource detectors (e.g. WithProcessRuntimeDescription, WithTelemetrySDK)
```
bwplotka: I am assuming we can equally in our fork remove the explicit schema at this point? 🤔 

Image

https://github.com/GoogleCloudPlatform/prometheus/blob/a24099b1f7e42d86bae4e911d245fdcf932f7cff/tracing/tracing.go#L146

bwplotka: Ok, we can delete the explicit scheme - confirmed with @dashpole  https://github.com/GoogleCloudPlatform/prometheus/blob/a24099b1f7e42d86bae4e911d245fdcf932f7cff/tracing/tracing.go#L33

ops/gmpctl/lib.sh
```diff
@@ -145,13 +145,14 @@ release-lib::gomod_vulnfix() {
 
 	# Read the vulnerability file line by line.
 	# The `|| [[ -n "$line" ]]` part handles the case where the last line doesn't have a newline.
+	pushd "${dir}"
```
dashpole, line 152: Moving `pushd` before the `while` loop means relative `vuln_file` paths will fail at `done <"${vuln_file}"` since the working directory has changed. Resolving `vuln_file` to an absolute path before entering the loop ensures redirection always succeeds:
```suggestion
	if [[ "${vuln_file}" != /* ]]; then
		vuln_file="$(pwd)/${vuln_file}"
	fi
	pushd "${dir}"
```

ops/gmpctl/lib.sh
```diff
@@ -160,11 +161,35 @@ release-lib::gomod_vulnfix() {
 			continue
 		fi
 
-		echo "🔄 Updating module '${mod_path}' to version '${desired_version}'..."
-		${SED} -i "s|\(	${mod_path} \).*|\1${desired_version}|" "${dir}/go.mod"
+		if [[ "${mod_path}" == go.opentelemetry.io/otel* ]]; then
+			# OpenTelemetry core API/SDK modules share versions and schema URLs across packages (e.g. otel, otel/sdk, otel/trace, otel/metric).
+			# Upgrade core otel modules present in the module graph together to avoid conflicting schema URL errors.
+			otel_mods=$(go list -m all 2>/dev/null | awk '{print $1}' | grep -E '^go\.opentelemetry\.io/otel($|/sdk$|/trace$|/metric$|/sdk/metric$|/log$|/sdk/log$)')
```
dashpole: If no matching core OpenTelemetry modules are found in the module graph, `grep` will exit with status 1 and terminate the script due to `set -o pipefail`. Appending `|| true` prevents this crash when zero matches occur:
```suggestion
			otel_mods=$(go list -m all 2>/dev/null | awk '{print $1}' | grep -E '^go\.opentelemetry\.io/otel($|/sdk$|/trace$|/metric$|/sdk/metric$|/log$|/sdk/log$)' || true)
```

ops/gmpctl/lib.sh
```diff
@@ -160,11 +161,35 @@ release-lib::gomod_vulnfix() {
...
+			all_otel=$(echo "${mod_path} ${otel_mods}" | tr ' ' '\n' | sort -u)
+			otel_args=""
+			for m in $(echo "${all_otel}" | tr ' ' '\n'); do
+				if go list -m "${m}@${desired_version}" >/dev/null 2>&1; then
+					otel_args="${otel_args} ${m}@${desired_version}"
+				fi
+			done
+			echo "🔄 Updating OpenTelemetry modules simultaneously:${otel_args}..."
+			go get ${otel_args}
```
dashpole: If none of the candidate modules pass the `go list` check, `otel_args` will remain empty and executing parameterless `go get` will fail with exit code 1 in modern Go module mode. Guarding the invocation avoids terminating the script when no packages match:
```suggestion
			if [[ -n "${otel_args// /}" ]]; then
				go get ${otel_args}
			fi
```

ops/gmpctl/lib.sh
bwplotka: Thanks! Applied in commit `7b21619e0` to resolve relative `vuln_file` paths to absolute paths before `pushd`.

ops/gmpctl/lib.sh
bwplotka: Thanks! Applied in commit `6ac7c1392` by appending `|| true` to `grep` pipelines to prevent failure under `set -o pipefail`.

ops/gmpctl/lib.sh
bwplotka: Thanks! Applied in commit `570505e70` to guard `go get ${otel_args}` with a non-empty string check.

ops/gmpctl/lib.sh
```diff
@@ -143,15 +143,20 @@ release-lib::gomod_vulnfix() {
...
 	while IFS= read -r line || [[ -n "$line" ]]; do
 		# Skip any empty lines in the input file.
 		if [ -z "$line" ]; then
 			continue
 		fi
 
-		mod=$(echo "$line" | awk '{print $2}')
+		mod=$(echo "$line" | awk '{print $1}')
 		mod_path=$(echo "${mod}" | cut -d'@' -f1)
 		desired_version=$(echo "${mod}" | cut -d'@' -f2)
```
dashpole: By default cut -d'@' -f2 outputs the whole input string if the @ delimiter is missing, which would bypass the malformed line check below. Using cut -s -d'@' -f2 suppresses output when @ is absent, correctly setting desired_version to an empty string on malformed lines:
```suggestion
		desired_version=$(echo "${mod}" | cut -s -d'@' -f2)
```

ops/gmpctl/lib.sh
```diff
@@ -160,11 +165,28 @@ release-lib::gomod_vulnfix() {
 			continue
 		fi
 
-		echo "🔄 Updating module '${mod_path}' to version '${desired_version}'..."
-		${SED} -i "s|\(	${mod_path} \).*|\1${desired_version}|" "${dir}/go.mod"
+		if [[ "${mod_path}" == go.opentelemetry.io/otel* ]]; then
+			# OpenTelemetry core API/SDK modules share versions and schema URLs across packages (e.g. otel, otel/sdk, otel/trace, otel/metric).
+			# Upgrade core otel modules present in the module graph together to avoid conflicting schema URL errors.
+			otel_mods=$(go list -m all 2>/dev/null | awk '{print $1}' | grep -E '^go\.opentelemetry\.io/otel($|/sdk$|/trace$|/metric$|/sdk/metric$|/log$|/sdk/log$)' || true)
```
dashpole: Core OpenTelemetry exporters share schema URLs and versions with the SDK, so we should group them here too. We can also filter directly in awk to avoid swallowing fatal go list errors under pipefail:
```suggestion
			otel_mods=$(go list -m all 2>/dev/null | awk '/^go\.opentelemetry\.io\/otel($|\/)/ && !/^go\.opentelemetry\.io\/otel\/(contrib|semconv)/ {print $1}')
```

ops/gmpctl/lib.sh
```diff
@@ -160,11 +165,28 @@ release-lib::gomod_vulnfix() {
...
+			all_otel=$(echo "${mod_path} ${otel_mods}" | tr ' ' '\n' | sort -u)
+			otel_args=""
+			for m in $(echo "${all_otel}" | tr ' ' '\n'); do
+				if go list -m "${m}@${desired_version}" >/dev/null 2>&1; then
+					otel_args="${otel_args} ${m}@${desired_version}"
+				fi
+			done
+			if [[ -n "${otel_args// /}" ]]; then
```
dashpole, line 179: If none of the OpenTelemetry modules pass the go list -m check for the desired version, otel_args will be empty and the upgrade will be silently skipped without any error or warning. We should add an else block here to log an error and return if we can't resolve any module targets.

dashpole (approved):

ops/gmpctl/lib.sh
```diff
@@ -143,28 +143,54 @@ release-lib::gomod_vulnfix() {
...
+				echo "🔄 Updating OpenTelemetry modules simultaneously:${otel_args}..."
+				go get ${otel_args}
```
bwplotka, line 181: BTW: sed was on purpose here -- it limits the chained updates  related to this dep. go mod tidy after sed-ing scopes updates only to things that truly are needed. 

e.g. client_golang update does not need latest 0.x common, and breaking changes on those occur.

```
${SED} -i "s|\(	${mod_path} \).*|\1${desired_version}|" "${dir}/go.mod"
```

Something to keep in mind, but we could try with this, especially AI is forcing this all the time ;p

PR #2022, opened 2026-07-29
PO to GMP Migration Tool: ServiceMonitor Spec Migration

karthunni:
* **ServiceMonitor Conversion & Registration:** Implements `ServiceMonitorConverter` in [pkg/migrate/servicemonitor.go](file:///usr/local/google/home/kunnikrishnan/Desktop/prometheus-engine/pkg/migrate/servicemonitor.go) and registers it in the CLI entrypoint ([cmd/gmp-migrate/main.go](file:///usr/local/google/home/kunnikrishnan/Desktop/prometheus-engine/cmd/gmp-migrate/main.go)).
* **Conflict-Aware Resource Splitting:** Resolves backing Kubernetes `Service` resources from the `ResourceCache` matching `spec.selector`. Groups Services by configuration compatibility (selectors, port mappings, labels) and automatically splits conflicting Services into distinct, uniquely-named `PodMonitoring` resources.
* **Namespace & Scoping Resolution:** Evaluates `spec.namespaceSelector` to generate either namespace-scoped `PodMonitoring` resources (cloned across all targeted namespaces) or cluster-scoped `ClusterPodMonitoring` resources when `any: true` is specified.
* **Endpoint & Label Translation:** Translates endpoint ports, paths, schemes, intervals, timeouts, and authentication configurations. Merges Service target labels (`spec.targetLabels`), Pod target labels (`spec.podTargetLabels`), `spec.jobLabel`, and pre-scrape relabeling rules into GMP scrape endpoints and label selectors.
* **Namespace-Scoped Secret Generation:** Ensures generated authentication secrets (bearer tokens, basic auth, TLS certificates) are properly scoped and cloned per target namespace ([pkg/migrate/helpers.go](file:///usr/local/google/home/kunnikrishnan/Desktop/prometheus-engine/pkg/migrate/helpers.go)).
* **Testing:** Adds table-driven unit tests in [pkg/migrate/servicemonitor_test.go](file:///usr/local/google/home/kunnikrishnan/Desktop/prometheus-engine/pkg/migrate/servicemonitor_test.go) covering multi-service conversion, conflict splitting, cluster-scoped monitoring, multi-namespace cloning, and secret generation.


pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,492 @@
...
+		for _, ep := range endpoints {
+			if ep.Port == "" {
+				continue
+			}
+			resolvedPort, err := resolveServicePort(svc, ep.Port)
+			if err != nil {
+				return nil, fmt.Errorf("service %q: failed to resolve port %q: %w", svc.GetName(), ep.Port, err)
+			}
+			portMap[ep.Port] = resolvedPort
+		}
```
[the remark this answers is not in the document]
karthunni: We want to fail in all of the above cases

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+	// 3. Resolve backing Services and generate base resources (handles splitting).
+	baseResources, generatedSecrets, err := c.convertToPodMonitoring(&serviceMonitor, logger, cache, targetNamespaces)
+	if err != nil {
+		return nil, err
+	}
+
+	// 4. Clone base resources for each target namespace.
+	var outputs []*unstructured.Unstructured
+	for _, base := range baseResources {
+		for _, ns := range targetNamespaces {
```
dashpole: Cloning every split group into all `targetNamespaces` creates a cross-product where PodMonitoring resources are placed in namespaces where their Service doesn't exist. Consider tracking target namespaces per `ServiceGroup` so each group is only emitted into namespaces of the Services that formed it.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+	}
+
+	if len(serviceTargetLabelRules) > 0 {
+		for i := range endpoints {
+			endpoints[i].MetricRelabeling = append(endpoints[i].MetricRelabeling, serviceTargetLabelRules...)
+		}
+	}
+
+	// Merge Pod target labels and selector.
+	mergedFromPod := mergeFromPod(logger, convertTargetLabels(logger, sm.Spec.PodTargetLabels, sm.Spec.JobLabel, "Pod"), rules.ResourceCombined.FromPod)
```
dashpole: In ServiceMonitor, `jobLabel` selects a label from the Service object rather than the Pod. Passing `sm.Spec.JobLabel` here reads it as a pod target label, so consider resolving it statically from the Service's labels alongside `spec.targetLabels` instead.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+	var rules []monitoringv1.RelabelingRule
+	var keys []string
+	for k := range labels {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+
+	for _, k := range keys {
+		v := labels[k]
+		target := k
```
dashpole: Kubernetes label keys like `app.kubernetes.io/name` need to be sanitized to valid Prometheus label names (invalid characters replaced with `_`) before setting `TargetLabel`. Also make sure to run the protected label check after sanitizing so keys like `app.kubernetes.io/job` don't bypass it. Also nit: `Action: "replace"` can use `string(relabel.Replace)`.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+	mergedSelector, err := mergeLabelSelector(baseSelector, rules.ResourceCombined.MatchLabels, rules.ResourceCombined.MatchExpressions)
+	if err != nil {
+		return nil, err
+	}
+
+	// Spec-level warnings.
+	resolveScrapeClass(sm.Spec.ScrapeClassName, logger)
+	validateScrapeProtocols(sm.Spec.ScrapeProtocols, logger)
+
+	metadata := resolveAttachMetadata(sm.Spec.AttachMetadata, rules.ResourceCombined.Metadata)
```
dashpole: This only calls `resolveAttachMetadata`, skipping default unioning and namespace label stripping done in `PodMonitorConverter`. Refactoring the metadata handling from `podmonitor.go` into a shared helper would ensure default fields aren't dropped and unallowed metadata fields aren't emitted.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+	epResults []PreScrapeRelabelingResult,
+	group *ServiceGroup,
+) ([]monitoringv1.ScrapeEndpoint, error) {
+	var gmpEndpoints []monitoringv1.ScrapeEndpoint
+
+	for i, ep := range endpoints {
+		gmpEp := monitoringv1.ScrapeEndpoint{}
+
+		// 1. Port mapping (Use pre-resolved port from group).
+		if ep.Port == "" {
```
dashpole: ServiceMonitors in the wild may use `TargetPort` instead of `Port`. We should support `TargetPort` fallback similar to `PodMonitorConverter.convertEndpoints`, or provide a clearer error message indicating `TargetPort` isn't supported.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+
+	// ClusterPodMonitoring is cluster-scoped and cannot be split by namespace.
+	// We fall back to using the first group and ignore the others.
+	if len(groups) > 1 {
+		logger.Warn("Multiple incompatible Service groups found for ClusterPodMonitoring. Using the first group and ignoring others.",
+			slog.Int("total_groups", len(groups)),
+			slog.String("used_group_service", groups[0].Services[0].GetName()))
+	}
+
+	res, err := c.buildSpecForGroup(sm, logger, cache, groups[0], true)
```
dashpole: If `groups` is empty (e.g., if missing Services are skipped in the future), accessing `groups[0]` will panic. Adding a check for `len(groups) == 0` here would keep this path safe.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+			if !reflect.DeepEqual(existingTargetPort, targetPort) {
+				return false
+			}
+		}
+	}
+
+	// 3. Check for target label conflicts.
+	for labelName, labelVal := range targetLabels {
+		if existingVal, exists := g.TargetLabels[labelName]; exists {
+			if existingVal != labelVal {
```
dashpole: Checking target labels only when both Services carry the key allows labeled and unlabeled Services to merge into the same group, applying the label to both. Consider checking for label presence/absence or splitting when label keys don't match.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+
+	// Apply Service targetLabels (statically resolved for this group).
+	var serviceTargetLabelRules []monitoringv1.RelabelingRule
+	if len(group.TargetLabels) > 0 {
+		serviceTargetLabelRules = convertStaticTargetLabels(logger, group.TargetLabels)
+	}
+
+	if len(serviceTargetLabelRules) > 0 {
+		for i := range endpoints {
+			endpoints[i].MetricRelabeling = append(endpoints[i].MetricRelabeling, serviceTargetLabelRules...)
```
dashpole: Target label rules are currently appended after user `metricRelabelConfigs`. Prepending them instead ensures user relabeling rules can operate on Service target labels as expected.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+	for _, group := range groups {
+		res, err := c.buildSpecForGroup(sm, logger, cache, group, false)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		name := sm.Name
+		if len(groups) > 1 {
+			// Suffix with the Service name to guarantee resource uniqueness when split.
+			name = fmt.Sprintf("%s-%s", sm.Name, group.Services[0].GetName())
```
dashpole: Concatenating `%s-%s` can produce resource names that exceed the 63-character DNS-1123 limit. Consider truncating or hashing when the generated name exceeds the maximum allowed length.

pkg/migrate/helpers.go
```diff
@@ -519,7 +525,11 @@ func (c *conversionContext) convertSecretSelector(sel *corev1.SecretKeySelector)
 		c.logger.Warn("Secret reference had 'optional: true'. GMP does not support optional secrets. The reference is now mandatory.",
 			slog.String("secret", sel.Name))
 	}
-	secretRef := &monitoringv1.SecretKeySelector{Name: sel.Name, Key: sel.Key, Namespace: c.namespace}
+	var ns string
+	if c.isClusterScoped {
+		ns = c.namespace
```
dashpole: Leaving `Namespace` empty for namespaced monitors is correct per API spec, but non-generated secrets aren't cloned into targeted namespaces during multi-namespace conversion. A warning suggesting users manually replicate existing referenced secrets into target namespaces would be helpful here.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+
+// canMergeWith checks if a Service's resolved selector, ports, and labels
+// are compatible with this ServiceGroup (i.e. no conflicts).
+func (g *ServiceGroup) canMergeWith(
+	selector map[string]string,
+	portMap map[string]intstr.IntOrString,
+	targetLabels map[string]string,
+) bool {
+	// 1. Selector must match exactly.
+	if !reflect.DeepEqual(g.Selector, selector) {
```
dashpole: Quick cleanup nit: `reflect.DeepEqual` can be replaced with `maps.Equal` for `g.Selector`, and line 474 manual map key sorting can use `slices.Sorted(maps.Keys(labels))`.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,500 @@
...
+		mergedFromPod:    mergedFromPod,
+		mergedSelector:   mergedSelector,
+		metadata:         metadata,
+		filterRunning:    filterRunning,
+		limits:           limits,
+		generatedSecrets: convCtx.getGeneratedSecrets(),
+	}, nil
+}
+
+func (c *ServiceMonitorConverter) convertEndpointsForGroup(
```
dashpole, line 317: Unlike `PodMonitorConverter.convertEndpoints`, this method is missing a guard check for `len(epResults) != len(endpoints)`. Adding that check prevents out-of-bounds indexing if slice lengths ever mismatch.

pkg/migrate/servicemonitor.go
dashpole: I believe this is still correct, and isn't the same mistake as https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1988#discussion_r3683427851.

pkg/migrate/servicemonitor.go
dashpole: Could also be left as a TODO if we don't think this will be common

pkg/migrate/servicemonitor.go
dashpole: Or just drop with an error? If we cut off the suffix, we might end up with a non-unique resource name...

pkg/migrate/servicemonitor_test.go
```diff
@@ -0,0 +1,416 @@
...
+					Namespace: "default",
+				},
+				Spec: pomonitoringv1.ServiceMonitorSpec{
+					Selector: metav1.LabelSelector{MatchLabels: map[string]string{"app": "foo"}},
+					Endpoints: []pomonitoringv1.Endpoint{
+						{Port: "web"},
+					},
+				},
+			},
+			verify: func(t *testing.T, outputs []*unstructured.Unstructured) {
```
bwplotka: Why not having an expected PodMonitoring object here? (declarative approach)

Could be more readable 🤔

pkg/migrate/servicemonitor.go
karthunni: Decided to a create a unique hash suffix instead

pkg/migrate/servicemonitor.go
karthunni: Checking with map.Equal instead

pkg/migrate/helpers.go
```diff
@@ -1348,3 +1418,29 @@ func convertServiceTargetLabels(logger *slog.Logger, svc *corev1.Service, target
...
+// makeUniqueResourceName joins a base name and suffix with a hyphen.
+// If the resulting name exceeds validation.DNS1123LabelMaxLength (63 characters),
+// it truncates the base name and appends a 6-character deterministic hash of the suffix.
+func makeUniqueResourceName(base, suffix string) string {
+	name := fmt.Sprintf("%s-%s", base, suffix)
+	if len(name) <= validation.DNS1123LabelMaxLength {
+		return name
+	}
+	h := fnv.New32a()
+	h.Write([]byte(suffix))
```
dashpole: If two long base names are identical up to the 56th character and use the same suffix, hashing only the suffix causes a naming collision when truncated. Consider hashing the complete combined name instead so any dropped characters from base are captured in the hash:
```suggestion
	h.Write([]byte(name))
```

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,516 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// ServiceGroup groups Services that share compatible selectors, ports, and labels.
+// This allows them to be merged into a single output PodMonitoring resource.
+type ServiceGroup struct {
```
dashpole: Since ServiceGroup and its Namespaces() method are only used internally for converting service monitors within this package, consider making serviceGroup unexported to keep the package API surface minimal.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,516 @@
...
+) (*unstructured.Unstructured, []*unstructured.Unstructured, error) {
+	groups, err := c.findAndGroupServices(sm, nil, logger, cache)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// ClusterPodMonitoring is cluster-scoped and cannot be split by namespace.
+	// We fall back to using the first group and ignore the others.
+	if len(groups) > 1 {
+		logger.Warn("Multiple incompatible Service groups found for ClusterPodMonitoring. Using the first group and ignoring others.",
```
dashpole: While ClusterPodMonitoring cannot be split by namespace, groups here represent services with incompatible selectors, ports, or target labels rather than namespaces. Instead of only using groups[0] and silently dropping scrape targets for subsequent service groups, consider returning a slice of cluster pod monitoring resources suffixed with the service name just like convertToPodMonitoring does.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,516 @@
...
+		if len(svc.Spec.Selector) == 0 {
+			return nil, fmt.Errorf("service %q has no selector (targets external or static endpoints). GMP Managed Collection only supports scraping in-cluster Pod targets", svc.Name)
+		}
+
+		// 2. Resolve ports for this Service.
+		portMap := make(map[string]intstr.IntOrString)
+		for _, ep := range sm.Spec.Endpoints {
+			portKey := endpointPortKey(ep)
+			if portKey == "" {
+				continue
```
dashpole: Quick nit: if portKey is empty here, convertEndpointsForGroup will return an error later on line 322. Returning an error immediately here rather than skipping via continue keeps endpoint validation consistent.

pkg/migrate/servicemonitor_test.go
```diff
@@ -0,0 +1,848 @@
...
+	"testing"
+
+	monitoringv1 "github.com/GoogleCloudPlatform/prometheus-engine/pkg/operator/apis/monitoring/v1"
+	"github.com/google/go-cmp/cmp"
+	pomonitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus/prometheus/model/relabel"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
```
dashpole, line 30: It looks like test cases for cluster-scoped service monitors (NamespaceSelector with Any: true) and secret generation or cloning across multiple target namespaces aren't included in this test suite yet. Adding test coverage for those execution paths in convertToClusterPodMonitoring and multi-namespace secret replication would be super helpful.

pkg/migrate/servicemonitor.go
karthunni: Would we just suffix the resource name with the namespace of the first Service in the ServiceGroup?

dashpole: Wouldn't they still collide since the namespace is the same for all of these?

> groups here represent services with incompatible selectors, ports, or target labels rather than namespaces

pkg/migrate/servicemonitor_test.go
karthunni: This might be better in a follow up PR for testing

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,512 @@
...
+		name := sm.Name
+		if len(groups) > 1 {
+			// Suffix with the Service name to guarantee resource uniqueness when split.
+			name = makeUniqueResourceName(sm.Name, group.Services[0].Name)
+		}
+
+		meta := sm.ObjectMeta.DeepCopy()
+		meta.Name = name
+
+		for _, ns := range group.Namespaces() {
```
dashpole: When generating monitors across multiple namespaces, `PodMonitorConverter` unconditionally iterates over all `targetNamespaces` and runs conversion cleanly in each namespace context (`convCtx.namespace = ns`). Here, conversion runs once using `sm.Namespace`, and resources are only emitted into `group.Namespaces()` (namespaces where backing Services are found in the offline cache). If a target namespace has no Service in the offline input bundle, no resource is generated for it. Consider verifying if this difference between unconditional generation (`PodMonitor`) and cache-dependent generation (`ServiceMonitor`) is intentional, and note that building `convCtx` within each target namespace loop would keep the conversion context fully aligned with `PodMonitor`.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+
+	// Merge Pod target labels and selector.
+	mergedFromPod := mergeFromPod(logger, convertTargetLabels(logger, sm.Spec.PodTargetLabels, "", "Pod"), rules.ResourceCombined.FromPod)
+
+	baseSelector := metav1.LabelSelector{MatchLabels: group.Selector}
+	mergedSelector, err := mergeLabelSelector(baseSelector, rules.ResourceCombined.MatchLabels, rules.ResourceCombined.MatchExpressions)
+	if err != nil {
+		return nil, err
+	}
```
dashpole, line 282: In `PodMonitorConverter.convertMonitorSpec`, an explicit defensive check follows label selector merging to warn when the resulting selector is empty (as it would scrape all pods across the namespace or cluster). Consider adding that same defensive warning log here after `mergeLabelSelector` completes to keep empty-selector safeguards consistent.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,512 @@
...
+
+	// Spec-level warnings.
+	resolveScrapeClass(sm.Spec.ScrapeClassName, logger)
```
dashpole, line 293: `PodMonitorConverter` explicitly calls `warnUnsupportedSpecFields` to notify users when `targetLimit`, `keepDroppedTargets`, or `bodySizeLimit` are configured so they know GMP drops or automatically manages them. Consider generalizing `warnUnsupportedSpecFields` in `helpers.go` or adding a companion `warnUnsupportedServiceMonitorSpecFields(logger, &sm.Spec)` call here to keep spec-level warnings consistent between `PodMonitor` and `ServiceMonitor`.

karthunni: It was added in the last commit:
https://github.com/GoogleCloudPlatform/prometheus-engine/blob/54e1f0122e63d273af54c49785c978acbe93cd83/pkg/migrate/servicemonitor.go#L281

pkg/migrate/servicemonitor.go
karthunni: This is expected as if a target namespace in `targetNamespaces` has no matching Service in the cache, the converter can't get corresponding Pod selector or target ports to use for that namespace. So we emit resources only for group.Namespaces() (namespaces where backing Service resources actually exist)

pkg/migrate/servicemonitor.go
karthunni: Oh true yeah I misunderstood, opting to use `<service[0]-namespace>-<service[0]-name>` instead

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,527 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// serviceGroup groups Services that share compatible selectors, ports, and labels.
+// This allows them to be merged into a single output PodMonitoring resource.
+type serviceGroup struct {
+	Selector     map[string]string
+	PortMap      map[string]intstr.IntOrString
```
dashpole: Quick nit: since `serviceGroup` is unexported, consider unexporting its struct fields (`selector`, `portMap`, etc.) and its `Namespaces()` method to keep internal symbols package-private.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,527 @@
...
+			}
+			portMap[portKey] = resolvedPort
+		}
+
+		// 3. Resolve target labels for this Service.
+		resolvedLabels := make(map[string]string)
+		for _, labelName := range targetLabels {
+			if val, found := svc.Labels[labelName]; found {
+				resolvedLabels[labelName] = val
+			}
```
dashpole, line 438: Design Doc Audit: Per the design document (M1 Service-level targetLabels), if a label in targetLabels does not exist on the Service, we should log a warning when skipping it:
```suggestion
		for _, labelName := range targetLabels {
			if val, found := svc.Labels[labelName]; found {
				resolvedLabels[labelName] = val
			} else {
				logger.Warn("Service-level targetLabel was not found on Service. Skipping mapping.",
					slog.String("label", labelName),
					slog.String("service", svc.Name))
			}
		}
```

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,527 @@
...
+				slog.String("label", k),
+				slog.String("renamed_target", target))
+		}
+		rule := monitoringv1.RelabelingRule{
+			TargetLabel: target,
+			Replacement: v,
+			Action:      string(relabel.Replace),
+		}
+		rules = append(rules, rule)
+		logger.Info("Service label mapped statically to metricRelabeling",
```
dashpole: Design Doc Audit: Per the design document, consider adding the note about static reflection to this info log so operators know runtime Service label changes won't automatically update metric labels:
```suggestion
		logger.Info("Service label mapped statically to metricRelabeling. Note: Changes to the Service label will not be dynamically reflected on metrics.",
			slog.String("label", fmt.Sprintf("%s: %s", k, v)))
```

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,527 @@
...
+			// nolint:staticcheck // Map deprecated BearerTokenSecret for backwards compatibility.
+			bearerTokenSecret = *ep.BearerTokenSecret
+		}
+		err = convCtx.applyAuthAndTLS(&gmpEp, ep.BasicAuth, ep.OAuth2, safeTLS, ep.Authorization, bearerTokenSecret) // nolint:staticcheck
+		if err != nil {
+			return nil, fmt.Errorf("endpoint [%d]: %w", i, err)
+		}
+
+		// Warnings for Unsupported Fields.
+		warnUnsupportedEndpointFields(convCtx.logger, ep.FollowRedirects, ep.EnableHttp2, ep.HonorLabels, ep.HonorTimestamps, ep.TrackTimestampsStaleness, i)
```
dashpole, line 391: Design Doc Audit: Per the design document compatibility matrix, advanced endpoint proxy settings like `noProxy`, `proxyConnectHeader`, and `proxyFromEnvironment` are unsupported in GMP CRDs. Consider adding warning checks for these fields when configured on the endpoint.

pkg/migrate/helpers.go
```diff
@@ -1014,18 +1025,15 @@ func convertRelabelingToMetricRelabeling(logger *slog.Logger, data *relabelingDa
...
 }
 
-// warnUnsupportedSpecFields logs warnings for spec-level fields that GMP does not support or need.
-func warnUnsupportedSpecFields(logger *slog.Logger, spec *pomonitoringv1.PodMonitorSpec) {
-	if spec == nil {
-		return
-	}
-	if spec.TargetLimit != nil {
+// warnUnsupportedMonitorSpecFields logs warnings for spec-level fields that GMP does not support or need.
+func warnUnsupportedMonitorSpecFields(logger *slog.Logger, targetLimit *uint64, keepDroppedTargets *uint64, bodySizeLimit *pomonitoringv1.ByteSize) {
```
dashpole, line 1023: Design Doc Audit: Per the design document table, consider expanding these warnings to also cover unsupported native histogram configurations (such as `nativeHistogramBucketLimit`, `nativeHistogramMinBucketFactor`, or `fallbackScrapeProtocols`) when set on the input spec.

pkg/migrate/servicemonitor.go
```diff
@@ -0,0 +1,527 @@
...
+	// Otherwise, unlabeled or partially labeled Services would incorrectly inherit static target label values from other Services in the group.
+	if !maps.Equal(g.targetLabels, targetLabels) {
+		return false
+	}
+
+	return true
+}
+
+// convertStaticTargetLabels maps Service target labels to static metricRelabeling rules.
+func convertStaticTargetLabels(logger *slog.Logger, labels map[string]string) []monitoringv1.RelabelingRule {
```
dashpole, line 510: With the introduction of `convertStaticTargetLabels` here for multi-service grouping, `convertServiceTargetLabels` in `helpers.go` is now unused dead code. We can clean up the helper in `helpers.go` and its corresponding test in `helpers_test.go`.

pkg/migrate/servicemonitor.go
karthunni: The reason I hadn't implemented this was because, our Prometheus Operator version is pinned to v0.75.0 since Prometheus is pinned in our repo. Thus, these fields get silently dropped when unmarshalling into the PodMonitor struct as they don't exist in this PO version, so we can't directly check for them through the struct.

Two options instead of silently dropping are to either:
- unconditionally log saying if these fields are set they are dropped
- check the unstructured map for these fields and log (meaning we'd have the pass the map through to `convertEndpoints`

I think the second option would be the best course of action if we feel this is necessary. 
Or I can leave TODO comments for once we are able to upgrade to the supported version of PO

pkg/migrate/helpers.go
karthunni: Same as https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2022#discussion_r3730155493

pkg/migrate/servicemonitor.go
dashpole: You can leave a TODO.

dashpole (approved):

PR #2042, opened 2026-08-04
docs(e2e): add GMP sidecar injection example script and test

bwplotka:
Adds a bash script example for injecting GMP sidecars (prometheus and config-reloader) into an existing Kubernetes Deployment via strategic merge patch, alongside an E2E test verifying the injection.

The naming has been updated to use `example-deployment` and `example-service` instead of `unreliable`.

TAG=agy
CONV=bbab9ab7-4203-4ea3-a69b-2c40577629c2

### DIFF
--- a/e2e/examples_test.go
+++ b/e2e/examples_test.go
@@ -1,0 +1,124 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"os/exec"
+	"testing"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestInjectGMPSidecarExample(t *testing.T) {
+	ctx := contextWithDeadline(t)
+
+	kubeClient, _, err := setupCluster(ctx, t)
+	if err != nil {
+		t.Fatalf("error setting up cluster: %s", err)
+	}
+
+	// 1. Create example-deployment in default namespace.
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example-deployment",
+			Namespace: "default",
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app": "example-service",
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app": "example-service",
+					},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  "example-service",
+							Image: "nginx:latest", // Just a dummy image.
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if err := kubeClient.Create(ctx, deployment); err != nil {
+		t.Fatalf("error creating example-deployment: %s", err)
+	}
+	defer func() {
+		_ = kubeClient.Delete(ctx, deployment)
+		_ = kubeClient.Delete(ctx, &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "example-deployment",
+				Namespace: deployment.Namespace,
+			},
+		})
+	}()
+
+	// 2. Run the bash script.
+	cmd := exec.CommandContext(ctx, "../examples/inject-gmp-sidecar.sh")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("error running inject-gmp-sidecar.sh: %s\n%s", err, string(out))
+	}
+	t.Logf("script output: %s", string(out))
+
+	// 3. Verify sidecars are injected and ConfigMap is created.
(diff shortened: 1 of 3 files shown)


examples/inject-gmp-sidecar.sh
```diff
@@ -0,0 +1,134 @@
...
+
+KIND="deployment"
+NAMESPACE="default"
+NAME="example-deployment"
+CONTAINER="example-service"
+METRICS_PATH="/metrics"
+PORT=80
+
+# Extract labels.
+CLUSTER_NAME=$(kubectl -n gmp-system get configmap/collector -o jsonpath='{.data.config\.yaml}' | grep 'cluster:' | awk '{print $2}' | tr -d '"')
```
bernot-dev: Any particular reason you changed this parsing from yq to grep/awk/tr? Seems more brittle and less readable to pass it through so many tools.

e2e/examples_test.go
```diff
@@ -0,0 +1,124 @@
+// Copyright 2024 Google LLC
```
bernot-dev, line 1: 2026?

Usually addlicense should handle this correctly.

bernot-dev:
1. Do we want this as part of our main e2e test suite? I supposed it's fine, but I didn't imagine the script being that "supported" because there still could be some user variation, depending on their setup.
2. Should we add more comments/documentation somewhere to explain what the script is and what it does?

examples/inject-gmp-sidecar.sh
```diff
@@ -0,0 +1,134 @@
...
+
+if [[ -z "${CLUSTER_NAME}" || -z "${LOCATION}" || -z "${PROJECT_ID}" ]]; then
+  echo "Error: Failed to extract cluster, location, or project_id from gmp-system/collector configmap." >&2
+  exit 1
+fi
+
+# Images need to be updated periodically.
+DISTROLESS_IMAGE=gke.gcr.io/gke-distroless/bash:gke_distroless_20260220.00_p0@sha256:828371616edc2c38e36868e2f8c992df37e484df72670f148de59867dfdd2490
+PROMETHEUS_IMAGE=gke.gcr.io/prometheus-engine/prometheus:v2.53.5-gmp.4-gke.0@sha256:6f349dc0be36c8a61be183254f1126c9935f5332daa96c481f7e0e1b20fe0513
+CONFIG_RELOADER_IMAGE=gke.gcr.io/prometheus-engine/config-reloader:v0.18.0-gke.2@sha256:b41862ee7ee3e9f24112ccdb0e53060085af1a8347054a7dbcff04467d3e1e9c
```
bernot-dev, line 44: Maybe we should integrate this into our regular scanning/bumping. Or reference images from the manifests? 🤔

examples/inject-gmp-sidecar.sh
bwplotka: AI decided there's no extra dependency to install, which is some benefit (:

examples/inject-gmp-sidecar.sh
bwplotka: Yea, I was thinking to do this in separate step, but we should

examples/inject-gmp-sidecar.sh
bernot-dev: Assuming awk is available and grep has consistent behavior, etc.

We already have yq as a dependency of the project.
https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/tools/go.mod#L252

examples/inject-gmp-sidecar.sh
```diff
@@ -0,0 +1,140 @@
...
+  exit 1
+fi
+
+# Extract labels.
+CLUSTER_NAME=$(kubectl -n gmp-system get configmap/collector -o jsonpath='{.data.config\.yaml}' | yq '.global.external_labels.cluster')
+LOCATION=$(kubectl -n gmp-system get configmap/collector -o jsonpath='{.data.config\.yaml}' | yq '.global.external_labels.location')
+PROJECT_ID=$(kubectl -n gmp-system get configmap/collector -o jsonpath='{.data.config\.yaml}' | yq '.global.external_labels.project_id')
+
+
+if [[ -z "${CLUSTER_NAME}" || -z "${LOCATION}" || -z "${PROJECT_ID}" ]]; then
```
dashpole, line 36: If any of these label keys are missing in the ConfigMap, yq outputs the literal string "null" rather than an empty string, which bypasses the -z check. We should check for "null" explicitly as well:
```suggestion
if [[ -z "${CLUSTER_NAME}" || "${CLUSTER_NAME}" == "null" || -z "${LOCATION}" || "${LOCATION}" == "null" || -z "${PROJECT_ID}" || "${PROJECT_ID}" == "null" ]]; then
```

e2e/examples_test.go
```diff
@@ -0,0 +1,124 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	if err := kubeClient.Create(ctx, deployment); err != nil {
+		t.Fatalf("error creating example-deployment: %s", err)
+	}
+	defer func() {
```
dashpole, line 71: If the test times out and `ctx` expires, these deferred deletions will fail with context deadline exceeded, leaving the deployment and ConfigMap behind in the test cluster. Consider creating a fresh background context with a short timeout here for cleanup operations.

e2e/examples_test.go
```diff
@@ -0,0 +1,124 @@
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("error running inject-gmp-sidecar.sh: %s\n%s", err, string(out))
+	}
+	t.Logf("script output: %s", string(out))
+
+	// 3. Verify sidecars are injected and ConfigMap is created.
+	if err := wait.PollUntilContextCancel(ctx, 2*time.Second, true, func(ctx context.Context) (bool, error) {
+		cm := &corev1.ConfigMap{}
+		if getErr := kubeClient.Get(ctx, client.ObjectKey{Name: "example-deployment", Namespace: "default"}, cm); getErr != nil {
```
dashpole, line 92: To keep this consistent with the cleanup logic above and avoid hardcoded strings, consider referencing `deployment.Name` and `deployment.Namespace` here and on line 100:
```suggestion
		if getErr := kubeClient.Get(ctx, client.ObjectKey{Name: deployment.Name, Namespace: deployment.Namespace}, cm); getErr != nil {
```

e2e/examples_test.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+			if c.Name == "config-reloader" {
+				hasReloader = true
+			}
+		}
+
+		return hasProm && hasReloader, nil
+	}); err != nil {
+		t.Fatalf("failed to verify injected sidecars or config map: %s", err)
+	}
```
dashpole, line 123: This polling loop confirms that the Deployment spec in the API server was patched with the sidecars, but it doesn't verify that the pods actually start up cleanly at runtime without crashing or failing image pulls. Consider waiting for the deployment pods to reach ready status after this check so the test catches runtime failures or config unmarshalling errors.

PR #2043, opened 2026-08-05
Security: Fix github actions security issues using zizmor

bwplotka:
Enabled zizmor security checker. Auto-fixed cache poisoning in setup-go by adding `cache: false`. Pinned all github actions to a specific hash.


.github/workflows/presubmit.yml
```diff
@@ -34,25 +34,26 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
         with:
           persist-credentials: false
-      - uses: actions/setup-go@v6
+      - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6
         with:
           go-version: 'stable'
+          cache: false
```
bwplotka, line 54: It's quite insane we are asked to ignore the most useful thing in GH action to speed up things because of some apparent security risk..

bwplotka:
I am going ahead and merge as it's trivial and unblocks sidecar example: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2042

PR #2060, opened 2026-08-06
feat: support Service discovery for Alertmanager

### LINKED ISSUE #595: Support non-Endpoint-backed Services for AlertmanagerEndpoints
The rule-evaluator is only able to [route alerts](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.7.4/doc/api.md#alertmanagerendpoints) to Alertmanagers backed by Kubernetes Endpoints.

However, valid cases exist to route to Alertmanagers outside of the current cluster.

See the discussion in https://github.com/GoogleCloudPlatform/prometheus-engine/discussions/576#discussioncomment-7128328 for context.

hugosmoreira:
## What this change does

- Adds an opt-in `Service` discovery mode to `AlertmanagerEndpoints` while preserving existing `Endpoints` discovery as the default.
- Generates a static Alertmanager target using the Kubernetes Service DNS name (`<name>.<namespace>.svc:<port>`), so portless `ExternalName` Services work without an Endpoints object.
- Requires a numeric port for Service discovery and retains the existing authorization, TLS, scheme, path prefix, API version, timeout, redirect, and HTTP/2 configuration.
- Regenerates the OperatorConfig CRDs, combined setup manifest, and API documentation.
- Adds validation, configuration-generation, and rule-evaluator golden coverage for the new mode.

## Why

The current Kubernetes discovery path requires an Endpoints object. Services without Endpoints, including the `ExternalName` setup described in #595, therefore cannot be used as Alertmanager targets without manually maintaining an Endpoints object.

Fixes #595.

## Validation

- `go test ./pkg/operator/... -count=1` in Go 1.25
- Repository-wide `./hack/presubmit.sh test` in the repository's hermetic Go 1.26.5 image
- Uncached `./hack/presubmit.sh all diff` in the repository's hermetic image
- Generated CRD, manifest, API documentation, and rule-evaluator golden output verified by the presubmit

## Assistance disclosure

This contribution was prepared with Codex assistance and the resulting code, generated artifacts, and tests were reviewed and validated locally before publication.

### DIFF
--- a/pkg/operator/apis/monitoring/v1/operator_types.go
+++ b/pkg/operator/apis/monitoring/v1/operator_types.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -88,6 +89,16 @@ func validateRules(rules *RuleEvaluatorSpec) error {
 }
 
 func validateAlertManagerEndpoint(alertManagerEndpoint *AlertmanagerEndpoints) error {
+	switch alertManagerEndpoint.DiscoveryType {
+	case "", AlertmanagerDiscoveryTypeEndpoints:
+	case AlertmanagerDiscoveryTypeService:
+		port, err := strconv.ParseUint(alertManagerEndpoint.Port.String(), 10, 16)
+		if err != nil || port == 0 {
+			return fmt.Errorf("service discovery requires a numeric port between 1 and 65535, got %q", alertManagerEndpoint.Port.String())
+		}
+	default:
+		return fmt.Errorf("unknown discovery type %q", alertManagerEndpoint.DiscoveryType)
+	}
 	if alertManagerEndpoint.Authorization != nil {
 		if err := validateSecretKeySelector(alertManagerEndpoint.Authorization.Credentials); err != nil {
 			return fmt.Errorf("invalid authorization credentials: %w", err)
@@ -301,15 +312,30 @@ type ManagedAlertmanagerSpec struct {
 	ExternalURL string `json:"externalURL,omitempty"`
 }
 
-// AlertmanagerEndpoints defines a selection of a single Endpoints object
-// containing alertmanager IPs to fire alerts against.
+// AlertmanagerDiscoveryType defines how Alertmanager targets are discovered.
+type AlertmanagerDiscoveryType string
+
+const (
+	// AlertmanagerDiscoveryTypeEndpoints discovers individual Alertmanager replicas from Endpoints.
+	AlertmanagerDiscoveryTypeEndpoints AlertmanagerDiscoveryType = "Endpoints"
+	// AlertmanagerDiscoveryTypeService routes alerts through a Kubernetes Service DNS name.
+	AlertmanagerDiscoveryTypeService AlertmanagerDiscoveryType = "Service"
+)
+
+// AlertmanagerEndpoints defines a selection of Alertmanager targets to fire alerts against.
 type AlertmanagerEndpoints struct {
-	// Namespace of Endpoints object.
+	// Namespace of the Endpoints object or Service.
 	Namespace string `json:"namespace"`
-	// Name of Endpoints object in Namespace.
+	// Name of the Endpoints object or Service in Namespace.
 	Name string `json:"name"`
 	// Port the Alertmanager API is exposed on.
 	Port intstr.IntOrString `json:"port"`
+	// DiscoveryType controls how Alertmanager targets are discovered. Endpoints, the default,
+	// discovers each Alertmanager replica from a Kubernetes Endpoints object. Service routes
+	// through the Kubernetes Service DNS name and requires a numeric Port. Service discovery
+	// supports Services without Endpoints, such as ExternalName Services.
+	// +kubebuilder:validation:Enum=Endpoints;Service
+	DiscoveryType AlertmanagerDiscoveryType `json:"discoveryType,omitempty"`
 	// Scheme to use when firing alerts.
 	Scheme string `json:"scheme,omitempty"`
 	// Prefix for the HTTP path alerts are pushed to.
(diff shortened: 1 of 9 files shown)


bwplotka:
Thanks! I think this change is reasonable. Do you self-deploy operator or do you use managed GMP?

hugosmoreira:
Thanks! I don't currently run either setup in production. I picked this up from #595 and validated the change against the open-source operator's configuration-generation path and repository tests, but I haven't tested it with managed GMP. The use case I had in mind is self-deployed operators that need to target an ExternalName Service without maintaining an Endpoints object. If managed GMP has additional API or compatibility constraints, I'm happy to adjust the scope.

bwplotka:
Cool, that's ok - let me know once it's ready for review!

hugosmoreira:
Thanks! It’s ready for review now. The implementation and relevant automated repository tests are complete and passing.

pkg/operator/operator_config_test.go
```diff
@@ -20,18 +20,46 @@ import (
...
+	require.Len(t, configs, 1)
+	require.Len(t, configs[0].ServiceDiscoveryConfigs, 1)
+
+	staticConfig, ok := configs[0].ServiceDiscoveryConfigs[0].(discovery.StaticConfig)
+	require.True(t, ok)
+	require.Len(t, staticConfig, 1)
+	require.Len(t, staticConfig[0].Targets, 1)
+	require.Equal(t,
+		prommodel.LabelValue("external-alertmanager.monitoring.svc:9093"),
+		staticConfig[0].Targets[0][prommodel.AddressLabel],
```
bwplotka, line 59: It looks like we added a test for `Service` discovery here, but it might be worth adding a regression test to ensure that the default `Endpoints` SD behaviour remains backwards compatible too.

pkg/operator/apis/monitoring/v1/operator_types_test.go
```diff
@@ -37,6 +38,40 @@ func TestOperatorConfigValidate(t *testing.T) {
 				},
 			},
 		},
+		{
```
bwplotka, line 41: Consider adding a test case to verify that an unknown `DiscoveryType` safely falls back and returns the expected unknown type validation error.

pkg/operator/apis/monitoring/v1/operator_types.go
```diff
@@ -88,6 +89,16 @@ func validateRules(rules *RuleEvaluatorSpec) error {
 }
 
 func validateAlertManagerEndpoint(alertManagerEndpoint *AlertmanagerEndpoints) error {
+	switch alertManagerEndpoint.DiscoveryType {
+	case "", AlertmanagerDiscoveryTypeEndpoints:
+	case AlertmanagerDiscoveryTypeService:
```
bwplotka, line 94: Since `Service` discovery bypasses standard endpoints validation, an empty `Name` or `Namespace` could yield a broken `svcDNSName` (e.g. `..svc:9093`). Consider adding an explicit sanity rejection for empty strings here.

pkg/operator/operator_config.go
```diff
@@ -651,6 +651,19 @@ func (r *operatorConfigReconciler) makeAlertmanagerConfigs(ctx context.Context,
 			cfg.HTTPClientConfig.TLSConfig = tlsCfg
 		}
 
+		if am.DiscoveryType == monitoringv1.AlertmanagerDiscoveryTypeService {
+			svcDNSName := fmt.Sprintf("%s.%s.svc:%s", am.Name, am.Namespace, am.Port.String())
+			cfg.ServiceDiscoveryConfigs = discovery.Configs{
+				discovery.StaticConfig{
+					&targetgroup.Group{
+						Targets: []prommodel.LabelSet{{prommodel.AddressLabel: prommodel.LabelValue(svcDNSName)}},
```
bwplotka, line 659: Technically this is a DNS SD not a "service" discovery through Kubernetes API.  Why not using Kube API for this (other than more code?)

bwplotka (approved):
One top level question first.
 
Just to understand the expectations again, is this something you need or just helping us with a tech debt?

bwplotka (changes requested):
Sorry I didn't meant to approve yet, just comment

pkg/operator/operator_config.go
bwplotka: Also if we need DNS we could use DNS SD too

hugosmoreira:
Thanks for checking. This is a community contribution addressing #595 rather than something I personally require in production. The issue describes a real user need for routing managed rule evaluation to an external Alertmanager through an ExternalName Service.

You are right that the current implementation constructs a static DNS target rather than using Kubernetes Service discovery. Since the rule evaluator already has Kubernetes API access and Service permissions, `kubernetes_sd_config` with `role: service`, filtered by namespace, Service name, and port, seems like the better fit. I can revise the implementation that way and add the requested compatibility, unknown-type, and empty-field tests. Would you prefer that over `dns_sd_config`?

hugosmoreira:
Following up before revising the discovery approach. I found an important edge case in the original #595 example: its ExternalName Service has no `spec.ports`; port 443 is specified only in OperatorConfig.

In the [pinned Prometheus fork](https://github.com/GoogleCloudPlatform/prometheus/blob/10ff5b96bd82/discovery/kubernetes/service.go), `role: service` creates targets by iterating `spec.ports`, so switching to it alone would discover no targets for that example.

Would requiring a declared Service port be acceptable, or should we preserve portless ExternalName support? If preserving it is preferred, would you favor `dns_sd_config`? I'd verify the HTTPS/hostname behavior before proposing that change.

I'll keep default Endpoints compatibility and the requested unknown-type and empty-name/namespace regression cases in scope.

hugosmoreira:
Gentle follow-up on the discovery contract for #595: should the portless ExternalName Service in the issue example remain supported, or is requiring a declared Service port acceptable? In the pinned Prometheus fork, `role: service` produces no target for that example because the Service has no `spec.ports`. I am keeping this PR in draft until we agree on the approach, so your guidance would help me avoid unnecessary churn. Thanks!

PR #2061, opened 2026-08-07
PO to GMP Migration Tool:  Add Draft Generation with TODO Annotations for Errors

karthunni:
This PR enhances the Prometheus Operator to GMP migration tool by implementing graceful draft generation for conversion errors instead of aborting the entire migration. The converter produces manifests with TODO annotations, placeholder values, and a safety guardrail label.

* **Missing Secret / ConfigMap Keys & References:**
  * Emits `TODO_MISSING_KEY_<KEY>_IN_<KIND>_<NAME>` and `TODO_SET_SECRET_NAME` placeholders with `[ERROR]` TODO annotations instead of returning hard errors.
* **Duration Parsing & Scrape Timeouts:**
  * Defaults invalid scrape intervals to `"30s"` with `[ERROR]` TODOs to preserve schema validity.
  * Drops invalid scrape timeouts and caps timeouts exceeding scrape intervals with actionable TODOs.
* **Endpoint & Service Port Resolution:**
  * Emits `port: "TODO_SET_PORT"` when an endpoint omits both `port` and `targetPort`.
  * Resolves unmapped Service ports to `port: "TODO_RESOLVE_PORT_<PORT>"` with `[WARNING]` TODOs.
* **Missing Backing Services for `ServiceMonitor`:**
  * When a `ServiceMonitor` has no matching backing `Service` in the inputs, generates a draft `PodMonitoring` with `selector: {app: "TODO_SET_POD_SELECTOR"}` and `port: "TODO_RESOLVE_PORT"`.
* **Selector & Keep Rule Conflicts:**
  * Preserves base selectors and records actionable TODOs when relabel keep rules conflict with existing match labels.
* **Safety Guardrail Label:**
  * Injects `gmp.googleapis.com/migration-review-required: "true"` into `spec.selector.matchLabels` whenever `[ERROR]` TODO items are present, preventing unintentional scraping until user review.


pkg/migrate/helpers.go
```diff
@@ -1267,41 +1412,60 @@ func resolveFilterRunning(filterRunnings []*bool, logger *slog.Logger, isCluster
...
+				category: "ERROR",
+				reason:   "Proxy URL contains embedded plaintext credentials. Credentials were removed.",
+				action:   "Configure proxy authentication via Kubernetes Secret or proxy server configuration.",
+			})
+			parsed.User = nil
+			return parsed.String(), nil
+		}
 	}
 	return *proxyURL, nil
 }
```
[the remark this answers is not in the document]
karthunni: This is the whole intent of the PR

bwplotka (commented):
Looks like a reasonable approach if you can make it a pattern for all errors/warnings (e.g. we don't forget about something).

Not sure about blocking scrape logic - it feels odd, but it is one of the least bad options if we assume people or agents will apply without looking

pkg/migrate/servicemonitor.go
```diff
@@ -208,16 +212,44 @@ func (c *ServiceMonitorConverter) findAndGroupServices(
...
+		portMap := make(map[string]intstr.IntOrString)
+		for _, ep := range sm.Spec.Endpoints {
+			k := endpointPortKey(ep)
+			if k == "" {
+				k = "TODO_SET_PORT"
+			}
+			portMap[k] = intstr.FromString("TODO_RESOLVE_PORT")
+		}
+		dummySvc := &corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{
```
dashpole: When backing Services are missing, using `sm.Namespace` for `dummySvc` ignores `targetNamespaces` (e.g. from `namespaceSelector.matchNames`), causing draft `PodMonitoring` manifests to be generated in `sm.Namespace` instead of the targeted namespaces. Consider creating dummy Services for each namespace in `targetNamespaces`.

pkg/migrate/helpers.go
```diff
@@ -831,34 +903,38 @@ func convertTargetLabels(logger *slog.Logger, sourceLabels []string, jobLabel st
...
+				return true, true, fmt.Sprintf("'%s' on '%s'", action, s)
+			}
+			return true, false, ""
 		}
-		if strings.HasPrefix(string(sl), "__meta_kubernetes_node_") && string(sl) != "__meta_kubernetes_node_name" {
-			logger.Warn(fmt.Sprintf("Relabeling rule referencing node metadata %q is unsupported in GMP (only node name is supported). The rule has been dropped.", string(sl)))
-			return true
+		if strings.HasPrefix(s, "__meta_kubernetes_node_") && s != "__meta_kubernetes_node_name" {
+			logger.Warn(fmt.Sprintf("Relabeling rule referencing node metadata %q is unsupported in GMP (only node name is supported). The rule has been dropped.", s))
+			return true, false, ""
```
dashpole, line 919: When dropping a `keep` or `drop` rule on node metadata, `scopeExpanded` is returned as `false`. Consider returning `scopeExpanded: true` if `action == relabel.Keep || action == relabel.Drop` (like lines 926-929 for pod annotations) so a TODO and guardrail label are attached.

pkg/migrate/migrate.go
```diff
@@ -141,11 +144,18 @@ func (m *Migrator) Run(inputPaths ...string) (*MigrationReport, error) {
...
+	fmt.Fprintf(m.Stderr, "  Migrated with Warnings:     %d\n", r.WarningsCount)
+	fmt.Fprintf(m.Stderr, "  Migrated with Action Items: %d\n", r.ActionItemsCount)
+	fmt.Fprintf(m.Stderr, "  Skipped (Unsupported):      %d\n", r.SkippedCount)
+	fmt.Fprintf(m.Stderr, "  Failed:                     %d\n", r.FailedCount)
 	fmt.Fprintln(m.Stderr, "=========================================")
+	if r.ActionItemsCount > 0 {
+		fmt.Fprintln(m.Stderr, "\nNOTE: Some resources were migrated with action items and contain TODO annotations.")
+		fmt.Fprintln(m.Stderr, "These resources include the safety guardrail label:")
+		fmt.Fprintln(m.Stderr, "  'gmp.googleapis.com/migration-review-required: \"true\"'")
+		fmt.Fprintln(m.Stderr, "Review the TODO annotations in the generated manifests and remove this label when ready to activate scraping.")
```
dashpole: Consider using the `GuardrailLabelKey` and `GuardrailLabelValue` constants here instead of hardcoded strings:
```suggestion
		fmt.Fprintf(m.Stderr, "  '%s: \"%s\"'\n", GuardrailLabelKey, GuardrailLabelValue)
```

pkg/migrate/servicemonitor_test.go
```diff
@@ -698,7 +724,8 @@ func TestServiceMonitorConverter_Convert(t *testing.T) {
 					},
 				},
 			},
-			wantErr: true,
+			expected: nil,
```
dashpole, line 838: Note that `TestServiceMonitorConverter_Convert` currently guards its output length assertion with `if tc.expected != nil`, which causes tests with `expected: nil` to skip verifying that `outputs` is actually empty.

pkg/migrate/helpers.go
```diff
@@ -1267,43 +1372,70 @@ func resolveFilterRunning(filterRunnings []*bool, logger *slog.Logger, isCluster
...
+	if parsed.User != nil {
+		if _, hasPass := parsed.User.Password(); hasPass {
+			c.todos = append(c.todos, todoItem{
+				category: "ERROR",
+				reason:   "Proxy URL contains embedded plaintext credentials. Credentials were removed.",
+				action:   "Configure proxy authentication via Kubernetes Secret or proxy server configuration.",
+			})
+			parsed.User = nil
+			return parsed.String()
+		}
```
dashpole: If the proxy URL contains a username without a password (such as `http://user@host`), `hasPass` is false so credentials aren't stripped and the resulting URL still contains `@` which is rejected by GMP. Consider checking `if parsed.User != nil` directly without requiring a password:
```suggestion
		c.todos = append(c.todos, todoItem{
			category: "ERROR",
			reason:   "Proxy URL contains embedded plaintext credentials. Credentials were removed.",
			action:   "Configure proxy authentication via Kubernetes Secret or proxy server configuration.",
		})
		parsed.User = nil
		return parsed.String()
```

karthunni:
> Looks like a reasonable approach if you can make it a pattern for all errors/warnings (e.g. we don't forget about something).
> 
> Not sure about blocking scrape logic - it feels odd, but it is one of the least bad options if we assume people or agents will apply without looking

Reconsidered potential options and decided to take the following approach:

On a default CLI call (i.e no flags), we output the manifests without TODO items, while noting in the stderr output summary: (such that users can safely blindly apply)
```
X manifests with action items were omitted from Stdout as they contain best-effort draft configurations with TODO annotations and placeholders.
Run with '--all' to output all manifests for review.
```

When the `--all` flag is added, all manifests are written to stdout including the ones with TODOs (still in the annotations but no longer blocking scraping logic with the matchLabels), and the output summary states: (such that an explicit call is needed to output "dangerous" CRs)
```
X manifests contain best-effort draft configurations with TODO annotations and placeholders.
Review the inline 'gmp.googleapis.com/todo-*' annotations in the generated manifests before applying to a cluster.
```

pkg/migrate/podmonitor.go
```diff
@@ -182,15 +177,25 @@ func (c *PodMonitorConverter) convertMonitorSpec(pm *pomonitoringv1.PodMonitor,
 	}
 
 	mergedFromPod := mergeFromPod(logger, convertTargetLabels(logger, pm.Spec.PodTargetLabels, pm.Spec.JobLabel, "Pod"), rules.ResourceCombined.FromPod)
-	mergedSelector, err := mergeLabelSelector(pm.Spec.Selector, rules.ResourceCombined.MatchLabels, rules.ResourceCombined.MatchExpressions)
-	if err != nil {
-		return nil, err
-	}
+	mergedSelector := convCtx.mergeLabelSelector(pm.Spec.Selector, rules.ResourceCombined.MatchLabels, rules.ResourceCombined.MatchExpressions)
```
dashpole, line 181: `convCtx` is initialized in `convertMonitorSpec` without setting `isClusterScoped: isCluster`. For cluster-scoped conversions (`isCluster == true`), referenced secrets in `convertSecretSelector` will omit their namespace reference.

pkg/migrate/servicemonitor.go
```diff
@@ -409,21 +446,33 @@ func groupServices(
 		// 1. Extract and validate selector.
 		selectorString := svc.Spec.Selector
 		if len(selectorString) == 0 {
-			return nil, fmt.Errorf("service %q has no selector (targets external or static endpoints). GMP Managed Collection only supports scraping in-cluster Pod targets", svc.GetName())
+			logger.Info("Service targets external endpoints without a pod selector. GMP Managed Collection only supports in-cluster Pods. Skipping resource.",
+				slog.String("migration_status", "skipped"),
```
dashpole: Logging `migration_status: "skipped"` here sets the parent ServiceMonitor's tracked status to `StatusSkipped`. If the ServiceMonitor has other valid services that convert successfully, `StatusSuccess` won't override `StatusSkipped` in `trackStatus`, so the resource gets counted in `SkippedCount` instead of `SuccessCount`. Consider omitting `migration_status` for skipped child services.

pkg/migrate/podmonitor_test.go
```diff
@@ -1641,7 +1683,33 @@ func TestPodMonitorConversion(t *testing.T) {
...
+								Port:     intstr.FromString("metrics"),
+								Interval: "30s",
+							},
+						},
+					},
+				},
+			},
 		},
 		{
 			name: "BearerTokenSecret with empty Name returns validation error",
```
dashpole: The test case name still mentions returning a validation error, but it now expects a draft with TODO annotations and guardrail labels:
```suggestion
			name: "BearerTokenSecret with empty Name generates draft with TODO annotation and guardrail label",
```

dashpole (approved):

PR #2068, opened 2026-08-12
PO to GMP Migration Tool: Add gmp-migrate agent skill

karthunni:
This PR introduces the `gmp-migrate` agent skill (`cmd/gmp-migrate/SKILL.md`) to guide AI coding agents through running the migration CLI, diagnosing migration reports, and safely reconciling draft GMP manifests with `gmp.googleapis.com/todo-*` annotations and `TODO_*` placeholders.
### Key Highlights
* **CLI Stream Separation & Modes**: Documents `--all` vs default mode, `Stdout`/`Stderr` stream separation, and exit code semantics.
* **Strict Read-Only Investigation**: Enforces a strict policy where all cluster investigation commands (`kubectl get`, `describe`, `logs`) are strictly read-only, prohibiting unapproved live cluster mutations.
* **Action Plan Catalog & Interactive Recipes**:
  * **Port & Selector Resolution**: Guided resolution for missing backing Services, unresolved named ports, and ambiguous multi-container pods.
  * **Auth & Secret Management**: Ingests companion Secrets, handles base64 payload errors, and prompts for missing OAuth2/Proxy endpoints.
  * **Relabeling & Scope Expansion**: Interactive reconciliation for dropped `keep`/`drop` annotation rules (`prometheus.io/scrape`), including `spec.selector` intersection queries, companion workload label patches, and `matchExpressions: NotIn` inverted selectors.
  * **Rolling Restart Safeguards**: Explicitly requires warning users and requesting permission before modifying workload pod templates (`spec.template.metadata.labels`).
* **Audit Reporting & Per-File Diffs**: Mandates presenting a structured Reconciliation Audit Table and file-by-file code diffs prior to executing dry-runs or `kubectl apply`.

### DIFF
--- a/cmd/gmp-migrate/SKILL.md
+++ b/cmd/gmp-migrate/SKILL.md
@@ -126,23 +126,29 @@ All TODO annotations follow the format: `gmp.googleapis.com/todo-N: "[WARNING|ER
        interval: 30s
   ```
 
-#### 1.2 Missing Backing Service (`TODO_RESOLVE_PORT` & `TODO_SET_POD_SELECTOR`)
+#### 1.2 Missing Backing Service (`TODO_RESOLVE_PORT` & `TODO_SET_POD_LABELS`)
+* **Context**: `gmp-migrate` emits `TODO_SET_POD_LABELS: TODO_SET_POD_LABELS` as a safe guardrail placeholder. Replace this placeholder with the complete set of pod template labels for the target workload.
 * **Annotation**: `[ERROR] Corresponding Kubernetes Service was not found... ACTION: Define target pod selector in 'spec.selector.matchLabels' and verify endpoint ports.`
 * **Commands**:
   ```bash
   kubectl get svc -n <namespace> -o wide
+  kubectl get deployment,statefulset,daemonset -n <namespace> -o wide
   ```
 * **Agent Interactive Step**:
-  - If a matching Service is identified, copy its `spec.selector` into `spec.selector.matchLabels` and resolve the endpoint port.
-  - If no Service exists or multiple ambiguous Services match, prompt the user:
-    > *"Could not find a unique backing Service for `<name>`. Please provide the target Pod label selector (e.g. `app=payment`) and container port."*
+  - If a matching Service or workload is identified, copy its pod label selector into `spec.selector.matchLabels` and resolve endpoint ports.
+  - If multiple ambiguous workloads match, prompt the user:
+    > *"Could not find a backing Service for `<name>`. Discovered candidate workloads: `[app=payment-processor, app=payment-worker]`. Please confirm the target pod labels."*
 * **Diff**:
   ```diff
+   metadata:
+  -  annotations:
+  -    gmp.googleapis.com/todo-1: "[ERROR] Corresponding Kubernetes Service was not found..."
    spec:
      selector:
        matchLabels:
-  -      app: TODO_SET_POD_SELECTOR
+  -      TODO_SET_POD_LABELS: TODO_SET_POD_LABELS
   +      app.kubernetes.io/name: payment-processor
+  +      app.kubernetes.io/component: backend
      endpoints:
   -  - port: TODO_RESOLVE_PORT
   +  - port: 9102
@@ -210,10 +216,13 @@ All TODO annotations follow the format: `gmp.googleapis.com/todo-N: "[WARNING|ER
 #### 2.3 Corrupt Base64 Data (`TODO_CORRUPT_SECRET_DATA_<KEY>`)
 * **Annotation**: `[ERROR] Failed to base64-decode key "<key>" in Secret "<name>"...`
 * **Agent Interactive Step**:
-  Do NOT fabricate passwords or placeholders. Prompt the user:
+  Do NOT fabricate key values or placeholders. Prompt the user:
   > *"Key `<key>` in Secret `<name>` contains invalid base64 data. Please provide the correct value or fix the Secret in the cluster."*
 * **Diff**:
   ```diff
+   metadata:
+  -  annotations:
+  -    gmp.googleapis.com/todo-1: "[ERROR] Failed to base64-decode key \"user\" in Secret \"app-auth\"..."
    spec:
      endpoints:
      - port: 8080
@@ -293,16 +302,36 @@ All TODO annotations follow the format: `gmp.googleapis.com/todo-N: "[WARNING|ER
              key: client-secret
   ```
 
-#### 4.2 Proxy URL Configuration (`TODO_SET_VALID_PROXY_URL` & Plaintext Passwords)
-* **Annotation**: `[ERROR] Proxy URL contains embedded plaintext credentials...`
-* **Resolution**: Prompt user for proxy host. Note that GMP endpoints do not support embedded basic auth in proxy URLs.
-* **Diff**:
+#### 4.2 Proxy URL Configuration (Malformed URLs & Stripped Credentials)
+* **Context**: GMP CRDs only support an unauthenticated `proxyUrl` string. GMP does not support proxy credentials (there are no Secret fields for proxies, and embedded credentials like `user:pass@` in `proxyUrl` are rejected by CRD admission rules).
+* **Annotations**:
+  * `[ERROR] Proxy URL "<url>" is invalid or malformed. ACTION: Specify a valid proxy URL (e.g. 'http://proxy.example.com:8080').`
+  * `[ERROR] Proxy URL contains embedded plaintext credentials. Credentials were removed. ACTION: Configure proxy authentication via proxy server configuration or network allowlist.`
+* **Agent Interactive Step**:
+  1. **Malformed URL (`TODO_SET_VALID_PROXY_URL`)**: Prompt the user for the valid corporate proxy host and port.
+  2. **Stripped Credentials**: `gmp-migrate` automatically strips `user:password@` so the manifest passes GMP validation while preserving the proxy host and port. Prompt the user:
+     > *"Proxy URL credentials were removed because GMP CRDs do not support proxy authentication. Please ensure proxy authentication is handled at the network level (e.g. IP allowlisting the GKE cluster at the proxy server or routing through an in-cluster forward proxy). If the target is inside the cluster/VPC, `proxyUrl` can be removed entirely."*
+  3. Once confirmed, remove the TODO annotation from `metadata.annotations`.
+* **Diff (Case A: Resolving Malformed URL Placeholder)**:
   ```diff
+   metadata:
+  -  annotations:
+  -    gmp.googleapis.com/todo-1: "[ERROR] Proxy URL is invalid or malformed..."
    spec:
      endpoints:
      - port: 8080
   -    proxyUrl: TODO_SET_VALID_PROXY_URL
   +    proxyUrl: http://proxy.corp.internal:3128
+  ```
+* **Diff (Case B: Reconciling Stripped Credentials Annotation)**:
+  ```diff
+   metadata:
+  -  annotations:
+  -    gmp.googleapis.com/todo-1: "[ERROR] Proxy URL contains embedded plaintext credentials. Credentials were removed..."
+   spec:
+     endpoints:
+     - port: 8080
+       proxyUrl: http://proxy.corp.internal:3128
   ```
 
 ---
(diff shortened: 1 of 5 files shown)


cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,521 @@
...
+> Under NO circumstances should an agent run state-changing commands (`kubectl apply`, `kubectl create`, `kubectl patch`, `kubectl delete`) against a live cluster without reaching Section 7 and receiving explicit user approval.
+
+All TODO annotations follow the format: `gmp.googleapis.com/todo-N: "[WARNING|ERROR] <reason> ACTION: <action>"`. Match the placeholder or annotation message against the recipes below:
+
+---
+
+### Recipe 1: Port & Pod Selector Resolution
+
+#### 1.1 Unresolved Named Port (`TODO_RESOLVE_PORT_<NAME>`)
+* **Annotation**: `[WARNING] Named port "<name>" was not found on Service "<svc>"... ACTION: Replace TODO_RESOLVE_PORT_<NAME> with the target container port name or number.`
```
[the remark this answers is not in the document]
karthunni: policy has changed

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,521 @@
...
+* **Agent Interactive Step**:
+  Check `Stderr` logs for conflicting values and ask the user to clarify:
+  > *"Detected conflicting label rules for label `env` (`prod` vs `staging`). Which workload should this PodMonitoring target?"*
+* **Diff**:
+  ```diff
+   spec:
+     selector:
+       matchLabels:
+  -      env: production # Conflicted with staging
+  +      env: production
```
dashpole: The diff replaces `env: production` with `env: production` and includes an inline comment that `gmp-migrate` does not emit in the YAML. If the user clarifies they want `staging`, maybe show updating the value to `staging` (or removing the TODO annotation from metadata if keeping `production`)?
```suggestion
  -      env: production
  +      env: staging
```

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,521 @@
...
+* **Agent Interactive Step**:
+  - If a matching Service is identified, copy its `spec.selector` into `spec.selector.matchLabels` and resolve the endpoint port.
+  - If no Service exists or multiple ambiguous Services match, prompt the user:
+    > *"Could not find a unique backing Service for `<name>`. Please provide the target Pod label selector (e.g. `app=payment`) and container port."*
+* **Diff**:
+  ```diff
+   spec:
+     selector:
+       matchLabels:
+  -      app: TODO_SET_POD_SELECTOR
```
dashpole: Interesting that the label key also changes?  Should it also be "TODO_SET_POD_SELECTOR" or should we omit the key (and expect that both the key and value are provided)? The current output implies that `app` should be preserved, and only the value should be set.

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,521 @@
...
+       basicAuth:
+  -      username: TODO_MISSING_KEY_USERNAME_IN_SECRET_APP-AUTH
+  +      username: admin
+  ```
+
+#### 2.3 Corrupt Base64 Data (`TODO_CORRUPT_SECRET_DATA_<KEY>`)
+* **Annotation**: `[ERROR] Failed to base64-decode key "<key>" in Secret "<name>"...`
+* **Agent Interactive Step**:
+  Do NOT fabricate passwords or placeholders. Prompt the user:
+  > *"Key `<key>` in Secret `<name>` contains invalid base64 data. Please provide the correct value or fix the Secret in the cluster."*
```
dashpole, line 257: This is only usernames, right?  We aren't having agents put passwords in PodMonitoring objects?

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,521 @@
...
+  +      tokenUrl: "https://oauth2.googleapis.com/token"
+         clientSecret:
+           secret:
+             name: oauth-secret
+             key: client-secret
+  ```
+
+#### 4.2 Proxy URL Configuration (`TODO_SET_VALID_PROXY_URL` & Plaintext Passwords)
+* **Annotation**: `[ERROR] Proxy URL contains embedded plaintext credentials...`
+* **Resolution**: Prompt user for proxy host. Note that GMP endpoints do not support embedded basic auth in proxy URLs.
```
dashpole: What should be done if the user was using embedded basic auth?  Put it in a secret?

dashpole (commented):
It would be useful to document (or have gemini document) the steps you've taken to test this. Hosting evals in this repo might be a bit heavy-weight, but some evidence that this does what we want it to ~most of the time would be good.

cmd/gmp-migrate/SKILL.md
karthunni: I think changing the label name to be TODO_SET_POD_SELECTOR makes sense to keep it inline with what the selector should look like. 

Also being more explicit in the skill that we're replacing with a set of labels (the full selector) and not just a one label change

cmd/gmp-migrate/SKILL.md
karthunni: Yeah they're all just SecretSelector references and passwords are never handled/put into PodMonitoring objects. This extraction is only done for non-sensitive string fields that PO had as secret references (specifically `BasicAuth.Username` and `OAuth2.ClientID`).

In the skill it'd be better to say:
"Do NOT fabricate key values or placeholders"

cmd/gmp-migrate/SKILL.md
karthunni: So actually it seems like nothing can be done with just a Secret (the action in the log should be changed), it would require some changes at the proxy configuration level (allowlisting the Cluster or adding a forward proxy)

Adding this added context in the skill

karthunni:
> It would be useful to document (or have gemini document) the steps you've taken to test this. Hosting evals in this repo might be a bit heavy-weight, but some evidence that this does what we want it to ~most of the time would be good.

Created this draft PR with all our placeholder cases against an example Kind Cluster. Had subagents run the tool 5 times per test case using the tool and had an agent audit the outputted trajectories (and tried to manually review them myself). From a first glance it looks pretty consistent
https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2070

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,563 @@
...
+---
+
+### Recipe 2: Missing or Incomplete Secrets & ConfigMaps
+
+#### 2.1 Missing Source Secret / ConfigMap (`TODO_SET_<KEY>_FROM_<KIND>_<NAME>`)
+* **Annotation**: `[ERROR] Referenced SECRET "<name>" for key "<key>" was not found in migration inputs...`
+* **Commands**:
+  ```bash
+  # Fetch Secret data from cluster
+  kubectl get secret <secret-name> -n <namespace> -o jsonpath='{.data.<key>}' | base64 --decode
```
dashpole: I think we should require a human in the loop and explicitly prohibit the agent from reading Secrets via `kubectl get secret`. To prevent credentials and sensitive data from accidentally leaking into agent prompt contexts, logs, or local disk, the agent should instead explain this and prompt the user to provide the non-sensitive string (e.g. username or clientID) or update the manifest directly.

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,563 @@
...
+         selector:
+           matchLabels:
+             app.kubernetes.io/name: payment-gateway
+           matchExpressions:
+           - key: prometheus.io/scrape
+             operator: NotIn
+             values:
+             - "false"
+       ```
+     > **Note**: Kubernetes `NotIn` matches all pods where the label is absent, or present with any value other than `"false"`.
```
dashpole: It might be worth explicitly highlighting here that because `NotIn` matches any pod where the label is absent, applying an inverted selector could match newly created or unannotated pods in the namespace if they share the rest of `matchLabels`. Adding a brief warning to ensure operators verify that unannotated pods won't be unintentionally scraped would help avoid unexpected scope expansion.

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,563 @@
...
+
+## 7. Verification & User Agency
+
+Before applying changes to a live cluster:
+1. **Present Reconciliation Steps & Per-File Code Diffs**: Output the structured Reconciliation Audit Table detailing the steps taken for each TODO, followed by complete per-file diffs showing the exact before-and-after modifications for all reconciled manifests and companion files.
+2. **Server-Side Dry Run**:
+   ```bash
+   kubectl apply --dry-run=server -f <output-manifests.yaml>
+   ```
+3. **Explicit User Approval**: Ask the user before running `kubectl apply`.
```
dashpole: Could we add a note here for GitOps-managed clusters (e.g. ArgoCD, Flux, Config Sync)? In those environments, running `kubectl apply` directly against the cluster will be reverted or cause drift, so the agent should instead output the reconciled manifests and companion patches for the user to commit to their Git repo.

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,645 @@
...
+* **Do Not Read Secrets via CLI**: Agents must never run `kubectl get secret` to inspect secret payloads. Reading cluster secrets risks leaking sensitive credentials (passwords, tokens, private keys) into agent context logs and transcripts.
+* **Dual-Path User Handoff**: When a Secret placeholder (`TODO_SET_USERNAME_FROM_SECRET_*` or `TODO_SET_CLIENT-ID_FROM_SECRET_*`) is encountered, the agent must output the targeted `jsonpath` command for the user to run in their terminal, along with the exact resource identifier, file link, field path, and resolution options:
+  1. **Option A (Interactive)**: The user replies in chat with the non-sensitive string (e.g. `username: <output>`) and the agent updates the manifest and removes the TODO annotation.
+  2. **Option B (Direct Edit)**: The user updates the manifest directly using the provided file link, target field path, and diff snippet.
+
+#### 2.1 Missing Source Secret / ConfigMap (`TODO_SET_<KEY>_FROM_<KIND>_<NAME>`)
+* **Annotation**: `[ERROR] Referenced SECRET "<name>" for key "<key>" was not found in migration inputs...`
+* **Agent Handoff Output Format**:
+
+  ```markdown
```
dashpole: Because this markdown template contains inner code blocks (like ```bash and ```diff), using 3 backticks for the outer block closes it early and breaks markdown rendering for subsequent sections. Using 4 backticks (````markdown / ````) for the outer fences in these template blocks (sections 2.1, 2.2, 3.2, 4.1) will keep the inner code blocks nested cleanly.

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,645 @@
...
+  ### 1. Identify the TLS Secret in your namespace:
+  ```bash
+  kubectl get secrets -n <namespace>
+  ```
+
+  ### 2. Resolution Options:
+  * **Option A (Interactive)**: Reply with `The TLS secret name is: <secret-name>` and I will update the manifest.
+  * **Option B (Direct Edit)**: Update `spec.endpoints[<index>].tls.ca.secret.name` in `<path-to-manifest>.yaml` and remove the TODO annotation.
+
+  ```
```
bwplotka: what's this?

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,645 @@
...
+      -       username: TODO_SET_<KEY>_FROM_SECRET_<NAME>
+      +       username: "<your-retrieved-username>"
+              password:
+                secret:
+                  name: <secret-name>
+                  key: password
+      ```
+    * Remove the `gmp.googleapis.com/todo-*` annotation from `metadata.annotations`.
+
+  ```
```
bwplotka: ?

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,645 @@
...
+     endpoints:
+  -  - port: TODO_SET_PORT
+  +  - port: metrics
+  ```
+
+---
+
+### Recipe 2: Missing or Incomplete Secrets & ConfigMaps
+
+#### Security Policy: Zero Secret Ingestion & Human-In-The-Loop
```
bwplotka, line 191: Can we add newlines between header and paragraphs everywhere?

bwplotka (approved):
Looks good, just a few nits

cmd/gmp-migrate/SKILL.md
karthunni: Whoops, need to use ```` instead

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,650 @@
...
+  GMP `spec.selector` **only matches Kubernetes Pod labels**, never Pod annotations. In Prometheus Operator, scraped pods are the intersection of the monitor's `spec.selector` and the annotation rule:
+  $$\text{Scraped Targets} = (\text{Pods matching } \texttt{spec.selector}) \cap (\text{Pods matching Annotation Rule})$$
+
+  ##### Case A: Reconciling `action: keep` (e.g. `prometheus.io/scrape: "true"`)
+
+  1. **Intersection Query**: Query for workloads matching **both** the monitor's original `spec.selector` AND the annotation:
+
+     ```bash
+     kubectl get deployment,statefulset,daemonset -n <namespace> -l <original-monitor-selector> \
+       -o jsonpath='{range .items[?(@.spec.template.metadata.annotations["prometheus.io/scrape"]=="true")]}{.kind}{"/"}{.metadata.name}{"\n"}{end}'
```
dashpole: Kubectl jsonpath fails with an invalid array index error on bracketed map keys. Do we need to use `jq` or something instead?

cmd/gmp-migrate/SKILL.md
```diff
@@ -0,0 +1,650 @@
...
+  3. **Generate Companion Workload Patches**: Add `labels: { prometheus.io/scrape: "true" }` to each matching workload discovered in step 1.
+  4. **Warn & Prompt**: Warn the user that patching the workload pod templates will trigger a rolling restart of their pods, and request approval before applying.
+
+  ##### Case B: Reconciling `action: drop` (e.g. `prometheus.io/scrape: "false"`)
+
+  1. **Identify Excluded Workloads**: Query for workloads matching `spec.selector` that have the exclusion annotation:
+
+     ```bash
+     kubectl get deployment,statefulset,daemonset -n <namespace> -l <original-monitor-selector> \
+       -o jsonpath='{range .items[?(@.spec.template.metadata.annotations["prometheus.io/scrape"]=="false")]}{.kind}{"/"}{.metadata.name}{"\n"}{end}'
```
dashpole: Kubectl jsonpath fails with an invalid array index error on bracketed map keys. We can pipe -o json to jq instead:
```suggestion
       -o json | jq -r '.items[] | select(.spec.template.metadata.annotations["prometheus.io/scrape"]? == "false") | "\(.kind)/\(.metadata.name)"'
```

dashpole (approved):

PR #2239, opened 2026-09-02
chore: bump config-reloader memory limit to 64M

avilevy18:
Bumps config-reloader container memory limit from 32M to 64M in charts/values.global.yaml and regenerates manifests.


bwplotka (approved):
Thanks!

PR #2260, opened 2026-09-04
fix(deps): update module google.golang.org/grpc to v1.83.2 [security]

avilevy18:
Updates `google.golang.org/grpc` to `v1.83.2` on `main` to resolve dependency security vulnerabilities.

This dependency update pulls in `google.golang.org/api` `v0.264.0`, where `option.WithCredentialsFile` and `option.WithCredentialsJSON` have been marked deprecated. We keep using the deprecated methods for now and add `//nolint:staticcheck` annotations to allow lint to pass cleanly.

Supersedes / fixes #2210.


bwplotka (approved):
Thanks - I agree it's a safer option.

We need to update https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2260/ then too.

bwplotka:
Ah you got it, thanks! https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2261

PR #2281, opened 2026-09-11
feat(ops): Allow gmpctl to work without a working `gh` CLI

quentinmit:


ops/gmpctl/github.go
```diff
@@ -16,13 +16,37 @@ package main
...
+	"fmt"
+	"net/url"
 )
 
 type prInfo struct {
 	URL string `json:"url"`
 }
 
-func mustEnsurePullRequest(dir, baseBranch, headBranch, title, body string) {
+func (p Project) mustEnsurePullRequest(dir, baseBranch, headBranch, title, body string) {
```
bwplotka, line 27: nit: is there a reason project is not an argument? I don't see the reason, could be a first arg? 🤔

bwplotka (approved):
Thanks!

ops/gmpctl/github.go
quentinmit: All the other functions that use `Project` are methods, so it seemed like I was following the local pattern to make this a method as well.

PR #2284, opened 2026-09-14
fix(operator): sort SecretConfigs to ensure reproducible collector co…

pintohutch:
…nfigs

When multiple secrets are referenced in PodMonitoring or ClusterPodMonitoring resources, Go's randomized map iteration over PrometheusSecretConfigs causes cfg.SecretConfigs to be emitted in non-deterministic order. This leads yaml.Marshal to generate different ConfigMap bytes across reconcile passes, which triggers the ConfigMap watch and creates an endless reconcile loop.

Sort SecretConfigs deterministically by name in PrometheusSecretConfigs.

Fixes b/561366861


bwplotka (approved):
Thanks!

PR #2285, opened 2026-09-14
fix(e2e): wait for operator webhook caBundle in WaitForOperatorReady

pintohutch:
When the operator starts up, continuouslySetCABundle waits 5 seconds before populating the CA bundle on the ValidatingWebhookConfiguration. Meanwhile, WaitForOperatorReady previously only checked that the operator deployment was ready. Because the readiness probe (/readyz) reports ready as soon as the HTTPS listener starts, e2e tests (e.g. setupCluster when configuring explicit GCM credentials) could immediately attempt to create or update resources like OperatorConfig.

When this occurs within the 5-second window, the API server invokes the validating webhook before the caBundle has been populated. Since the operator generates a self-signed certificate, the API server fails TLS verification with:
  "tls: failed to verify certificate: x509: certificate signed by unknown authority"

Avoid this race condition by waiting for the ValidatingWebhookConfiguration to have a non-empty CABundle in WaitForOperatorReady before declaring the operator ready.


bwplotka (approved):
Even more args to remove webhook.. Thanks!

PR #2288, opened 2026-09-15
perf(ops): ignore ops/.bin in .dockerignore

pintohutch:
Previously, .dockerignore had '!ops', which un-ignored all of ops/ including ops/.bin. Because ops/.bin stores ~155 MB of compiled host binaries (helm, yq, gcrane, addlicense, govulncheck) downloaded by gmpctl.sh, this 155 MB payload was transferred and copied into Docker build contexts (such as in hack/Dockerfile).

Explicitly ignore ops/.bin in .dockerignore to prevent local binaries from leaking into Docker images.


bwplotka (approved):

.dockerignore
```diff
@@ -16,6 +16,7 @@
 !third_party
 !vendor
 !internal
+ops/.bin
```
bwplotka: Let's ignore all - notably ops/data can be expensive

```suggestion
ops
```

bwplotka:
data is also useful to ignore, added in separate PR https://github.com/GoogleCloudPlatform/prometheus-engine/pull/2290

PR #2289, opened 2026-09-15
perf(main): add BuildKit Go module and build cache mounts

pintohutch:
Add BuildKit cache mounts (--mount=type=cache) for /go/pkg/mod and /root/.cache/go-build to all Go Dockerfiles.

See https://docs.docker.com/build/cache/optimize/#use-cache-mounts for more detail.

Previously, whenever source code changed, the Docker layer was invalidated and 'go build' had to re-download all Go modules from proxy.golang.org and recompile all unchanged packages and CGO BoringCrypto dependencies from scratch. With BuildKit cache mounts, modules and compiler artifacts are cached across builds, which should speed up incremental rebuilds.


cmd/config-reloader/Dockerfile
```diff
@@ -32,7 +32,9 @@ ENV GOFIPS140=off
 ENV GOTOOLCHAIN=local
 ENV GOOS=${TARGETOS}
 ENV GOARCH=${TARGETARCH}
-RUN if [ "${TARGETARCH}" = "arm64" ] && [ "${BUILDARCH}" != "arm64" ]; then \
+RUN --mount=type=cache,target=/go/pkg/mod \
+	--mount=type=cache,target=/root/.cache/go-build \
+	if [ "${TARGETARCH}" = "arm64" ] && [ "${BUILDARCH}" != "arm64" ]; then \
```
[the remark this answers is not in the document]
bwplotka: not true, it's a chained command

cmd/config-reloader/Dockerfile
```diff
@@ -32,7 +32,9 @@ ENV GOFIPS140=off
 ENV GOTOOLCHAIN=local
 ENV GOOS=${TARGETOS}
 ENV GOARCH=${TARGETARCH}
-RUN if [ "${TARGETARCH}" = "arm64" ] && [ "${BUILDARCH}" != "arm64" ]; then \
+RUN --mount=type=cache,target=/go/pkg/mod \
+	--mount=type=cache,target=/root/.cache/go-build \
```
bwplotka, line 36: Suspicious about the "/root" path here. Checked and they seemed to be correct at the moment:

```
$ docker run --rm google-go.pkg.dev/golang:1.26.7@sha256:2f110d472e26e5c6b1c93f40a3ea6f88fe8954902cd4518d7ce99d836a158ceb \
    sh -c 'id; echo HOME=$HOME; go env GOCACHE GOMODCACHE'
uid=0(root) gid=0(root) groups=0(root)
HOME=/root
/root/.cache/go-build
/go/pkg/mod
```

Some notes though:

1. We can make it more flexible for future: `/root/.cache/go-build` is only correct because the image happens to run as root with `HOME=/root`. If the base ever moves to a non-root user, changes `HOME`, or sets `XDG_CACHE_HOME`, the mount keeps working but caches nothing, and there's no error to tell us. We could set the paths ourselves removes the coupling:

   ```dockerfile
   ENV GOMODCACHE=/gocache/mod
   ENV GOCACHE=/gocache/build
   RUN --mount=type=cache,target=/gocache/mod \
   	--mount=type=cache,target=/gocache/build \
   	...
   ```

2. Just for awareness: I don't think CI gets anything from this 🤔  The win is local iterative builds, which is still worth it.
3. Just for awareness: While this should be fine for every build path we have, plain `docker build` on a daemon without BuildKit now fails with `Unknown flag: mount`. 
4. The cross-compiler `apt-get` in the same `RUN` is still uncached. `--mount=type=cache,target=/var/cache/apt,sharing=locked` plus dropping `/etc/apt/apt.conf.d/docker-clean` would cover that too.

bwplotka (approved):
This blocks builds without buildkit, but I think that's ok. Some nits, but we can iterate.

cmd/config-reloader/Dockerfile
pintohutch: Thanks for the feedback!

1. Good point - I went ahead with your recommendation.
2. Ack yep
3. I think that's an acceptable trade-off considering docker supports buildkit OOTB for [~3y](https://docs.docker.com/engine/release-notes/23.0/#2300) now.
4. Made this change as well.

PR #2300, opened 2026-09-18
examples: add config-reloader e2e memory benchmark manifests

bwplotka:
Adds an e2e memory benchmark setup (`examples/config-reloader-bench/`) to compare memory usage (`container_memory_rss`, `container_memory_working_set_bytes`, `process_resident_memory_bytes`, `go_memstats_*`, and `container_oom_events_total`) between:
- `v0.18` (buffered in-memory reloader) vs. `v0.19.2`/`main` (streaming `thanos-io/thanos@531d98a` reloader)
- Default 4KB page size nodes (`default-pool`) vs. 64KB page size ARM64 nodes (`ubuntu-gke-*-arm64-*-linux64k`)
- Varying Prometheus configuration sizes (10, 1,000, up to 8,500 scrape jobs / ~1 MiB gzipped -> ~33.7 MiB uncompressed via `gen_configmap.go`)

Each simplified Collector Pod uses a lightweight `prometheus-noop` container (`--export.disable` with a minimal static config) so `config-reloader` runs its full decompression, environment variable expansion, atomic write, and `/-/reload` cycle without Prometheus scraping targets or exporting metrics.


bwplotka:
Too convoluted, we need something simpler.

PR #2301, opened 2026-09-18
chore(deps): bump prometheus to v3.13.0-gmp.1-gke.0

bwplotka:
This effectively switches GMP to the 3.x Prometheus fork (`release-3.13.0-gmp`).

### Changes
- **Dependencies & Images**:
  - Bump `github.com/prometheus/prometheus` to `github.com/GoogleCloudPlatform/prometheus` `v0.0.0-20260903164310-65985e7c948b` (`release-3.13.0-gmp`) and remove obsolete v2.53 `replace` directives.
  - Switch Prometheus image to `gke.gcr.io/prometheus-engine/prometheus:v3.13.0-gmp.1-gke.0@sha256:b8e81d2737a3ca7126b06cbd4bb495b4f9c694e9431e5a280d7d37f289ff2708` (`charts/values.global.yaml`, `examples/prometheus.yaml`, `manifests/operator.yaml`).
- **Prometheus 2.x Compatibility Defaults**:
  - Configure `metric_name_validation_scheme: legacy` and `metric_name_escaping_scheme: underscores` in collector `global` config (`pkg/operator/collection.go`).
  - Configure `fallback_scrape_protocol: PrometheusText0.0.4` on generated `ScrapeConfig`s (`pkg/operator/apis/monitoring/v1/common_types.go`, `pkg/operator/apis/monitoring/v1/operator_config.go`).
- **Prometheus 3.x & `controller-runtime` API Updates**:
  - `cmd/rule-evaluator`: Migrate to `promslog` (`*slog.Logger`), `notifier.Send`, `rules.NewManager` options, `labels.Compare`, and updated `storage.Select` / `labels.Labels` (`stringlabels`) APIs.
  - `pkg/operator`: Default unset `relabel.Config.Regex` to `relabel.DefaultRelabelConfig.Regex` (required by `relabel.Regexp.IsZero()`), update `rulefmt.Parse` calls, and migrate webhooks to typed `admission.Validator[T]` / `admission.Defaulter[T]`.


cmd/rule-evaluator/main.go
```diff
@@ -900,6 +906,7 @@ func (db *queryAccess) Close() error {
 type ruleEvaluator struct {
 	ctx             context.Context
 	logger          log.Logger
```
bwplotka: Do we need two loggers? Can we remove one?

cmd/rule-evaluator/main_test.go
```diff
@@ -179,7 +180,7 @@ func TestSelect(t *testing.T) {
 			if !cmp.Equal(got.Warnings(), c.want.Warnings(), cmp.Comparer(compareAnnotationsEquality)) {
 				t.Errorf("Case %d: Expected warnings %s, Actual warnings: %s", i, c.want.Warnings(), got.Warnings())
 			}
-			if diff := cmp.Diff(expandSeriesSet(got), c.want.m); diff != "" {
+			if diff := cmp.Diff(expandSeriesSet(got), c.want.m, cmp.Transformer("Labels", func(ls labels.Labels) string { return ls.String() })); diff != "" {
```
bwplotka: can we use equal?

https://github.com/prometheus/prometheus/blob/296080c0a504ea0e72941ef16143f0a9702b137b/util/testutil/cmp.go#L36

pkg/operator/apis/monitoring/v1/common_types.go
```diff
@@ -142,6 +142,16 @@ func buildPrometheusScrapeConfig(jobName string, discoverCfgs discovery.Configs,
...
+	for _, c := range relabelCfgs {
+		if c.Regex.Regexp == nil {
+			c.Regex = relabel.DefaultRelabelConfig.Regex
+		}
+	}
+	for _, c := range metricRelabelCfgs {
+		if c.Regex.Regexp == nil {
+			c.Regex = relabel.DefaultRelabelConfig.Regex
+		}
+	}
```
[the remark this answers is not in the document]
bwplotka: addressed in fcc2e3687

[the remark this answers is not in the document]
bwplotka: addressed in 7b076dce9

pkg/operator/operator_config_test.go
```diff
@@ -63,7 +66,7 @@ google_cloud:
 			CredentialsFile: "credentials2.json",
 		},
 	}
-	if diff := cmp.Diff(expected, out); diff != "" {
+	if diff := cmp.Diff(expected, out, cmp.Transformer("Labels", func(ls labels.Labels) string { return ls.String() }), cmpopts.IgnoreUnexported(promforkconfig.Config{})); diff != "" {
```
bwplotka: This should be labels.Equal

pkg/operator/operator_config_test.go
```diff
@@ -77,13 +80,32 @@ google_cloud:
 	expectedYAML := `global:
     scrape_interval: 1m
     scrape_timeout: 10s
-    scrape_protocols:
-        - OpenMetricsText1.0.0
-        - OpenMetricsText0.0.1
-        - PrometheusText0.0.4
     evaluation_interval: 1m
+    metric_name_validation_scheme: utf8
```
bwplotka, line 86: This needs to be legacy, tbd

pkg/operator/apis/monitoring/v1/common_types.go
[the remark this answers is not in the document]
bwplotka: We never need to do this, seems odd.

bwplotka (commented):
Self review.

pkg/operator/operator_config_test.go
```diff
@@ -93,7 +115,7 @@ runtime:
 	if err := yaml.Unmarshal(outBytes, &out); err != nil {
 		t.Fatal(err)
 	}
-	if diff := cmp.Diff(expected, out); diff != "" {
+	if diff := cmp.Diff(expected, out, cmp.Transformer("Labels", func(ls labels.Labels) string { return ls.String() }), cmpopts.IgnoreUnexported(promforkconfig.Config{})); diff != "" {
```
bwplotka: ditto

cmd/rule-evaluator/internal/api.go
```diff
@@ -138,19 +139,11 @@ func alertsToAPIAlerts(alerts []*rules.Alert) []*apiv1.Alert {
 	}
 	// Sort for testability.
 	slices.SortFunc(apiAlerts, func(a, b *apiv1.Alert) int {
-		ha, hb := a.Labels.Hash(), b.Labels.Hash()
-		if ha != hb {
-			if ha > hb {
-				return -1
-			}
-			return 1
+		if cmp := labels.Compare(a.Labels, b.Labels); cmp != 0 {
```
bwplotka, line 142: Rationale, for reviewers:
  * In Prometheus 2.53 (`slicelabels`), `Labels.Hash()` hashed `\xff`-separated strings (`name\xffvalue\xff...`).
  * In Prometheus 3.13 (`stringlabels` in `labels_stringlabels.go`), `Labels` is stored as a length-prefixed byte string (`ls.data`), and `Hash()` runs `xxhash.Sum64(yoloBytes(ls.data))`.

Because the `xxhash` output values changed, sorting alerts by descending `a.Labels.Hash()` in `alertsToAPIAlerts` reversed the relative order of `test-alert-1` and `test-alert-2`, causing `TestAPI_HandleAlertsEndpoint/only_firing_alerts` to fail (`test-alert-2` sorted before `test-alert-1`).

pkg/migrate/golden_test.go
```diff
@@ -55,12 +55,16 @@ func newPORelabelRule(action relabel.Action, src []string, target string, regex
 	for _, s := range src {
 		srcNames = append(srcNames, prommodel.LabelName(s))
 	}
+	re := relabel.DefaultRelabelConfig.Regex
```
bwplotka, line 58: Generally I find those golden test odd.. we take PO objects, migrate to GMP and normalize GMP, so we can compare with different PO objects. We probably could store expected GMP objects as golden tests files? 🤔 

Nevertheless rationale for reviewers for this change: 

When we construct `&relabel.Config{...}` struct literals in Go without setting `Regex`, `c.Regex.Regexp` is `nil`:
1. `c.Regex.IsZero()` evaluates `nil == DefaultRelabelConfig.Regex.Regexp` $\rightarrow$ **`false`**.
2. Because `IsZero()` is `false`, `yaml.Marshal` does **not** omit `regex` (`omitempty` is skipped) and calls `MarshalYAML()`, which returns `(nil, nil)` and serializes **`regex: null`** into every relabel rule in YAML.
3. When Prometheus unmarshals `regex: null`, `Regexp.UnmarshalYAML` unmarshals `null` as `""` (`^(?s:)$` — matching only empty strings instead of the default `(.*)`), which breaks relabeling.

Setting `c.Regex = relabel.DefaultRelabelConfig.Regex` when `c.Regex.Regexp == nil` makes `IsZero()` return `true` so `yaml.Marshal` omits `regex` as intended.

bwplotka:
This should be ready to go! cc @bernot-dev

bernot-dev:
I think rule evaluator also needs to enforce `NameValidationScheme: model.LegacyValidation`.

https://github.com/bwplotka/prometheus-engine/blob/bump_prometheus_version_3_13/cmd/rule-evaluator/main.go#L953

pkg/operator/apis/monitoring/v1/rules_config.go
bernot-dev: This doesn't actually happen. Looks like it's been broken for a while. The yaml parsing is external to Prometheus.

I think we should call the new [Parse](https://pkg.go.dev/github.com/prometheus/prometheus@v0.313.0/model/rulefmt#Parse)/[Validate](https://pkg.go.dev/github.com/prometheus/prometheus@v0.313.0/model/rulefmt#Rule.Validate)

cmd/rule-evaluator/main.go
bernot-dev, line 42: If we're moving to slog/promslog, let's get rid of go-kit entirely.

bernot-dev (changes requested):
Thanks for working on this. I just found a few more things.

cmd/rule-evaluator/main.go
bwplotka: Agree, but let's do in separate PR, started early work on this: https://github.com/bwplotka/prometheus-engine/compare/bump_prometheus_version_3_13...bwplotka:prometheus-engine:remove_gokit_logger_adapter?expand=1

We won't remove it from go.mod though as `reloader` (Thanos) and forks' `opts.NewExporter` still use go-kit as of now -- that's fine, we can iterate.

bernot-dev (approved):
LGTM

cmd/rule-evaluator/main.go
```diff
@@ -942,10 +951,12 @@ func newRuleEvaluator(
 	queryFunc := newQueryFunc(logger, v1api)
 
 	rulesManager := rules.NewManager(&rules.ManagerOptions{
-		ExternalURL: getExternalURL(evaluatorOpts.GeneratorURL, evaluatorOpts.ProjectID),
-		QueryFunc:   queryFunc,
-		Context:     ctx,
-		Appendable:  appendable,
+		// Ensure compatibility with 2.x Prometheus logic. See go/gmp:prom-3.13.
+		NameValidationScheme: model.LegacyValidation,
```
bwplotka, line 955: Actually we could allow it here - it's not a breaking change here. To discuss in separate change.

PR #2303, opened 2026-09-21
docs(agents): add agent contribution guidelines

bwplotka:
This is mostly to ensure agents apply "conform" rules, but added a few other items based on the last PR reviews.


AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+2. **`replace` Directives & Pins in `go.mod`**:
+   When adding or updating a `replace` directive (e.g., for `github.com/prometheus/prometheus`, `client_golang`, `common`, or `thanos`), keep the `require` and `replace` versions consistent and document the reason for the pin in a comment directly above the `replace` entry.
+3. **Dockerfile Efficiency & BuildKit Caching**:
+   * Scope `COPY` directives in `cmd/<component>/Dockerfile` to only the required paths (`go.mod`, `go.sum`, `pkg/`, `internal/`, `cmd/<component>/`) rather than copying the entire repository or all of `cmd/`.
+   * Preserve BuildKit cache mounts (`--mount=type=cache,target=/go/pkg/mod` and `--mount=type=cache,target=/root/.cache/go-build`) and keep heavy local data/binary directories (`ops/gmpctl/data`, `ops/.bin`) excluded in `.dockerignore`. Note that BuildKit cache mounts apply to the entire `RUN` instruction they are defined on, including all parts of a chained command.
```
dashpole: Commit `78e9043e` updated all `Dockerfile`s to set `GOMODCACHE=/gocache/mod` and `GOCACHE=/gocache/build`, so the cache mount targets here should be `/gocache/mod` and `/gocache/build` instead of `/go/pkg/mod` and `/root/.cache/go-build`:
```suggestion
   * Preserve BuildKit cache mounts (`--mount=type=cache,target=/gocache/mod` and `--mount=type=cache,target=/gocache/build`) and keep heavy local data/binary directories (`ops/gmpctl/data`, `ops/.bin`) excluded in `.dockerignore`. Note that BuildKit cache mounts apply to the entire `RUN` instruction they are defined on, including all parts of a chained command.
```

